PR #95, opened 2022-11-15
layout: ELK improvements

alixander:
it's hard to say if this is strictly better. we want the node flexibility, but as a result of it, it feels like half of the tests got worse. 

Namely, there's no way for edges to prefer symmetry

Screen Shot 2022-11-28 at 10 07 19 AM


nhooyr:
Update the changelog please.

nhooyr:
@alixander since this has been approved by @ejulio-ts, what's keeping it from merging? gnarly merge conflicts? think we can get this in for [v0.0.14](https://github.com/terrastruct/d2/milestone/1) on monday?

alixander:
i'm uncertain that these new parameters make elk overall better. i still need to fine tune i think. as well as investigate improvements from issues like https://github.com/terrastruct/d2/issues/282

oslTrifork:
Hi, sorry for the intrusion but given the broad aspect of the PR it seemed appropriate.

Would it be possible to improve the layout case of this:
image

although it's a stylistic choice of course, I feel like it would be more appropriate if connection arrows went out of the edges of the diamond rather than symmetrically from the bottom. This would only be specific to the diamond shape though.

I also agree that the changes does not seem like improvements to me, at least when looking at the tests... I'd rather have symmetry and bended lines than straight ones if the overall layout is shifted as a result of the changes.

alixander:
https://github.com/terrastruct/d2/pull/627

PR #97, opened 2022-11-15
Add [From/To]TableRowIndex to edges

ejulio-ts:
We want to be able to route between table columns (rows in the diagram table) so that edges follow from the source to target rows, thus making it easier to read Entity Relationship Diagrams.

Once we have this routing in place, we don't need to label the edges with the connection anymore.

The routing is specific to each layout engine, so for `dagre` and `elk` we need to investigate how we can route this properly, or we can work on a post-processing step to handle it properly.
An alternative, is to have this behavior as a compiler option and for some layout engines we enable `row to row` routing and for others we fallback to the previous behavior of having the edge label.

Tests changed because of the two new attributes.


d2graph/d2graph.go
```diff
@@ -607,6 +607,9 @@ type Edge struct {
 	MinWidth  int `json:"minWidth"`
 	MinHeight int `json:"minHeight"`
 
+	FromTableRowIndex int `json:"fromTableRowIndex"`
```
alixander, line 610: let's make these pointers and omit empty, since they won't exist most of the time

d2compiler/compile.go
```diff
@@ -682,10 +682,9 @@ func (c *compiler) compileSQLTable(obj *d2graph.Object) {
 			if srcID == absID {
 				// Frontend isn't aware of container IDs.
```
alixander, line 683: we can get rid of this comment while ur here

testdata/d2compiler/TestCompile/edge_arrowhead_fields.exp.json
```diff
@@ -253,6 +253,8 @@
         "index": 0,
         "minWidth": 0,
         "minHeight": 0,
+        "fromTableRowIndex": 0,
```
alixander, line 256: is there a test where this is non 0?

d2compiler/compile.go
```diff
@@ -682,10 +682,9 @@ func (c *compiler) compileSQLTable(obj *d2graph.Object) {
...
 				d2Col.Reference = strings.TrimPrefix(dstID, parentID+".")
-				relSrc := strings.TrimPrefix(absID, parentID+".")
-				e.Attributes.Label.Value = fmt.Sprintf("%s %s %s", relSrc, e.ArrowString(), d2Col.Reference)
-				// removeContainer() will adjust the edge to point to the table and not inside.
-				break
+				e.FromTableRowIndex = len(obj.SQLTable.Columns)
```
alixander, line 685: let's keep the terminology the same.

s/from/src
s/to/dst

alixander (changes requested):
s/column/row. the rendering of a SQL table where its columns are rows is just one representation, maybe in the future it can be rendered vertically with fake values or something, idk.

nhooyr:
Remember to update the changelog if applicable. I don't fully understand the scope of this PR's changes and whether they'll be user facing.

testdata/d2compiler/TestCompile/edge_arrowhead_fields.exp.json
ejulio-ts: good catch.
I added the sql tables test in the previous PR, but it seems it got lost during the rebase 🤔

ci/release/changelogs/next.md
```diff
@@ -6,7 +6,7 @@ For v0.0.99 we focused on X, Y and Z. Enjoy!
 
 #### Improvements 🔧
 
-- Improves something or the other #9999
+- Add table columns indices in edges between SQL Tables so that layout engines can route exactly between rows
```
alixander, line 9: columns

alixander (changes requested):

ci/release/changelogs/next.md
ejulio-ts: 😂

e2etests/testdata/stable/sql_tables/dagre/board.exp.json
```diff
@@ -0,0 +1,396 @@
...
+      "language": "",
+      "color": "#0A0F25",
+      "italic": false,
+      "bold": true,
+      "underline": false,
+      "labelWidth": 116,
+      "labelHeight": 36
+    }
+  ],
+  "connections": [
```
alixander, line 253: this test doesn't have the variable?

a d2compiler/compile_test.go entry is appropriate here that checks value of field

alixander (changes requested):

e2etests/testdata/stable/sql_tables/dagre/board.exp.json
ejulio-ts: Because I never added these fields to the exporter.
For a moment I thought I was delusional that everything was working by chance 😂

alixander: i don't actually think this field should be exported. it's just for the layout step

ejulio-ts: so, removing then 😅

alixander: yes, this test wasn't the right one to test this change. but i think u should keep it around anyway, cuz we don't have a many to 1 sql table e2e test

d2compiler/compile.go
```diff
@@ -746,6 +746,14 @@ func flattenContainer(g *d2graph.Graph, obj *d2graph.Object) {
 			newEdge, _ = g.Root.Connect(e.Src.AbsIDArray(), obj.AbsIDArray(), e.SrcArrow, e.DstArrow, e.Attributes.Label.Value)
 		}
 		// TODO more attributes
+		if e.SrcTableColumnIndex != nil {
+			newEdge.SrcTableColumnIndex = new(int)
+			*newEdge.SrcTableColumnIndex = *e.SrcTableColumnIndex
```
alixander, line 751: one line `newEdge.SrcIndex = e.SrcIndex`?

d2compiler/compile_test.go
```diff
@@ -1474,6 +1474,32 @@ b`, g.Objects[0].Attributes.Label.Value)
...
+	id: int
+	name: string
+}
+
+dst.id <-> src.dst_id
+`,
+			assertions: func(t *testing.T, g *d2graph.Graph) {
+				srcIndex := g.Edges[0].SrcTableColumnIndex
+				if srcIndex == nil || *srcIndex != 0 {
+					t.Fatalf("Expected SrcTableColumnIndex to be 0, got %v", srcIndex)
```
alixander, line 1496: minor: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

e2etests/stable_test.go
```diff
@@ -831,6 +831,44 @@ a -> md -> b
...
+			name: "sql_tables",
+			script: `
+			users: {
+			  shape: sql_table
+			  id: int
+			  name: string
+			  email: string
+			  password: string
+			  last_login: datetime
+			}
```
alixander, line 844: i think this will be a ton of indent. look at the surrounding examples

alixander (changes requested):

d2compiler/compile.go
ejulio-ts: Not sure about copying the pointer around, so I decided to go with a new one as this is a new edge

alixander (approved):

PR #111, opened 2022-11-17
export: PNG

berniexie:
Use Playwright to convert generated svgs into pngs.

Uses Playwright-Go: https://github.com/playwright-community/playwright-go


cmd/d2/main.go
```diff
@@ -113,12 +113,28 @@ func run(ctx context.Context, ms *xmain.State) (err error) {
...
 
+	var pw *playwright.Playwright
+	var browser playwright.Browser
+	if filepath.Ext(outputPath) == ".png" {
+		pw, browser, err = png.InitPlaywright()
+		if err != nil {
+			return err
+		}
+	}
+	defer func() error {
```
alixander, line 124: put in scope above

alixander (changes requested):

lib/png/png.go
```diff
@@ -0,0 +1,103 @@
...
+
+	if browser == nil {
+		return nil, fmt.Errorf("browser is not initialized for png export")
+	}
+	page, err = browser.NewPage()
+	if err != nil {
+		return nil, err
+	}
+	encodedSVG := base64.StdEncoding.EncodeToString(svg)
+	pngInterface, err := page.Evaluate(genPNGScript, "data:image/svg+xml;charset=utf-8;base64,"+encodedSVG)
```
alixander, line 72: looking at the API for Evaluate, it looks like u also have to check for values like `undefined`, which might have err = nil but were unserializable. there are others, look at their docs or source code for exhaustive list. 

or maybe u can have the javascript return an object, with single field "data", and unmarshal it here. if the unmarshaling fails, it means something went wrong

lib/png/png.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			return err
+		}
+		return nil
+	}()
+
+	if browser == nil {
+		return nil, fmt.Errorf("browser is not initialized for png export")
+	}
+	page, err = browser.NewPage()
```
alixander, line 67: is this the right usage? do we need to keep opening and closing new pages?

from https://playwright.dev/docs/api/class-browser#browser-new-page


> This is a convenience API that should only be used for the single-page scenarios and short snippets. Production code and testing frameworks should explicitly create [browser.newContext([options])](https://playwright.dev/docs/api/class-browser#browser-new-context) followed by the [browserContext.newPage()](https://playwright.dev/docs/api/class-browsercontext#browser-context-new-page) to control their exact life times.

lib/png/png.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, err
+	}
+
+	pngString := fmt.Sprintf("%v", pngInterface)
+	pngPrefix := "data:image/png;base64,"
+	if !strings.HasPrefix(pngString, pngPrefix) {
+		return nil, fmt.Errorf("playwright export generated invalid png")
+	}
+	splicedPNGString := pngString[len(pngPrefix):]
+	outputImage, err = base64.StdEncoding.DecodeString(splicedPNGString)
```
alixander, line 83: can just return this

cmd/d2/main.go
```diff
@@ -113,12 +113,28 @@ func run(ctx context.Context, ms *xmain.State) (err error) {
...
 	ms.Log.Debug.Printf("using layout plugin %s (%s)", envD2Layout, pluginLocation)
 
+	var pw *playwright.Playwright
+	var browser playwright.Browser
+	if filepath.Ext(outputPath) == ".png" {
+		pw, browser, err = png.InitPlaywright()
+		if err != nil {
+			return err
+		}
+		defer func() error {
```
alixander, line 123: ur not doing anything with this error. maybe just ms.log.error it if cleanup returns an error

alixander (changes requested):
tested and works well. exciting.

lib/png/png.go
berniexie: Interesting. In this case I'm gonna create a new object to store all the various parts of playwright instead of passing them all around.

berniexie: There's a little bit of trickiness here. I can only close browserContext properly when we don't use watcher. If we command+c out of watcher and the browserContext.close() function throws an error saying the browser is already closed, not sure why. Maybe the signal interrupt just shuts everything down ASAP?

lib/png/png.go
```diff
@@ -0,0 +1,137 @@
...
+	"os/exec"
+	"strings"
+
+	_ "embed"
+
+	"github.com/playwright-community/playwright-go"
+	"oss.terrastruct.com/d2/lib/xmain"
+)
+
+type Playwright struct {
```
berniexie, line 18: Taking suggestions for a better name for this struct

lib/png/png.go
```diff
@@ -0,0 +1,137 @@
...
+	encodedSVG := base64.StdEncoding.EncodeToString(svg)
+	pngInterface, err := page.Evaluate(genPNGScript, "data:image/svg+xml;charset=utf-8;base64,"+encodedSVG)
+	if err != nil {
+		return nil, err
+	}
+
+	pngString := fmt.Sprintf("%v", pngInterface)
+	pngPrefix := "data:image/png;base64,"
+	if !strings.HasPrefix(pngString, pngPrefix) {
+		ms.Log.Error.Printf("failed to convert D2 file to PNG")
```
berniexie, line 132: failed to export?

do we need to mention D2 here?

lib/png/png.go
alixander: it could always be in a "closed" state, we should be able to handle that (ignore)

lib/png/png.go
alixander: seems fine. maybe `PlaywrightContext`

lib/png/png.go
alixander: u don't need 2 errors msgs (here and above), the cli should report the error returned.

this one should be `invalid PNG: ` and then have the pngString as part of the error msg

alixander (changes requested):

alixander (changes requested):
also update the references to PNG. the 3 mentions from top-down:

[file.svg|file.png]

d2 compiles and renders file.d2 to svg or png. 

when outputting SVG, bundle all assets and layers into the output file.

Screen Shot 2022-11-17 at 12 01 12 PM
 
you'll need to update the man page too

lib/png/png.go
berniexie: The weird thing is Browser.IsConnected() returns true when browserContext.close() is called. Based on what the creator says here though https://github.com/playwright-community/playwright-go/issues/247#issuecomment-1065885796, it seems like Browser.close() should be enough as it closes everything underneath it.

ci/release/template/man/d2.1
```diff
@@ -9,7 +9,7 @@
 .Op Fl -watch Ar false
```
berniexie, line 9: I ran ./install.sh after updating this but the man pages still rendered the old version. When I manually updated the man page though it showed me this:
Screen Shot 2022-11-17 at 1 34 03 PM

alixander: just run `man ./path/to/man`

berniexie:
Screen Shot 2022-11-17 at 2 16 33 PM

alixander:
just `to svg or png.`, make sure the ending period is there

berniexie:
Screen Shot 2022-11-17 at 2 33 49 PM

cmd/d2/main.go
```diff
@@ -165,12 +180,19 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, input
...
 	}
-	svg, err = plugin.PostProcess(ctx, svg)
+	outputImage, err := plugin.PostProcess(ctx, svg)
 	if err != nil {
 		return nil, err
 	}
 
-	err = ms.WritePath(outputPath, svg)
+	if filepath.Ext(outputPath) == ".png" {
+		outputImage, err = png.ExportPNG(ms, page, svg)
```
alixander, line 189: wouldn't this ignore the postprocessing? u want to use `outputImage` not `svg`

alixander (changes requested):

cmd/d2/watch.go
```diff
@@ -325,7 +329,15 @@ func (w *watcher) compileLoop(ctx context.Context) error {
 			recompiledPrefix = "re"
 		}
 
-		b, err := compile(ctx, w.ms, w.layoutPlugin, w.inputPath, w.outputPath)
+		if filepath.Ext(w.outputPath) == ".png" && !w.pw.Browser.IsConnected() {
+			newPW, err := w.pw.RestartBrowser()
+			if err != nil {
+				return fmt.Errorf("png exporter has disconnected")
```
alixander, line 335: wrap the error

lib/png/generate_png.js
```diff
@@ -0,0 +1,23 @@
...
+      tempImg.onerror = () => {
+        reject("error loading string as an image");
+      };
+      tempImg.src = imgString;
+    });
+  };
+  const img = await loadImage();
+  const canvas = document.createElement("canvas");
+  canvas.width = Math.min(img.width, exportMaxWidth);
+  canvas.height = (canvas.width * img.height) / img.width;
```
alixander, line 16: do u wanna use the img width or the potentially clipped canvas width?

lib/png/png.go
```diff
@@ -0,0 +1,127 @@
...
+	driver, err := playwright.NewDriver(&playwright.RunOptions{})
+	if err != nil {
+		return Playwright{}, err
+	}
+	if _, err := os.Stat(driver.DriverBinaryLocation); errors.Is(err, os.ErrNotExist) {
+		err = playwright.Install()
+		if err != nil {
+			return Playwright{}, err
+		}
+	} else if err == nil {
```
alixander, line 69: what if err != nil and not errNotExist?

lib/png/png.go
```diff
@@ -0,0 +1,127 @@
...
+		return Playwright{}, err
+	}
+	context, err := browser.NewContext()
+	if err != nil {
+		return Playwright{}, err
+	}
+	page, err := context.NewPage()
+	if err != nil {
+		return Playwright{}, err
+	}
```
alixander, line 95: this is repeated. can make this into a `Start()` function.

alixander (changes requested):

lib/png/generate_png.js
berniexie: I'm not sure what you mean? if the canvas width is maxWidth then the canvasHeight needs to be adjusted accordingly too. This code is the same as in the exporter

alixander: because of the Math.min, canvas.width and img.width might be different. just confirming that this is intentional the way u have it

berniexie: the image will be drawn onto the canvas and the canvas.width width, not the img.width, so it won't be clipped

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+			return Playwright{}, fmt.Errorf("could not install Playwright driver: %v\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", err.Error())
+		}
+		if !bytes.Contains(output, []byte(driver.Version)) {
+			err = playwright.Install()
+			if err != nil {
+				return Playwright{}, err
+			}
+		}
+	} else {
+		return Playwright{}, fmt.Errorf("could not find Playwright binary: %v\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", err.Error())
```
berniexie, line 85: Unlikely errors such as Permissions or invalid path. I thought it'd be better if users reported such things were happening

alixander (changes requested):
let's treat all non-usage errors as reportable errors and `defer xerrors` append the reporting link

berniexie:
For now I kept the reporting message from errors that I believe to be are unpredictable/user generated

cmd/d2/watch.go
```diff
@@ -325,7 +329,19 @@ func (w *watcher) compileLoop(ctx context.Context) error {
...
 
-		b, err := compile(ctx, w.ms, w.layoutPlugin, w.inputPath, w.outputPath)
+		if filepath.Ext(w.outputPath) == ".png" && !w.pw.Browser.IsConnected() {
+			newPW, err := w.pw.RestartBrowser()
+			if err != nil {
+				err = fmt.Errorf("png exporter has disconnected")
+				w.ms.Log.Error.Print(err)
+				w.broadcast(&compileResult{
+					Err: err.Error(),
+				})
```
alixander, line 339: shouldn't this just return here?

lib/png/generate_png.js
```diff
@@ -0,0 +1,22 @@
...
+      tempImg.src = imgString;
+    });
+  };
+  const img = await loadImage();
+  const canvas = document.createElement("canvas");
+  canvas.width = img.width;
+  canvas.height = img.height;
+  const ctx = canvas.getContext("2d");
+  if (!ctx) {
+    return "";
```
alixander, line 18: reject?

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (pw *Playwright) RestartBrowser() (newPW Playwright, err error) {
+	if err = pw.Browser.Close(); err != nil {
+		return Playwright{}, err
+	}
+	return startPlaywright(pw.PW)
+}
+
+func (pw *Playwright) Cleanup() (err error) {
```
alixander, line 30: no point in naming these returns

cmd/d2/watch.go
```diff
@@ -325,7 +329,19 @@ func (w *watcher) compileLoop(ctx context.Context) error {
 			recompiledPrefix = "re"
 		}
 
-		b, err := compile(ctx, w.ms, w.layoutPlugin, w.inputPath, w.outputPath)
+		if filepath.Ext(w.outputPath) == ".png" && !w.pw.Browser.IsConnected() {
+			newPW, err := w.pw.RestartBrowser()
+			if err != nil {
+				err = fmt.Errorf("png exporter has disconnected")
```
alixander, line 335: "playwright could not be restarted", and then include the original err

alixander (changes requested):
don't say "png exporter" when referring to Playwright, it makes it hard for users to help themselves (e.g. i might go and search playwright dependencies if i saw that err from a tool i was using)

lib/png/generate_png.js
berniexie: Why reject, there's no promise here

berniexie: we can return an error instead

lib/png/png.go
```diff
@@ -82,7 +82,7 @@ func InitPlaywright() (Playwright, error) {
 			}
 		}
 	} else {
-		return Playwright{}, fmt.Errorf("could not install png exporter: %v\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", err.Error())
+		return Playwright{}, fmt.Errorf("could not access playwright binary location: %w\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", err)
```
alixander, line 85: should include the actual location in the msg

alixander (changes requested):
Playwright is capitalized

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		if !bytes.Contains(output, []byte(driver.Version)) {
+			err = playwright.Install()
+			if err != nil {
+				return Playwright{}, err
+			}
+		}
+	} else {
+		return Playwright{}, fmt.Errorf("could not access Playwright binary location: %v\nerror: %w\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", driver.DriverBinaryLocation, err)
+	}
```
alixander, line 86: flow is disjointed now. just do standard `if err != nil`, no need for the else.

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+		return Playwright{}, err
+	}
+	return startPlaywright(pw)
+}
+
+//go:embed generate_png.js
+var genPNGScript string
+
+func ExportPNG(ms *xmain.State, page playwright.Page, svg []byte) (outputImage []byte, err error) {
+	if page == nil {
```
alixander, line 99: this doesn't look possible. can remove?

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return Playwright{}, err
+	}
+	return startPlaywright(pw)
+}
+
+//go:embed generate_png.js
+var genPNGScript string
+
+func ExportPNG(ms *xmain.State, page playwright.Page, svg []byte) (outputImage []byte, err error) {
```
alixander, line 98: `ConvertSVG` (i dislike public methods repeating their package names, e.g. png.ExportPNG)

cmd/d2/watch.go
```diff
@@ -325,7 +329,15 @@ func (w *watcher) compileLoop(ctx context.Context) error {
 			recompiledPrefix = "re"
 		}
 
-		b, err := compile(ctx, w.ms, w.layoutPlugin, w.inputPath, w.outputPath)
+		if filepath.Ext(w.outputPath) == ".png" && !w.pw.Browser.IsConnected() {
+			newPW, err := w.pw.RestartBrowser()
+			if err != nil {
```
alixander, line 334: u still want to broadcast the error. during `watch`, the user might have their browser and editor open (no terminal), and it'd just look frozen, so you want to give a human-friendly error msg broadcast like "issue encountered with PNG exporter, stopping".

btw everywhere else here doesn't return because it's recoverable (they just change the D2 text to be well-formed), but here it's not recoverable.

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	encodedSVG := base64.StdEncoding.EncodeToString(svg)
+	pngInterface, err := page.Evaluate(genPNGScript, "data:image/svg+xml;charset=utf-8;base64,"+encodedSVG)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate png: %w\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", err)
+	}
+
+	pngString := fmt.Sprintf("%v", pngInterface)
+	pngPrefix := "data:image/png;base64,"
```
alixander, line 110: const right above this function

alixander (changes requested):

alixander (approved):
lgtm. let's leave it up for 24 hours in case @nhooyr wants to take a look, as it's not a lot of code. otherwise plz merge after.

ci/release/template/man/d2.1
```diff
@@ -21,9 +21,7 @@
 .Nm
 compiles and renders
 .Ar file.d2
-to
-.Ar file.svg
```
nhooyr, line 25: Change this back to include the output path. Just add a comma in between the two `.Ar`.

ci/release/template/man/d2.1
```diff
@@ -9,7 +9,7 @@
 .Op Fl -watch Ar false
 .Op Fl -theme Em 0
 .Ar file.d2
-.Op Ar file.svg
+.Op Ar file.svg|file.png
```
nhooyr, line 12: This should be two separate .Op with a `|` in between.

cmd/d2/help.go
```diff
@@ -15,9 +15,9 @@ import (
 
 func help(ms *xmain.State) {
 	fmt.Fprintf(ms.Stdout, `Usage:
-  %s [--watch=false] [--theme=0] file.d2 [file.svg]
+  %s [--watch=false] [--theme=0] file.d2 [file.svg|file.png]
 
-%[1]s compiles and renders file.d2 to file.svg
+%[1]s compiles and renders file.d2 to svg or png.
```
nhooyr, line 20: .

cmd/d2/main.go
```diff
@@ -113,12 +113,26 @@ func run(ctx context.Context, ms *xmain.State) (err error) {
...
 
+	var pw png.Playwright
+	if filepath.Ext(outputPath) == ".png" {
+		pw, err = png.InitPlaywright()
+		if err != nil {
+			return err
+		}
+		defer func() {
+			cleanupErr := pw.Cleanup()
+			if cleanupErr != nil {
```
nhooyr, line 124: check if `err != nil` and set to the named return error.

cmd/d2/main.go
```diff
@@ -165,12 +180,19 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, input
 	if err != nil {
 		return nil, err
 	}
-	svg, err = plugin.PostProcess(ctx, svg)
+	outputImage, err := plugin.PostProcess(ctx, svg)
```
nhooyr, line 183: why not just `out`? image to me means png, not png or svg.

cmd/d2/watch.go
```diff
@@ -325,7 +329,20 @@ func (w *watcher) compileLoop(ctx context.Context) error {
...
+				broadcastErr := fmt.Errorf("issue encountered with PNG exporter, stopping")
+				w.ms.Log.Error.Print(broadcastErr)
+				w.broadcast(&compileResult{
+					Err: broadcastErr.Error(),
+				})
+				return fmt.Errorf("Playwright could not be restarted: %w", err)
```
nhooyr, line 340: This loop is only supposed to return if the context is cancelled. So remove this line.

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+	"oss.terrastruct.com/d2/lib/xmain"
+)
+
+type Playwright struct {
+	PW      *playwright.Playwright
+	Browser playwright.Browser
+	Page    playwright.Page
+}
+
+func (pw *Playwright) RestartBrowser() (newPW Playwright, err error) {
```
nhooyr, line 23: no need for named returns

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+	if err = pw.Browser.Close(); err != nil {
+		return Playwright{}, err
+	}
+	return startPlaywright(pw.PW)
+}
+
+func (pw *Playwright) Cleanup() error {
```
nhooyr, line 30: annotate all the error paths in this file with `fmt.Errorf`, grep the code for examples.

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return nil
+}
+
+func startPlaywright(pw *playwright.Playwright) (Playwright, error) {
+	browser, err := pw.Chromium.Launch()
+	if err != nil {
+		return Playwright{}, err
+	}
+	context, err := browser.NewContext()
```
nhooyr, line 45: Why do you need the context? Why not `browser.NewPage`?

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+		return Playwright{}, err
+	}
+	return Playwright{
+		PW:      pw,
+		Browser: browser,
+		Page:    page,
+	}, nil
+}
+
+func InitPlaywright() (Playwright, error) {
```
nhooyr, line 60: why do you need to reimplement this? Why can't `playwright.Install` be used?

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+const pngPrefix = "data:image/png;base64,"
+
+func ConvertSVG(ms *xmain.State, page playwright.Page, svg []byte) (outputImage []byte, err error) {
+	encodedSVG := base64.StdEncoding.EncodeToString(svg)
+	pngInterface, err := page.Evaluate(genPNGScript, "data:image/svg+xml;charset=utf-8;base64,"+encodedSVG)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate png: %w\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", err)
+	}
+
+	pngString := fmt.Sprintf("%v", pngInterface)
```
nhooyr, line 110: why do you need to stringify the returned interface? shouldn't it be a string already?

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+//go:embed generate_png.js
+var genPNGScript string
+
+const pngPrefix = "data:image/png;base64,"
+
+func ConvertSVG(ms *xmain.State, page playwright.Page, svg []byte) (outputImage []byte, err error) {
+	encodedSVG := base64.StdEncoding.EncodeToString(svg)
+	pngInterface, err := page.Evaluate(genPNGScript, "data:image/svg+xml;charset=utf-8;base64,"+encodedSVG)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate png: %w\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", err)
```
nhooyr, line 107: We don't need `please report this issue here`. I've never understood the value of such notices and if we do want them, that's a separate PR and they should be centrally located in a constructor of some sort or when outputting errors.

nhooyr: If someone can't find our issue tracker within 10 seconds, frankly I don't want them reporting an issue.

lib/png/png.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	pngString := fmt.Sprintf("%v", pngInterface)
+	if !strings.HasPrefix(pngString, pngPrefix) {
+		if len(pngString) > 50 {
+			pngString = pngString[0:50] + "..."
+		}
+		return nil, fmt.Errorf("invalid PNG: %v\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", pngString)
```
nhooyr, line 115:
```suggestion
		return nil, fmt.Errorf("invalid PNG: %q\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", pngString)
```

nhooyr (changes requested):

cmd/d2/watch.go
alixander: then it'd proceed to compile?

lib/png/png.go
alixander: their source code always prints out a "installing..." even when it's not.

lib/png/png.go
alixander: it's not about discoverability. it's removing friction to behavior that's helpful to us. this is basic UX to not make users work. even if the work seems trivially easy, removing it is likely many useful reports that otherwise would not be made.

cmd/d2/main.go
```diff
@@ -165,12 +180,19 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, input
...
+		outputImage, err = png.ConvertSVG(ms, page, outputImage)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	err = ms.WritePath(outputPath, outputImage)
 	if err != nil {
 		return nil, err
 	}
```
alixander, line 198: i just noticed that below this line, ur still returning `svg`, which is wrong now.

alixander (changes requested):

lib/png/png.go
alixander: agreed on the second point tho, let's just save these for separate PR that centralizes

lib/png/png.go
nhooyr: this needs a comment then. and can you link me to where in the source this is?

cmd/d2/watch.go
nhooyr: Ah right, then make it `continue`.

lib/png/png.go
nhooyr: i disagree. endless bug reports are not behaviour that's helpful for us. i promise you, if people start using d2 and it gets even a thousand, they *will* start making issues left and right regardless of whether we show this message. we'll need to hire someone full time just to manage the repo, the questions and the discussions. it will snowball hard.

at which point this will just be noise. and without centralizing, it screws up go's `%w` fmt verb which has to be at the end of the error format string. we also have no indication that it's a mistake on our end, maybe it's something wrong with their setup. giving them the issue link for everything will make users open lazy issues and we don't have time for that.

nhooyr: and linking directly to the new issue page means people won't search first, they'll just fire it off.

alixander: how about we include it while it's small scale (prioritize visibility into issues), and then remove it when it's big?

nhooyr: no i think it's useless but i'll defer as long as we remove it eventually.

cmd/d2/help.go
berniexie: Not sure what this means

lib/png/png.go
berniexie: In the Playwright docs it says
```
**This is a convenience API that should only be used for the single-page scenarios and short snippets. Production code and testing frameworks should explicitly create [browser.newContext([options])](https://playwright.dev/docs/api/class-browser#browser-new-context) followed by the [browserContext.newPage()](https://playwright.dev/docs/api/class-browsercontext#browser-context-new-page) to control their exact life times.**
```

https://playwright.dev/docs/api/class-browser#browser-new-page

lib/png/png.go
berniexie: @nhooyr It was here https://github.com/playwright-community/playwright-go/blob/8e8f670b5fa7ba5365ae4bfc123fea4aac359763/run.go#L92 but I missed that you can disable the Verbose flag to not show that line of text. Implementing that instead

lib/png/png.go
berniexie: it comes back as an interface{} type though? Not sure how to convert it properly

ci/release/template/man/d2.1
berniexie: Ok I try but I"m not sure what the expected output is here

cmd/d2/help.go
berniexie: Yeah but i'm not sure which comment hes asking to repeat. is it the "|" one or the "," one

cmd/d2/watch.go
berniexie: If we do continue, will it continue to run even though they're trying export to ".png" but the png Exporter has died?

ci/release/template/man/d2.1
alixander: same as `--help`. i actually think it should be `|`

cmd/d2/help.go
alixander: let's do `|`

cmd/d2/watch.go
alixander: no ctx.err will be hit

lib/png/png.go
alixander: can u just cast it as string? `pngInterface.(string)`

cmd/d2/main.go
berniexie: So this is svg because the browser only accepts SVGs right now as a render.
```
			w.broadcast(&compileResult{
				SVG: string(b),
			})
```

Should we make it so that it can broadcast PNGs?

alixander: no, the `svg` variable used to also include the postprocess. u changed that but didn't change the return

ci/release/changelogs/next.md
```diff
@@ -3,6 +3,7 @@ For v0.0.99 we focused on X, Y and Z. Enjoy!
 #### Features 💸
 
 - Now you can easily do x, y and z #9999
+- Ability to export to PNGs
```
alixander, line 6: replace the first line. `to PNGs` -> `to PNG`

ci/release/template/man/d2.1
```diff
@@ -21,8 +23,10 @@
 .Nm
 compiles and renders
 .Ar file.d2
-to
+to
```
alixander, line 26: trailing empty space?

cmd/d2/main.go
```diff
@@ -127,19 +127,34 @@ func run(ctx context.Context, ms *xmain.State) (err error) {
...
+		pw, err = png.InitPlaywright()
+		if err != nil {
+			return err
+		}
+		defer func() error {
+			err = pw.Cleanup()
+			if err != nil {
+				return err
+			}
+			return nil
```
alixander, line 141: again, returning an error doesn't do anything. u have a named return parameter, called `err`. if that is nil, then take the result of Cleanup() and set it to that. if it's not nil, then just ignore, there's another error anyway.

lib/png/png.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate png: %w\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", err)
+	}
+
+	pngString := pngInterface.(string)
+	if !strings.HasPrefix(pngString, pngPrefix) {
+		if len(pngString) > 50 {
+			pngString = pngString[0:50] + "..."
+		}
+		return nil, fmt.Errorf("invalid PNG: %q\nplease report this issue here: https://github.com/terrastruct/d2/issues/new", pngString)
```
alixander, line 88: remove all these "report this issue" plz, we'll centralize it in another PR

alixander (changes requested):

cmd/d2/main.go
berniexie: Ohhhhhhhh

alixander (approved):

PR #114, opened 2022-11-17
install: add callout to next steps in terminal display

alixander:
@ejulio-ts missed that he had to do something after install, and I think it won't be uncommon for others to as well. This adds a little callout that should catch the majority of cases.

Also adds a link to report issue if something went wrong

## Before

Screen Shot 2022-11-17 at 11 24 10 AM

## After

Screen Shot 2022-11-17 at 11 30 53 AM

### DIFF
--- a/ci/release/_install.sh
+++ b/ci/release/_install.sh
@@ -318,7 +318,7 @@ uninstall_standalone_d2() {
 
   if [ ! -e "$INSTALL_DIR/d2-$INSTALLED_VERSION" ]; then
     warn "missing standalone install release directory $INSTALL_DIR/d2-$INSTALLED_VERSION"
-    warn "d2 has been installed via some other installation method."
+    warn "d2 must have been installed via some other installation method."
     return 1
   fi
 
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -214,7 +214,7 @@ hide() {
     return
   fi
   cat "$out" >&2
-  exit "$code"
+  return "$code"
```
alixander, line 217: i didn't change these, just wasn't `gen`d before

nhooyr (approved):
✅

ci/release/_install.sh
```diff
@@ -318,7 +318,7 @@ uninstall_standalone_d2() {
 
   if [ ! -e "$INSTALL_DIR/d2-$INSTALLED_VERSION" ]; then
     warn "missing standalone install release directory $INSTALL_DIR/d2-$INSTALLED_VERSION"
-    warn "d2 have been installed via some other installation method."
+    warn "d2 has been installed via some other installation method."
```
nhooyr, line 321: must have

nhooyr (changes requested):

nhooyr (approved):

ejulio-ts (approved):
😅

PR #122, opened 2022-11-21
install.sh: --uninstall implies uninstall tala too

nhooyr:


alixander (approved):

ci/release/_install.sh
```diff
@@ -82,6 +82,7 @@ Flags:
   as for installation. i.e if you used --method standalone you must again use --method
   standalone for uninstallation. With detect, the install script will try to use the OS
   package manager to uninstall instead.
+  note: tala will also be uninstalled if installed.
```
alixander, line 85: actually, can we change this to: "all binary plugins will also be uninstalled, if any"

alixander (changes requested):

ci/release/_install.sh
nhooyr: No, there's no concept of binary plugins in the install script. There's just d2 and tala. We do not need to complicate the docs until necessary.

alixander (approved):

PR #137, opened 2022-11-22
Fix int min/max initialization for 32-bit targets

skeeto:
Without this change, D2 does not compile for 32-bit targets.

### DIFF
--- a/d2target/d2target.go
+++ b/d2target/d2target.go
@@ -26,10 +26,10 @@ type Diagram struct {
 }
 
 func (diagram Diagram) BoundingBox() (topLeft, bottomRight Point) {
-	x1 := int(math.MaxInt64)
-	y1 := int(math.MaxInt64)
-	x2 := int(-math.MaxInt64)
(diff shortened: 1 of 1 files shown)


nhooyr:
Thank you for your contribution @skeeto !

We're figuring out whether we need a CLA before we can begin accepting external contributions. So it might be a little while before we can merge this in.

alixander:
Hi @skeeto, we do require a one-time CLA: https://github.com/terrastruct/d2/blob/master/docs/CONTRIBUTING.md#logistics

It's an email to cla@terrastruct.com with your name and github username that you agree to [Terrastruct's
  CLA](https://terrastruct-site-assets.s3.us-west-1.amazonaws.com/documents/terrastruct_cla.pdf).

skeeto:
Thanks for following up, but I'm not interested in signing anything.

nhooyr:
We respect that. We'll continue discussing the CLA over time to see if it can be eliminated. Thank you for your feedback.

PR #178, opened 2022-11-23
export: Encode SVG images in PNG export.

berniexie:
Uses the code we have in our internal PNG export tool.

Example png export:
out


lib/png/generate_png.js
```diff
@@ -1,12 +1,61 @@
...
+      });
+    });
+    return promise;
+  };
+
+  const loadSVGImages = async () => {
+    const images = tempDiv.querySelectorAll("image");
+    const promises = [];
+    for (const image of images) {
+      const url = image.getAttribute("href");
```
alixander, line 28: these may be local images. `file://`. it doesn't work right now but we'll want to. you should only fetch if it starts with `http`.

alixander (changes requested):
actually i don't think we should fetch in JS at all. fetching from Go will be a server request which bypasses all the bot browser checks, we can log errors easily when an img fetch fails, we can fetch all images concurrently, and if in watch mode, we can cache images to not fetch on every rerender.

alixander:
https://github.com/terrastruct/d2/pull/224

PR #208, opened 2022-11-25
ci(lint): add shell linter - Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR. The beauty of differential scans is that you will get reports only about newly added defects. 

Since this repository has few shell scripts, I think you might find some value in having a shell linter. Differential ShellCheck action is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,23 @@
+name: Differential ShellCheck
+on:
+  pull_request:
+    branches: [ master ]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
+        with:
+          severity: warning
+          token: ${{ secrets.GITHUB_TOKEN }}


nhooyr:
Hi @jamacku!

Thank you for your contribution but I explicitly decided against using shellcheck. I've found it gives false positives or parses complex shell scripts incorrectly.

For the false positives, I don't want to pollute our scripts with comments asking shellcheck to ignore a rule on a given line. And for incorrect parse results, there's nothing I can do except ignore the file entirely.

While I'm not perfect, I'm a very careful programmer that understands the pitfalls of the unix shell well and so my shell code doesn't tend to suffer from the kinds of basic issues that shellcheck detects.

And so I'd rather not use it at all. I wish I could provide specific examples but I've already given shellcheck a shot multiple times over the years and each time stopped for the same reasons.

jamacku:
Hi @nhooyr

Thank you for your comprehensive answer. I completely understand your points and reasoning behind rejecting this PR. I would like to note something here. Even though ShellCheck isn't perfect and many "defects" are related to style preferences, there is, unfortunately, nothing better for linting shell scripts.

To make ShellCheck more developer friendly, you can increase the severity level of defects that will be reported. Additionally, you can start performing differential scans that take into consideration only newly added "defects" (e.g. only added defects with severity "error" will be reported).

I am a maintainer of [legacy project](https://github.com/fedora-sysv/initscripts) with huge amount of "defects" reported by ShellCheck, but I can still use ShellCheck thanks to differential scanning.

You don't have to pollute your scripts with shellcheck directives masking false positives. You will review reports in PR, and if you find them "OK" and merge the PR, you will not see that defect reported again.

I just wanted to explain what Differential ShellCheck GitHub Action does.

Issue #153, opened 2022-11-22
tooling: Embed script as URL?


alixander:
https://news.ycombinator.com/item?id=33710146

Interesting idea. maybe we can have `https://d2.dev/?lksadjlkjadsf` return SVG?

nhooyr:
note: Upgraded version with no rate limits that users can pay for. Requested on discord. https://discord.com/channels/1039184639652265985/1039185315685998693/1044721094222942240

nhooyr:
This ties into our plan to allow rendering D2 files in github repos within their README.

e.g. `d2.dev/github?file=something.svg` and we automatically know which repo it is based on the `Referer` header.

nhooyr:
ooo we need a D2 playground and it ties into this.

see https://www.reddit.com/r/programming/comments/z1qoxc/d2_is_now_open_source_a_new_modern_language_that/ixczyul/

`play.d2lang.com` for the live editor.

`api.d2lang.com` for the API to text -> svg and with files in repos.

nhooyr:
Each example on https://text-to-diagram.com/ then should link to `play.d2lang.com`. Including TALA.

Only the API should require a paid license for TALA.

alixander:
> e.g. d2.dev/github?file=something.svg 

you mean `something.d2` right

nhooyr:
Yup!

YashBhalodi:
Hey guys.

I was fascinated by D2 and wanted to contribute somehow, and I saw this issue. So i decided to work on this.

### Api
An web-api that takes diagram code and returns an svg.
[here](https://github.com/YashBhalodi/d2api) is the link to go server repo serving this API.

### Playground
https://playground-d2-lang.netlify.app/

Currently, I don't have a way to host the golang server, because I literally wrote the "hello world" in go first time just yesterday. I thought maybe I could host it in netlify, but that's not gonna work, and heroku is not free anymore. So if you guys have any suggestion, lemme know.

All of this is just an MVP, so I hope we can collaborate to make things better on both playground and API front.


### Note
Currently, since I haven't hosted API anywhere, you will need to clone the [repo](https://github.com/YashBhalodi/d2api) and do the usual go installation, starting go program stuff. Then the playground will interact with that local server and yep that should work.

alixander:
@YashBhalodi so glad to hear you're interested in helping out! 

[...]

PR #217, opened 2022-11-26
compress + decompress

alixander:
necessary functions for #153 . right now this can be used by integrations. later we can figure out CLI usage

### DIFF
--- a/lib/compress/compress.go
+++ b/lib/compress/compress.go
@@ -1,0 +1,67 @@
+package compress
+
+import (
+	"bytes"
+	"compress/flate"
+	"encoding/base64"
+	"io"
+	"strings"
+
+	"oss.terrastruct.com/d2/d2graph"
+)
+
+var compressionDict = "->" +
+	"<-" +
+	"--" +
+	"<->"
+
+var compressionDictBytes []byte
+
+// Compress takes a D2 script and compresses it to a URL-encoded string
+func Compress(raw string) (string, error) {
+	var b bytes.Buffer
+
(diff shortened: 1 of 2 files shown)


nhooyr (commented):
why do you need flate here? why encode at all? it's just text, you can just put it in a query parameter directly. I'm sure it's less bytes with flate on larger diagrams but for shorter diagrams it will be significantly more, especially with the overhead of base64.

alixander:
> it's just text, you can just put it in a query parameter directly

i don't understand. how do you do this? what happens when u run into a `/` character?

> I'm sure it's less bytes with flate on larger diagrams but for shorter diagrams it will be significantly more

i think practically, it'll be less bytes for the majority of use cases.

nhooyr:
> i think practically, it'll be less bytes for the majority of use cases.

No way. Have you tested this? I very strongly doubt it. Seriously. It makes little sense with the combined overhead.

Try some examples and report back compared to ascii, I'm genuinely curious now lol.

> i don't understand. how do you do this? what happens when u run into a / character?

URL query parameters accept arbitrary bytes. Characters like `/`, ` ` and other disallowed bytes get encoded with percent signs.

See
1. https://pkg.go.dev/net/url#QueryEscape
2. https://pkg.go.dev/net/url#Values
3. https://pkg.go.dev/net/url#Values.Encode

alixander:
resolved offline. "url encoding is extremely verbose. every non representable character becomes 3 bytes"

PR #224, opened 2022-11-26
inline images

### LINKED ISSUE #146: icons: Local images
https://stackoverflow.com/questions/4090712/why-cant-i-do-img-src-c-localfile-jpg

People should be able to, though. Should come up with a good way to allow this.

### LINKED ISSUE #176: exports: PNG exports have to wait for images to load
otherwise:
out

alixander:
Got the method from @berniexie's PR https://github.com/terrastruct/d2/pull/178 (closes #176 )

Just adapted to use Go to fetch images instead of JS.

Also, reads local images and inlines them in the SVG, since localhost cannot access local files. This closes #146


berniexie (approved):
Makes sense to me. Very cool

cmd/d2/main.go
```diff
@@ -204,9 +205,18 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, theme
 	if err != nil {
 		return nil, err
 	}
+	svg, err = imgbundler.InlineLocal(ms, svg)
```
nhooyr, line 208: Put this in a new else of the png check below. No need to run this for png outputs.

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func InlineRemote(ms *xmain.State, in []byte) ([]byte, error) {
+	return inline(ms, in, true)
+}
+
+func inline(ms *xmain.State, in []byte, isRemote bool) ([]byte, error) {
+	svg := string(in)
+
+	imgs := imageRe.FindAllStringSubmatch(svg, -1)
```
nhooyr, line 38: Why not use https://pkg.go.dev/regexp#Regexp.FindAllSubmatch and avoid the extra alloc?

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,136 @@
...
+	"os"
+	"regexp"
+	"strings"
+	"sync"
+	"time"
+
+	"oss.terrastruct.com/d2/lib/xmain"
+)
+
+var imageRe = regexp.MustCompile(`<image href="([^"]+)"`)
```
nhooyr, line 19: I think it's time to parse the svg correctly. I can see someone embedding some HTML code and us screwing it up.

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,136 @@
...
+
+	var wg sync.WaitGroup
+	respChan := make(chan resp)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	wg.Add(len(filtered))
+	for _, img := range filtered {
+		if isRemote {
+			go fetch(ctx, img[0], img[1], respChan)
```
nhooyr, line 57: Rewrite this to avoid passing in `respChan`. Make fetch/read return values and then pass their return values into respChan in an anonymous function here. Also add a `defer wg.Done()` in the anon function.

You only need one anon function, check `isRemote` in just the one.

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,136 @@
...
+
+	go func() {
+		for {
+			select {
+			case resp, ok := <-respChan:
+				if !ok {
+					return
+				}
+				if resp.err != nil {
+					ms.Log.Error.Printf("image failed to fetch: %s", resp.err.Error())
```
nhooyr, line 71:
```suggestion
					ms.Log.Error.Printf("image failed to fetch: %v", resp.err)
```

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,136 @@
...
+				} else {
+					svg = strings.Replace(svg, resp.srctxt, fmt.Sprintf(`<image href="%s"`, resp.data), 1)
+				}
+				wg.Done()
+			}
+		}
+	}()
+
+	wg.Wait()
```
nhooyr, line 80: I don't like this structure. If fetch/read panic you'll end up with this goroutine freezing forever. Likewise with the above respChan listener one.

You need a semaphore here to limit the number of concurrency requests to something reasonable like 16.

Use `defer` on wg.Done **always** without exception. And do it in the actual goroutine itself, never in a secondary one.

Then what you want to do is run the `wg.Wait` in a separate goroutine and have it close the `respChan` channel when all worker goroutines are done.

This does two things. First, it enables you to correctly return the last non nil error instead of just logging it. Or you could use that multierr package, I forget which one we use in the monorepo, use the same one.
Second, this means you correctly return when all responses are finished processing, not when all worker goroutines are finished which is what's happening now.

lib/imgbundler/imgbundler_test.go
```diff
@@ -0,0 +1,185 @@
...
+//go:embed test_png.png
+var testPNGFile []byte
+
+type RoundTripFunc func(req *http.Request) *http.Response
+
+func (f RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f(req), nil
+}
+
+func TestRegex(t *testing.T) {
```
nhooyr, line 29: God bless.

nhooyr: We should always have a set of tests for every regex, seriously.

lib/imgbundler/imgbundler_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"oss.terrastruct.com/cmdlog"
+	"oss.terrastruct.com/xos"
+
+	"oss.terrastruct.com/d2/lib/xmain"
+)
+
+//go:embed test_png.png
+var testPNGFile []byte
+
+type RoundTripFunc func(req *http.Request) *http.Response
```
nhooyr, line 23: this doesn't need to be exported.

lib/imgbundler/imgbundler_test.go
```diff
@@ -0,0 +1,185 @@
...
+		case pngURL:
+			body = string(testPNGFile)
+		default:
+			t.Fatal(req.URL.String())
+		}
+		return &http.Response{
```
nhooyr, line 93: Use `httptest` to craft the response instead.
- https://pkg.go.dev/net/http/httptest#ResponseRecorder.WriteHeader
- https://pkg.go.dev/net/http/httptest#ResponseRecorder.Result
- https://pkg.go.dev/net/http/httptest#ResponseRecorder.WriteString
- https://pkg.go.dev/net/http/httptest#ResponseRecorder.Header

lib/imgbundler/imgbundler_test.go
```diff
@@ -0,0 +1,185 @@
...
+		case pngURL:
+			body = string(testPNGFile)
+		default:
+			t.Fatal(req.URL.String())
```
nhooyr, line 91:
```suggestion
			t.Fatal(req.URL)
```

You need to remember to stop adding `.String()`!!!

nhooyr: You almost never need it unless you're actually operating on the string representation of a type.

lib/imgbundler/imgbundler_test.go
```diff
@@ -0,0 +1,185 @@
...
+	ms := &xmain.State{
+		Name: "test",
+
+		Stdin:  os.Stdin,
+		Stdout: os.Stdout,
+		Stderr: os.Stderr,
+
+		Env: xos.NewEnv(os.Environ()),
+	}
+	ms.Log = cmdlog.Log(ms.Env, os.Stderr)
```
nhooyr, line 171: some day we'll fix this: https://github.com/terrastruct/cmdlog/issues/4

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,136 @@
...
+		respChan <- resp{err: err}
+		return
+	}
+
+	mimeType := http.DetectContentType(data)
+	mimeType = strings.Replace(mimeType, "text/xml", "image/svg+xml", 1)
+
+	enc := base64.StdEncoding.EncodeToString(data)
+
+	respChan <- resp{
```
nhooyr, line 132: each of these sends needs to be async with the context being cancelled. another good reason to centralize the sending on the response channel into the loop as an anon function.

I would also have two separate channels instead, one for errs called `errc` and one for the responses.

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return []byte(svg), nil
+}
+
+var transport = http.DefaultTransport
+
+func fetch(ctx context.Context, srctxt, href string, respChan chan resp) {
+	req, err := http.NewRequestWithContext(ctx, "GET", href, nil)
+	if err != nil {
+
```
nhooyr, line 91: space

nhooyr (changes requested):

lib/imgbundler/imgbundler.go
nhooyr: Although I guess since you need a resp type anyway, using a single channel is fine.

cmd/d2/main.go
alixander: u do need to. InlineLocal and InlineRemote are mutually exclusive

nhooyr: That's what I thought so why is InlineRemote running after InlineLocal? Shouldn't you only need to run one of them given like you said they're mutually exclusive.

lib/imgbundler/imgbundler.go
nhooyr: Nvm, I guess even now you're returning when all worker goroutines are finished but once you move the `wg.Done` into the worker goroutine then you'll definitely need to foreground the respChan goroutine.

cmd/d2/main.go
alixander: u always need to read local images off disk. localhost won't let u access them otherwise.

lib/imgbundler/imgbundler.go
alixander: > You need a semaphore here to limit the number of concurrent requests to something reasonable like 16.

is this just to avoid hogging the user machine resources or is there another reason?

nhooyr: Yea, if they have 100 images, you don't want to fire off 100 requests concurrently. It'll saturate their network connection and make all the requests very slow. I know it's counterinuitive as you'd figure things would work faster if you did it with maximum concurrency but things get bottlenecked and it ends up being slower than limiting and using a semaphore.

cmd/d2/main.go
nhooyr: Ohhh I see, my bad I misunderstood. makes sense.

lib/imgbundler/imgbundler.go
alixander: i'd prefer to delay for another minor version or two. the SVG structure is still changing rapidly, as bugs are being fixed in the render and functionality is added (e.g. latex). it should stabilize much more after this batch of changes.

lib/imgbundler/imgbundler.go
alixander: what's the extra alloc? it looks like it just gives it in a different form ([]bytes), one that I don't want.

nhooyr: Converting from `[]byte` to `string` is an extra alloc. Why can't you just use `[]byte` though?

alixander: cuz in the next loop i url.Parse the string?

lib/imgbundler/imgbundler.go
nhooyr: https://github.com/terrastruct/d2/issues/258

lib/imgbundler/imgbundler.go
nhooyr: Ah right. my b

nhooyr: actually in that case add a `string()` in `url.Parse` and then use `bytes` everywhere else. that way it'll be more clear.

alixander: these are urls. why would it be more clear as bytes?

nhooyr: just because something is utf-8 doesn't mean you immediately turn it into a `string`. there's a real cost. you want to keep things as `[]byte` if you can.

here if you only converted to string for `url.Parse`, it'd be clear to me why you're doing that instead of what you're doing now where you're converting the svg all at once at the beginning.

lib/imgbundler/imgbundler.go
nhooyr: Oh and the semaphore here sould be implemented as a `make(chan struct{}, X)` where X is the number of concurrent requests you want to allow. Then to acquire you'd try to send on the channel (remember to listen on ctx.Done too). To release just read from the channel.

lib/imgbundler/imgbundler.go
alixander: it's not even an svg i'm converting. it's literally an image URL. utf-8 has nothing to do with this. what is this real cost other than style here

nhooyr: how is line 36 not converting the entire svg? i'm asking you to convert just the image URL instead on `url.Parse`.

lib/imgbundler/imgbundler.go
alixander: > First, it enables you to correctly return the last non nil error instead of just logging it.

the logging is intentional. if a user has 20 icons and one breaks, we tell them but render the rest of the diagram

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,134 @@
...
+	defer imgResp.Body.Close()
+	data, err := ioutil.ReadAll(imgResp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	mimeType := http.DetectContentType(data)
+	mimeType = strings.Replace(mimeType, "text/xml", "image/svg+xml", 1)
+
+	enc := base64.StdEncoding.EncodeToString(data)
```
nhooyr, line 117: Don't you want `URLEncoding` here?

nhooyr (changes requested):

lib/imgbundler/imgbundler.go
nhooyr: I don't think that should be default behaviour. That should be under a flag to ignore warnings. I don't want to sift through logs to see if my diagram has rendered correctly. I need to be able to rely on the exit status of the CLI.

lib/imgbundler/imgbundler.go
alixander: ohh i didn't see that FindAllSubmatch accepted a []byte too, only that it returned it in that form. okay, agreed

lib/imgbundler/imgbundler.go
alixander: i think this goes back to "warnings > errors" principle. what if they just want to rapidly iterate and start with some paths that they'll then populate with images?

lib/imgbundler/imgbundler.go
alixander: actually, i still don't. it's just too many string() conversions. not just in url.Parse, but fetch and read take in string too.

nhooyr: Ah, convert the match then and pass around a `urlStr` instead.

lib/imgbundler/imgbundler.go
nhooyr: We've decided warnings or errors on a case by case basis, not as a general principle. I can think of multiple instances where I've requested errors be returned where there were warnings. And I don't think this is the time for a warning, at least not by default. It's much too easy to miss.

They can leave those paths in comments or use a dummy image URL. Or we can handle `--watch` differently and still render as we can't exit with a status code there anyway.

lib/imgbundler/imgbundler.go
alixander: good catch

lib/imgbundler/imgbundler.go
alixander: it's not easy to miss at all.

yo

even if u make the argument that they're not looking at the render for some reason, the logs are also not easy to miss

Screen Shot 2022-11-29 at 10 41 31 AM

lib/imgbundler/imgbundler.go
alixander: ya know, i thought so, but turns out it doesn't work with URLEncoding.

nhooyr: Weird. When you say it doesn't work wdym? Is there an error of some kind? Or just no image being shown.

lib/imgbundler/imgbundler.go
alixander: it's also not convenient (filtering step).

lib/imgbundler/imgbundler.go
alixander: just no image

lib/imgbundler/imgbundler.go
nhooyr: Just keep filtered a `[][]string` and do `[]string{string(img[0]), string(img[1])}` when appending. This way you don't need conversions anywhere else other than this loop.

lib/imgbundler/imgbundler.go
nhooyr: weird idk

lib/imgbundler/imgbundler.go
alixander: u can argue for a lot of unnecessary things if your demographic is unconscious bodies.

and no i think that compromise is even worse. an error should not bail out saving the errored version

can u think of software that produces assets that refuses to work given broken linked assets? i strongly suspect this is not behavior users will desire in practice.

nhooyr: > u can argue for a lot of unnecessary things if your demographic is unconscious bodies.

I clearly meant not fully conscious. Not unconscious. Or frankly even newbs.

For example Go doesn't by default just give you a warning if you have an unused variable/import, it errors. Now Go isn't the best example as you can't disable these often annoying ass warnings, but I still strongly prefer it being the default behaviour.

> and no i think that compromise is even worse. an error should not bail out saving the errored version

Ok well what about the `--watch` exception? I think that's perfect.

> can u think of software that produces assets that refuses to work given broken linked assets? i strongly suspect this is not behavior users will desire in practice.

Software package managers. even webpack i think? At least if you're bundling the asset in. Yea so that's the primary difference. We're bundling the asset here, we're not just adding a link.

Oh that reminds me we wanted a `-bundle` flag to bundle in remote links for svgs too right?

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,133 @@
...
+
+func inline(ms *xmain.State, svg []byte, isRemote bool) ([]byte, error) {
+	imgs := imageRe.FindAllSubmatch(svg, -1)
+
+	var filtered [][]string
+	for _, img := range imgs {
+		u, err := url.Parse(string(img[1]))
+		isRemoteImg := err == nil && strings.HasPrefix(u.Scheme, "http")
+		if isRemoteImg == isRemote {
+			filtered = append(filtered, []string{string(img[0]), string(img[0])})
```
nhooyr, line 43:
```suggestion
			filtered = append(filtered, []string{string(img[0]), string(img[1])})
```

lib/imgbundler/imgbundler.go
nhooyr: > yes obviously i did too lol

My man you literally used the word "unconscious body".

> i can see the argument from this angle. i wonder if we should make a --prod flag instead though, because when they're watching with png output, they're not necessarily trying to get production build every time. such a flag would minify svg, error on any broken assets, perhaps do other things.

No, the default should be prod mode. Dev should be the exception. When you compile d2 without any flags, you should know you're getting the safest sanest configuration.

We could add a `--dev` flag but we already have `--watch`. And that's why I'm suggesting reusing `--watch` for now, as that's the only time you'll actually want to ignore such errors and continue rendering.

alixander: okay. that's fair. doesn't error on watch. errors without. yes?

nhooyr: Yup.

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,154 @@
...
+		return "", err
+	}
+
+	client := &http.Client{Transport: transport}
+	imgResp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer imgResp.Body.Close()
+	data, err := ioutil.ReadAll(imgResp.Body)
```
nhooyr, line 129: use https://pkg.go.dev/net/http#MaxBytesReader to limit the response body size to 1 meg.

it's undocumented right now but you can use it without a ResponseWriter by passing nil as the response writer. there's an open issue on the go repo to add docs for it so we're good to use it knowing it will be a supported use case.

nhooyr: `1 << 20` is 1 meg

alixander: 1mb? that's a small cap. how about 20?

nhooyr: fair enough. how about 32 megs to keep it a nice power of 2? `2^25 = 33_554_432`

lib/imgbundler/imgbundler.go
ejulio-ts: I find it a bit awkward that an offline tool fails, when it could succeed, if there's no internet connection.
I'd say that we're more into draft/print modes than anything else. While working on my diagram, I might just want to draft it and even replace images with just boxes (to avoid rendering delays, or even downloading the same image multiple times) and get something quickly..

@nhooyr, even using your presentation example. I would be pissed off if my deck software failed to help me build a presentation because it couldn't get an image. Even more if the image source is offline (like the website is down, or some other intermediate network issue). Like, latex will render the PDF and just show a broken image thumbnail. 

My 2 cents here is for "make it work and if you need images strictly, it must be flagged"

alixander:
Screen Shot 2022-11-29 at 1 31 47 PM

lib/imgbundler/imgbundler.go
nhooyr: we're not in disagreement @ejulio-ts 

I want the exit code to be non zero if images fail to be fetched but I'm ok with writing partial output.

ejulio-ts: ok, maybe I missed something on the long read 😂

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,164 @@
...
+	"time"
+
+	"go.uber.org/multierr"
+	"oss.terrastruct.com/xdefer"
+
+	"oss.terrastruct.com/d2/lib/xmain"
+)
+
+// 32 MB
+var max_img_size int64 = 33_554_432
```
nhooyr, line 24:
```suggestion
const maxImageSize int64 = 1 << 25 // 33_554_432
```

nhooyr (changes requested):

lib/imgbundler/imgbundler.go
nhooyr: haha no worries, it was my suggested compromise above. But I conceded to Alex as I thought he felt strongly it shouldn't be so based on this response:

> and no i think that compromise is even worse. an error should not bail out saving the errored version

lib/imgbundler/imgbundler_test.go
```diff
@@ -0,0 +1,221 @@
...
+		t.Fatal("no svg image inserted")
+	}
+	if !strings.Contains(string(out), "image/png") {
+		t.Fatal("no png image inserted")
+	}
+
+	// Test almost too large response
+	transport = roundTripFunc(func(req *http.Request) *http.Response {
+		respRecorder := httptest.NewRecorder()
+		bytes := make([]byte, maxImageSize-1)
```
nhooyr, line 124:
```suggestion
		bytes := make([]byte, maxImageSize)
```

nhooyr (changes requested):

cmd/d2/main.go
```diff
@@ -211,9 +212,26 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, theme
 	if err != nil {
 		return nil, err
 	}
+	svg, err = imgbundler.InlineLocal(ctx, ms, svg)
+	if err != nil {
+		if !isWatching {
+			ms.Log.Error.Printf("missing/broken remote image(s), writing partial output: %v", err)
```
nhooyr, line 218:
```suggestion
			ms.Log.Error.Printf("missing/broken local image(s), writing partial output: %v", err)
```

cmd/d2/main.go
```diff
@@ -211,9 +212,26 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, theme
...
+	}
 
 	out := svg
 	if filepath.Ext(outputPath) == ".png" {
+		svg, err = imgbundler.InlineRemote(ctx, ms, svg)
+		if err != nil {
+			if !isWatching {
+				ms.Log.Error.Printf("missing/broken remote image(s), writing partial output: %v", err)
+			} else {
+				ms.Log.Debug.Printf("ignoring missing/broken remote image(s) in watch mode: %v", err)
```
nhooyr, line 231: This should be an Error too. In fact, I don't think the message needs to change at all anymore between watch/nowatch, just always log the error.

cmd/d2/main.go
```diff
@@ -224,6 +242,13 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, theme
 	if err != nil {
 		return nil, err
 	}
+
+	// Missing/broken images are fine during watch mode, as the user is likely building up a diagram.
+	// Otherwise, the assumption is that this diagram is building for production, and broken images are not okay.
+	if !isWatching && ms.Log.Nerrors() > 0 {
+		os.Exit(1)
```
nhooyr, line 249: Never do `os.Exit` outside of `main`, never! Here be dragons 🐉 😈 👹 
Seriously, zero exceptions, absolutely zero.

Instead return `xmain.ExitErrorf(1, "errors logged while rendering, partial output written to %v", outputPath)`

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,163 @@
...
+
+	go func() {
+		wg.Wait()
+		close(respChan)
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			ms.Log.Debug.Printf("there")
```
nhooyr, line 100:
```suggestion
```

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,163 @@
...
+			return nil, fmt.Errorf("failed to wait for imgbundler workers: %w", ctx.Err())
+		case <-time.After(time.Second * 5):
+			ms.Log.Info.Printf("fetching images...")
+		case resp, ok := <-respChan:
+			if !ok {
+				return svg, err
+			}
+			if resp.err != nil {
+				err = multierr.Combine(err, resp.err)
```
nhooyr, line 109: Remove `multierr` and log the errors. Otherwise `cmdlog.Logger.Nerrors` will always return zero.

lib/imgbundler/imgbundler.go
```diff
@@ -0,0 +1,163 @@
...
+func fetch(ctx context.Context, href string) (string, error) {
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, "GET", href, nil)
+	if err != nil {
+		return "", err
+	}
+
+	client := &http.Client{Transport: transport}
```
nhooyr, line 128: Make `*http.Client` the global and replace `client.Transport` in tests instead. Will mean we're keeping cookies between requests which is what we want. Transport already reuses connections between requests.

nhooyr (changes requested):

lib/imgbundler/imgbundler.go
alixander: 🤦

lib/imgbundler/imgbundler.go
alixander: i changed the `return svg, nil` to `return svg, err` above.

nhooyr: Oh nice catch but I'm saying we get rid of multierr to log all errors as they come in instead of waiting for all images to be fetched before we log errors. Otherwise there's no point in using `Nerrors`, you could just set a bool in compile() when the imgBundler returns an error and then return an xmain.ExitError with with status 1 if said bool is true at the end of the function.

cmd/d2/main.go
```diff
@@ -224,6 +234,13 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, theme
 	if err != nil {
 		return nil, err
 	}
+
+	// Missing/broken images are fine during watch mode, as the user is likely building up a diagram.
+	// Otherwise, the assumption is that this diagram is building for production, and broken images are not okay.
+	if !isWatching && ms.Log.Nerrors() > 0 {
+		xmain.ExitErrorf(1, "errors logged while rendering, partial output written to %v", outputPath)
```
nhooyr, line 241:
```suggestion
		return nil, xmain.ExitErrorf(1, "errors logged while rendering, partial output written to %v", outputPath)
```

nhooyr (changes requested):

lib/imgbundler/imgbundler.go
alixander: hmmm, then this whole thing just doesn't return error at all?

alixander: nvm

alixander: actually yeah, then there'd be no point in returning an error. which is odd. i think the current way is fine. Nerrors is just checking > 0 anyway, so it's all the same. yes it's the same as setting a bool, but it's cleaner, no extra var

nhooyr: I don't see why that's odd? It means the CLI is more responsive.

Or actually, yea you ought to return a boolean here then and not even rely on `Nerrors`, I'm going to remove that. It's not cleaner, it's an unnecessary/nonlocal/magic API.

i.e return an ok bool from this function when the svg is totally good and no images errored and return false when at least one image failed. and then use that bool in compile to return the exit error.

alixander: ok sure. but why return a bool? why not just return the error like before?

nhooyr: I figure you don't need to, you're going to log that error, you're just going to check if it exists so a simple boolean suffices.

But now that I think about it, just return the `xmain.ExitError` here and then in `compile()` return the return error of `imgbundler.Inline<whatever>` after writing the output. That's perfect.

berniexie:
Can we have the bundle flag also inline all the remote images?

alixander:
merging for now, @nhooyr will followup

alixander:
> Can we have the bundle flag also inline all the remote images? Or create a separate flag

is this for Obsidian integration? they don't let SVGs load remote images?

nhooyr:
> is this for Obsidian integration? they don't let SVGs load remote images?

No I don't know anything about Obsidian. literally nothing. we just talked about it once, creating standalone output svgs with zero linked assets.

alixander:
bro what, i was asking bernard lol

nhooyr:
I had a comment above before him about literally the exact same thing!

Didn't even realize he also commented lmfao

nhooyr:
Actually idk if I did comment or i was thinking about it and was going to open an issue later lol. yea cause we already have a bundle flag, so we should listen to it for InlineRemote.

alixander:
yeah for context we're testing obsidian plugins and imgs aren't working rn. we're not sure if obsidian embedded browser has something to prevent remote imgs from loading in svg's or something. in which case, the bundle flag working for svg's will be a blocker for obsidian plugin

nhooyr:
ok i'll throw it in my follow up

PR #229, opened 2022-11-27
render: Latex support

### LINKED ISSUE #135: syntax: Support latex
https://github.com/mermaid-js/mermaid/issues/1387

alixander:
the language tag `latex` and `tex` will now render LaTeX (or rather, the MathJax flavor of it)

Screen Shot 2022-11-27 at 7 54 14 PM

Docs: https://github.com/terrastruct/d2-docs/pull/15

closes #135


d2renderers/d2latex/latex.go
```diff
@@ -0,0 +1,79 @@
...
+	wf, err := strconv.ParseFloat(wEx, 64)
+	if err != nil {
+		return 0, 0, fmt.Errorf("svg parsing failed for latex: %v", svg)
+	}
+	hf, err := strconv.ParseFloat(hEx, 64)
+	if err != nil {
+		return 0, 0, fmt.Errorf("svg parsing failed for latex: %v", svg)
+	}
+
+	return int(math.Ceil(wf * 8)), int(math.Ceil(hf * 8)), nil
```
ejulio-ts, line 78: Why `* 8` here? 🤔

d2renderers/d2latex/latex.go
```diff
@@ -0,0 +1,79 @@
...
+//go:embed polyfills.js
+var polyfillsJS string
+
+//go:embed setup.js
+var setupJS string
+
+//go:embed mathjax.js
+var mathjaxJS string
+
+var svgRe = regexp.MustCompile(`<svg[^>]+width="([0-9\.]+)ex" height="([0-9\.]+)ex"[^>]+>`)
```
ejulio-ts, line 25: Maybe just an example that it matches `width=".." height=".."`

ejulio-ts: is it true that `width` will always come before `height` and nothing will be in between?

d2renderers/d2latex/latex.go
```diff
@@ -0,0 +1,79 @@
...
+	if len(dims) != 1 || len(dims[0]) != 3 {
+		return 0, 0, fmt.Errorf("svg parsing failed for latex: %v", svg)
+	}
+
+	wEx := dims[0][1]
+	hEx := dims[0][2]
+
+	wf, err := strconv.ParseFloat(wEx, 64)
+	if err != nil {
+		return 0, 0, fmt.Errorf("svg parsing failed for latex: %v", svg)
```
ejulio-ts, line 71: not `sag parsing failed`, `float parsing failed` and then you can use `wEx` and `hEx` or you can get rid of them

d2renderers/d2latex/latex_test.go
```diff
@@ -0,0 +1,30 @@
...
+		if err := xml.Unmarshal([]byte(svg), &xmlParsed); err != nil {
+			t.Fatalf("invalid SVG: %v", err)
+		}
+	}
+}
+
+func TestRenderError(t *testing.T) {
+	_, err := Render(`\frac{1}{2}`)
+	if err == nil {
+		t.Fatal(err)
```
ejulio-ts, line 28: what will it log if `err` is `nil` here? 
also, you can make it an one liner `if _, err := ..`

e2etests/stable_test.go
```diff
@@ -961,6 +961,31 @@ beta: {
 alpha -> beta: gamma {
 	style.font-color: green
 }
+`,
+		},
+		{
+			name: "latex",
+			script: `a: |latex
+\\Huge{\\frac{\\alpha g^2}{\\omega^5} e^{[ -0.74\\bigl\\{\\frac{\\omega U_\\omega 19.5}{g}\\bigr\\}^{\\!-4}\\,]}}
```
ejulio-ts, line 969: why does it need to be `\\`?
I'd expect `\\` to render the actual `\` and not as a prefix to a latex command

ejulio-ts (changes requested):
Edges are handled
```
a: |latex
x = \\frac{1}{2}
|

b: |latex
e = mc^2
|

a -> b: |latex
e = mc^2
|
```

If we're not going to handle them, we should show an error
If we're going to handle it in another PR, totally fine

I'm used to having `latex` inside markdown with `$$` tags `mathjax`, but it doesn't seem we support it right now. I get that t it is probably out of scope for this PR, so just letting you know inc are you want to open an issue

e2etests/stable_test.go
alixander: https://github.com/terrastruct/d2-docs/pull/15/files#diff-d23c28129c5e819c092f20bfcbf291203169d455eabed54a3498779040e06b20R36

d2renderers/d2latex/latex_test.go
alixander: oh right lol

ejulio-ts:
Don't you need to add some `NOTICE` for `mathjax`?

alixander:
>  Don't you need to add some NOTICE for mathjax?

even better -- the attribution is on the file itself (top)

d2renderers/d2latex/latex.go
alixander: i haven't verified in MathJax's source code, but I see no reason why it'd change

d2renderers/d2latex/latex.go
alixander: float parsing must've failed because svg parsing failed though, it's a more informative error statement

ejulio-ts (approved):

d2renderers/d2svg/d2svg.go
```diff
@@ -701,17 +702,27 @@ func drawShape(writer io.Writer, targetShape d2target.Shape) error {
...
 		case d2target.ShapeText:
-			render, err := textmeasure.RenderMarkdown(targetShape.Label)
-			if err != nil {
-				return err
+			if targetShape.Language == "latex" {
+				render, err := d2latex.Render(targetShape.Label)
+				if err != nil {
+					return err
+				}
+				fmt.Fprintf(writer, `<g transform="translate(%f %f)" style="opacity:%f">`, box.TopLeft.X, box.TopLeft.Y, targetShape.Opacity)
```
gavin-ts, line 710:
```suggestion
				fmt.Fprintf(writer, `<g transform="translate(%f, %f)" style="opacity:%f">`, box.TopLeft.X, box.TopLeft.Y, targetShape.Opacity)
```

alixander: that's not right: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#translate

nhooyr (commented):
nice

PR #238, opened 2022-11-28
Fix trailing whitespace

nhooyr:
cc @gavin-ts


gavin-ts:
is your editor complaining about the trailing whitespace in these tests? why am I cc'd? add a PR description?

alixander:
this was prompted by me. we generally want to not have trailing whitespaces, maybe even include that in the autofmter, and my editor would default remove it. so when we actually want trailing whitespace, it's better to be clear in a visible way

alixander:
(cc'd b/c it's your test)

alixander (approved):

Issue #99, opened 2022-11-16
layout, syntax: Sequence diagrams


alixander:
i can't find an open-source standalone renderer for sequence diagrams. There's some that are integrated into other tools, but they're coupled with their tool. Might have to build one from scratch, which shouldn't be too hard.

Will probably also have to modify syntax

alixander:
work has begun: https://github.com/terrastruct/d2/pull/219

PR #245, opened 2022-11-28
syntax: shape: sequence

alixander:
This enables running sequence layout engine only on sequence shapes.

Supports #99


ejulio-ts (approved):
👌 
Just wondering if `sequence_diagram` (as `sql_table`) would be better.
Not sure if we'll ever have an actual `sequence` shape.

alixander:
oh that's a good point. yeah i think sequence_diagram more clear for scanning

nhooyr (approved):

PR #251, opened 2022-11-29
layout: `direction` keyword

alixander:
Specify `horizontal` or `vertical`.

dagre and ELK both only support a single top-level orientation. We can support multiple orientations in TALA.

Dagre makes this clear in their docs that it's just a top level setting.
There's a few issues asking this from ELK: https://github.com/kieler/elkjs/issues/26

It's also not supported in Graphviz, but people have been able to finangle the fuck out of it with invisible nodes and edges, customized individually to the graph: https://stackoverflow.com/questions/7777722/top-down-subgraphs-left-right-inside-subgraphs .

- this PR also specifies ELK to render top-down by default to be consistent with dagre.

Closes #240


d2compiler/compile_test.go
```diff
@@ -1521,6 +1521,41 @@ dst.id <-> src.dst_id
...
+			assertions: func(t *testing.T, g *d2graph.Graph) {
+				diff.AssertStringEq(t, "horizontal", g.Root.Attributes.Orientation.Value)
+			},
+		},
+		{
+			name: "default_orientation",
+
+			text: `x`,
+			assertions: func(t *testing.T, g *d2graph.Graph) {
+				diff.AssertStringEq(t, "vertical", g.Objects[0].Attributes.Orientation.Value)
```
nhooyr, line 1537: why is this vertical but root above is horizontal by default?

nhooyr (commented):
should we instead support values like `up-down`, `down-up`, `left-right` and `right-left`?

alixander:
> should we instead support values like up-down, down-up, left-right and right-left?

oh interesting, for right-to-left languages like japanese/chinese, i can see that. maybe it should just be called direction then

alixander:
nvm, direction feels like it should be a property from node to another node. for a property on a parent controlling its children, orientation still sounds more right

d2compiler/compile_test.go
nhooyr: lol my b

nhooyr:
I prefer direction now, I think it matches the constants better. I don't share your concern.

alixander:
ok good to know, maybe just me then. as mentioned, +1 to direction being a slightly simpler vocab than orientation

e2etests/stable_test.go
```diff
@@ -986,6 +986,20 @@ sugar -> c
...
 
 c -> solution: we get
+`,
+		},
+		{
+			name: "direction",
+			script: `a -> b -> c -> d -> e
+b: {
+  direction: right
+	1 -> 2 -> 3 -> 4 -> 5
```
nhooyr, line 996: indent

nhooyr (changes requested):

nhooyr (approved):

PR #259, opened 2022-11-29
Sequence diagrams: activation boxes

ejulio-ts:
Supports #99 

- Adds support for activation boxes "over" the objects lifeline
  -  "over" because right now we don't have a way to control the SVG drawing order and edges are rendered after shapes. So I'll need to follow-up with another PR to fix that
- Activation boxes can be as nested as you want and as they get more nested, their `width` grows to show some difference from its parent
  - horizontal growth seemed better than padding as, otherwise, edges could end up overlapping the parent activation box (see examples below for nested sequential edges)
- The width is constant and only changes when nesting
- The height is based on the first/last edges or the min/max children position
  - If there's only one edge, the height is a constant factor then, or the children height + pad
  - The children must be considered because of nesting (see examples below)
- The syntax is just `actor.activation_box`
- activation boxes are always rectangles and have no labels (though, we could rethink the labels decision)
- Added a basic test to check the basic properties. Testing the nested properties seemed too much for this and later on we'll have e2e tests, so left it out for now

Examples

```
scorer.t -> itemResponse.t: getItem()
scorer.t <- itemResponse.t: item

scorer.t -> item.t1: getRubric()
scorer.t <- item.t1: rubric

scorer.t -> essayRubric.t: applyTo(essayResp)
itemResponse -> essayRubric.t.c
essayRubric.t.c -> concept.t: match(essayResponse)
scorer <- essayRubric.t: score

scorer.t -> itemOutcome.t1: new
scorer.t -> item.t2: getNormalMinimum()
scorer.t -> item.t3: getNormalMaximum()

scorer.t -> itemOutcome.t2: setScore(score)
scorer.t -> itemOutcome.t3: setFeedback(missingConcepts)
``` 
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (1)

```
scorer -> itemResponse.a
itemResponse.a -> item.a.b
item.a.b -> essayRubric.a.b.c
essayRubric.a.b.c -> concept.a.b.c.d
item.a -> essayRubric.a.b
concept.a.b.c.d -> itemOutcome.a.b.c.d.e

scorer.abc -> item.a

itemOutcome.a.b.c.d -> concept.a.b.c
itemOutcome.a.b.c -> essayRubric.a.b
itemOutcome.a.b -> item.a
itemOutcome.a -> itemResponse
itemOutcome.a -> scorer.abc
```
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (2)

Edge cases (not sure how to handle properly)

same as above, except for the two `# new edges`
```
scorer -> itemResponse.a
itemResponse.a -> item.a.b
item.a.b -> essayRubric.a.b.c
essayRubric.a.b.c -> concept.a.b.c.d
item.a -> essayRubric.a.b
concept.a.b.c.d -> itemOutcome.a.b.c.d.e

scorer.abc -> item.a

itemOutcome.a.b.c.d -> concept.a.b.c
itemOutcome.a.b.c -> essayRubric.a.b
itemOutcome.a.b -> item.a

# new edge
concept -> itemOutcome 

itemOutcome.a -> itemResponse
itemOutcome.a -> scorer.abc

# new edge
itemOutcome.a.b.c.d.e -> scorer
```
image

The edges are referencing activation boxes that got hidden because the most nested child expanded over them.
As activation boxes should show a sort lifespan of something, I wonder if this can happen in real life examples. Anyway, it's here for discussion..


alixander (commented):
can we add a litttle padding to the arrows so they don't point right on the outer edge of the box corners? i want like... inner corner + 1px or something?

also do people prefer the term activation box? sounds like some UML bs. almost prefer just `span`.

ejulio-ts:
according to https://www.uml-diagrams.org/sequence-diagrams-reference.html
it is `execution` and `activation` is more informal

also, checking the examples here https://www.uml-diagrams.org/sequence-diagrams-reference.html and thinking about `activation` I don't think there'll be an edge on the left of an activation..

Mermaid, lucid chart, plant uml also use `activation`

ejulio-ts:
_Users_juliobatista_Projects_terrastruct_d2_sequence svg

ejulio-ts:
boxes grow by `5` on each direction and edges are padded by `5`
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (1)

gavin-ts:
> boxes grow by `5` on each direction and edges are padded by `5` _Users_juliobatista_Projects_terrastruct_d2_sequence svg (1)

I think the vertical padding is good but I think it makes more sense to have the arrows go all the way to the box.

d2layouts/d2sequence/layout.go
```diff
@@ -4,90 +4,135 @@ import (
...
+		objectRank:     make(map[*d2graph.Object]int),
+		objectDepth:    make(map[*d2graph.Object]int),
+		minEdgeRank:    make(map[*d2graph.Object]int),
+		maxEdgeRank:    make(map[*d2graph.Object]int),
+		edgeYStep:      MIN_EDGE_DISTANCE,
+		actorXStep:     MIN_ACTOR_DISTANCE,
+		maxActorHeight: 0.,
+	}
+
+	sd.init()
```
gavin-ts, line 28: Is there a reason we want this split setup over something like `sd := NewSequenceDiagram(g)` instead? will these values be configured regularly? the maps seem like they can be in `init` unless there is a scenario where we want to preload them with values before calling `init`.

ejulio-ts: I just decided there was no need for `NewSequenceDiagram` right now as it is only used in `Layout`.
Maybe, later on when I work on having them as parts in bigger diagrams I 'll put this in another function to have proper initialization

alixander (commented):
we already break away from styles/aesthetics of sequence diagrams in UML, i don't think deviating from their vernacular is sacrilege, especially if we don't think it makes sense. 

as the target demographic user myself, activation is just a meaningless word to me for this. it makes me feel like i'm missing something or using it incorrectly, when rly it's just... spans, like in tracing.

d2layouts/d2sequence/constants.go
```diff
@@ -3,7 +3,22 @@ package d2sequence
...
+const ACTIVATION_BOX_WIDTH = 20.
+
+// small pad so that edges don't touch lifelines and activation boxes
+const ACTIVATION_BOX_EDGE_PAD = 5.
+
+// as the activation boxes start getting nested, their size grows
+const ACTIVATION_BOX_DEPTH_GROW_FACTOR = 10.
+
+// when a activation box has a single edge
+const DEFAULT_ACTIVATION_BOX_HEIGHT = MIN_EDGE_DISTANCE / 2.
```
alixander, line 24: should call this MIN

d2layouts/d2sequence/layout.go
```diff
@@ -105,3 +150,129 @@ func addLifelineEdges(g *d2graph.Graph, actors []*d2graph.Object, yStep float64)
...
+	}
+}
+
+func (sd *sequenceDiagram) getEdgeY(rank int) float64 {
+	// +1 so that the first edge has the top padding for its label
+	return ((float64(rank) + 1.) * sd.edgeYStep) + sd.maxActorHeight
+}
+
+func (sd *sequenceDiagram) isActor(obj *d2graph.Object) bool {
+	return obj.Parent == sd.graph.Root
```
alixander, line 277: this isn't right. u can have a very nested sequence diagram.

d2layouts/d2sequence/layout.go
```diff
@@ -4,90 +4,135 @@ import (
...
+	copy(queue, sd.graph.Root.ChildrenArray)
+	for len(queue) > 0 {
+		obj := queue[0]
+		queue = queue[1:]
+
+		if sd.isActor(obj) {
+			sd.actors = append(sd.actors, obj)
+			sd.objectRank[obj] = len(sd.actors)
+			sd.objectDepth[obj] = 0
+			sd.maxActorHeight = math.Max(sd.maxActorHeight, obj.Height+VERTICAL_PAD)
```
alixander, line 72: add vertical_pad after loop

d2layouts/d2sequence/layout.go
```diff
@@ -4,90 +4,135 @@ import (
...
+		} else {
+			// activations boxes are always rectangles and have no labels
+			obj.Attributes.Label = d2graph.Scalar{Value: ""}
+			obj.Attributes.Shape = d2graph.Scalar{Value: shape.SQUARE_TYPE}
```
alixander, line 76: we should enforce this in the compiler, help users out with error messages. https://github.com/terrastruct/d2/issues/268

d2layouts/d2sequence/layout.go
```diff
@@ -4,90 +4,135 @@ import (
...
+	activations []*d2graph.Object
+
+	// can be either actors or activation boxes
+	objectRank  map[*d2graph.Object]int
+	objectDepth map[*d2graph.Object]int
+
+	// keep track of the first and last edge of a given actor
+	// needed for activation boxes
+	minEdgeRank map[*d2graph.Object]int
+	maxEdgeRank map[*d2graph.Object]int
```
alixander, line 51: can u define rank and depth in comments

alixander (changes requested):

d2layouts/d2sequence/layout.go
ejulio-ts: not right now..

d2layouts/d2sequence/layout.go
ejulio-ts: ok, will get it..

d2layouts/d2sequence/layout.go
```diff
@@ -4,90 +4,140 @@ import (
...
+
+	edges  []*d2graph.Edge
+	actors []*d2graph.Object
+	spans  []*d2graph.Object
+
+	// can be either actors or span boxes
+	// rank: left to right position of actors/span boxes
+	objectRank map[*d2graph.Object]int
+	// depth: the nested levels of a given actor/span
+	objectDepth map[*d2graph.Object]int
```
alixander, line 48: can you just use this? https://github.com/terrastruct/d2/blob/27a6e219a4c92726dfc40c05a3837076ed208cea/d2graph/d2graph.go#L386

d2layouts/d2sequence/layout.go
```diff
@@ -4,90 +4,140 @@ import (
...
+
+type sequenceDiagram struct {
+	graph *d2graph.Graph
+
+	edges  []*d2graph.Edge
+	actors []*d2graph.Object
+	spans  []*d2graph.Object
+
+	// can be either actors or span boxes
+	// rank: left to right position of actors/span boxes
```
alixander, line 45: u mean index?

alixander (changes requested):

alixander (changes requested):
span boxes -> spans

d2layouts/d2sequence/layout.go
ejulio-ts: Indeed.. 😓

d2layouts/d2sequence/layout.go
ejulio-ts: To some extent.
Like, `spans` have the same rank as their parents.
Index gets some ambiguity with the array where they are stored.

d2layouts/d2sequence/layout.go
ejulio-ts: Though, I'd rather have this here as `Level()` is a recursive call and here I just keep the state in ~3 places I need to use it

alixander: nah code cleanliness over performance here, it's all negligible

d2layouts/d2sequence/constants.go
```diff
@@ -3,7 +3,22 @@ package d2sequence
...
 const MIN_EDGE_DISTANCE = 100.
+
+// default size
+const SPAN_WIDTH = 20.
+
+// small pad so that edges don't touch lifelines and spans
+const SPAN_EDGE_PAD = 5.
+
+// as the spans start getting nested, their size grows
+const SPAN_BOX_DEPTH_GROW_FACTOR = 10.
```
alixander, line 21: .

alixander (changes requested):

alixander (approved):

PR #266, opened 2022-11-29
install.sh: Improve dry run message and add short flag -d

nhooyr:


alixander (approved):
can u add a screenshot? and i think this worth going into release notes, even if technically release-agnostic

nhooyr:
Screenshot 2022-11-29 at 2 32 04 PM

alixander (approved):

PR #267, opened 2022-11-29
Render priority

ejulio-ts:
supports #99 followup from #259 

allow to specify the rendering order of the graph objects.
`RenderPriority` was chosen as it gives the idea of a priority queue.
`Order`/`Sequence` would give the idea of unique and incremental numbers which would be hard to control. This way, it is simply a priority queue, drawing in ascending order.

No changes in SVGs, just `json` adding the new property


d2graph/d2graph.go
```diff
@@ -78,6 +78,8 @@ type Object struct {
 	ChildrenArray []*Object          `json:"-"`
 
 	Attributes Attributes `json:"attributes"`
+
+	RenderPriority *int `json:"renderPriority",omitempty`
```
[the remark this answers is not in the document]
ejulio-ts: I guess it goes in the same direction as in the description.
It gives the impression that it should be unique and sequential, while it doesn't need to

nhooyr:
@ejulio-ts what's a d2 example of where this would be necessary?

d2exporter/export.go
```diff
@@ -16,13 +17,16 @@ func Export(ctx context.Context, g *d2graph.Graph, themeID int64) (*d2target.Dia
 	diagram := d2target.NewDiagram()
 
 	diagram.Shapes = make([]d2target.Shape, len(g.Objects))
+	highestObjectPriority := 0
```
alixander, line 20: don't mix terms. `maxObjectZIndex`

alixander (changes requested):
i was going to say we don't need full blown z-index and could do something simpler, but....maybe one day that could be useful. like we allow users to lock in some objects, and they want to overlap some and specify a z-index? i could see that. 

i'm not sure about z-index being globally unique like u have it. like what if u just made all objects at 0 and edges at 1? and then for those special boxes u want over edges, u just bump them. otherwise if i want an object to go over all edges, i gotta find which z-index that is.

ejulio-ts:
@alixander they are not unique, see my previous discussion with @gavin-ts (I want to avoid having them unique).
I set different values using `Level` as it was before, so many children can have the same `ZIndex` and all edges have same value which is `max(object ZIndex) + `.
The objects must have different `ZIndex` so that containers are rendered before the descendants and that was the code already, just changed to include the edges.

ejulio-ts:
What was the alternative you had in mind?
This the index here is just at export and we don't need to expose it.
It is just a clue of the graph to the rendering engine about the order of the objects.

alixander:
> I set different values using Level as it was before, so many children can have the same ZIndex and all edges have same value which is max(object ZIndex) + .
The objects must have different ZIndex so that containers are rendered before the descendants and that was the code already, just changed to include the edges.

but u can just have all objects and edges as 0, and then for each unique z-index, do the current rendering order (that is, objects in order they appear in []d2graph.Object and then edges in order they appear in []d2graph.Edge).

that's how css works. sort first by z-index, then sort by order defined.

ejulio-ts:
> but u can just have all objects and edges as 0, and then for each unique z-index, do the current rendering order (that is, objects in order they appear in []d2graph.Object and then edges in order they appear in []d2graph.Edge).


I see it works for `Object` and `Edge` as I can control the order I insert them and I can control sorting with `SortStable`.
Though, I don't think it will work for containers and descendants.
If it was true, we wouldn't need the previous code to render the containers first.

https://github.com/terrastruct/d2/blob/master/d2renderers/d2svg/d2svg.go#L983-L1000

ejulio-ts:
there you go @alixander 
```
// sortObjects sorts all diagrams objects (shapes and connections) in the desired drawing order
 // the sorting criteria is:
 // 1. zIndex, lower comes first
 // 2. two shapes with the same zIndex are sorted by their level (container nesting), containers come first
 // 3. two shapes with the same zIndex and same level, are sorted in the order they were exported
 // 4. shape and edge, shapes come first
```

added a test case just to make it easier to check.
also, by default ZIndex is `0` everywhere then

d2renderers/d2svg/d2svg.go
```diff
@@ -1010,6 +1015,34 @@ func Render(diagram *d2target.Diagram) ([]byte, error) {
...
+func sortObjects(allObjects []d2target.DiagramObject) {
+	sort.SliceStable(allObjects, func(i, j int) bool {
+		// first sort by zIndex
+		iZIndex := allObjects[i].GetZIndex()
+		jZIndex := allObjects[j].GetZIndex()
+		if iZIndex != jZIndex {
+			return iZIndex < jZIndex
+		}
+
+		// then, if both are shapes, the containers come first
```
alixander, line 1033: `then, if both are shapes, parents come before their children`

d2renderers/d2svg/d2svg.go
```diff
@@ -1010,6 +1015,34 @@ func Render(diagram *d2target.Diagram) ([]byte, error) {
...
+		iShape, iIsShape := allObjects[i].(d2target.Shape)
+		jShape, jIsShape := allObjects[j].(d2target.Shape)
+		if iIsShape && jIsShape {
+			return iShape.Level < jShape.Level
+		}
+
+		// then, shapes come before connections
+		_, jIsConnection := allObjects[j].(d2target.Connection)
+		return iIsShape && jIsConnection
```
alixander, line 1042: u technically don't need this, cuz u add objects before connections already and it's sortstable. but i guess it's good to clarify

d2target/d2target.go
```diff
@@ -17,6 +17,11 @@ const (
 	MAX_ICON_SIZE     = 64
 )
 
+type DiagramObject interface {
```
alixander, line 20: unlike in Java, Go interfaces belong at site of usage, not definition. (so multiple usages can have their own interface)

alixander (changes requested):

d2renderers/d2svg/d2svg.go
ejulio-ts: you say this is always true, like that it can't happen for _i to be a connection and j be a shape_?

d2target/d2target.go
ejulio-ts: cool, TIL..
Now it somehow makes sense why the interface is not set with the struct..

d2renderers/d2svg/d2svg.go
alixander: no i mean, if u remove these 3 lines and return 0 or something (tie), then sortstable would already have the right order (cuz it's stable)

alixander (approved):

PR #273, opened 2022-11-30
self referencing connections

alixander:


d2graph/d2graph.go
```diff
@@ -719,10 +719,6 @@ func (obj *Object) Connect(srcID, dstID []string, srcArrow, dstArrow bool, label
 	src := srcObj.EnsureChild(srcID)
 	dst := dstObj.EnsureChild(dstID)
 
-	if src == dst {
-		return nil, errors.New("self-referencing connection")
```
nhooyr, line 723: wait what it just worked lol by getting rid of this error?

nhooyr: is it only TALA that doesn't support self referencing connections?

pleshevskiy: Seems like dagre too... without fail but... don't render connection

alixander: yeah it's wip, hopefully i'll find a way to configure dagre and elk to handle this. TALA doesn't support either but we can get that going in a day or so

alixander: alright fixed the bug in dagre and configured elk to have more spacing. it's okayy i guess. i'll turn this on for the sake of sequence diagrams for next release

pleshevskiy:
@alixander I had an arrow in ELK in the first version (WIP), but now it doesn't display.

# Before (WIP)

image

# After 

image

PR #291, opened 2022-12-01
Update cmdlog

nhooyr:


nhooyr:
This update adds support for logging to `*testing.T` and also correctly prefixes all standard library logs with a blue stdlog prefix. This will thus make playwright installation logs match as playwright uses the standard library logger to log.

alixander (approved):

PR #292, opened 2022-12-01
d2format: Expose in CLI under fmt subcommand

nhooyr:
- ci/dev.sh: Add
- lib/xmain: Print usage info in separate log
- d2format: Expose in CLI under fmt subcommand

### DIFF
--- a/autofmt.go
+++ b/autofmt.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"bytes"
+	"context"
+
+	"oss.terrastruct.com/xdefer"
+
+	"oss.terrastruct.com/d2/d2format"
+	"oss.terrastruct.com/d2/d2parser"
+	"oss.terrastruct.com/d2/lib/xmain"
+)
+
+func autofmt(ctx context.Context, ms *xmain.State) (err error) {
+	defer xdefer.Errorf(&err, "autofmt failed")
+
+	ms.Opts = xmain.NewOpts(ms.Env, ms.Log, ms.Opts.Flags.Args()[1:])
+	if len(ms.Opts.Args) == 0 {
+		return xmain.UsageErrorf("fmt must be passed the file to be formatted")
+	} else if len(ms.Opts.Args) > 1 {
+		return xmain.UsageErrorf("fmt only accepts one argument for the file to be formatted")
+	}
+
+	inputPath := ms.Opts.Args[0]
+	input, err := ms.ReadPath(inputPath)
+	if err != nil {
+		return err
+	}
+
+	m, err := d2parser.Parse(inputPath, bytes.NewReader(input), nil)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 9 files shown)


scsmithr:
I have a small request on this. Would it be possible to allow providing the input on stdin? I'd like to get formatting integrated into my emacs setup and it's easier to do that with tools that accept file contents on stdin.

scsmithr:
Oh nice, then this works for me.

ci/release/changelogs/next.md
```diff
@@ -1,5 +1,7 @@
 #### Features 🚀
 
+- autoformat is supported on the CLI with the `fmt` subcommand.
```
alixander, line 3: we should order changelogs by level of importance to user. i think Latex support is higher for example, maybe bump this down a few

autofmt.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func autofmt(ctx context.Context, ms *xmain.State) (err error) {
+	defer xdefer.Errorf(&err, "autofmt failed")
+
+	ms.Opts = xmain.NewOpts(ms.Env, ms.Log, ms.Opts.Flags.Args()[1:])
+	if len(ms.Opts.Args) == 0 {
+		return xmain.UsageErrorf("fmt must be passed the file to be formatted")
+	} else if len(ms.Opts.Args) > 1 {
+		return xmain.UsageErrorf("fmt only accepts one argument for the file to be formatted")
```
alixander, line 21: "accepts only one"

alixander: i think u can get rid of the other err msg too and just combine into this. e.g. `if len(ms.Opts.Args) != 1`

help.go
```diff
@@ -29,10 +31,11 @@ Flags:
 
 Subcommands:
   %[1]s layout - Lists available layout engine options with short help
-  %[1]s layout [layout name] - Display long help for a particular layout engine
+  %[1]s layout [name] - Display long help for a particular layout engine
+  %[1]s fmt file.d2 - Autoformat file.d2
```
alixander, line 35: should we just call it Format? Like, in the IDE plugins, it's auto, but in the CLI the user is just... formatting

alixander (changes requested):

ci/release/changelogs/next.md
nhooyr: I very very strongly disagree. autofmt is much more important... latex is a subset of users, autofmt is literally everyone.

autofmt.go
nhooyr: nah it's not clear if just one, will just update the grammar.

help.go
nhooyr: true, will change

ci/release/changelogs/next.md
nhooyr: We don't have any IDE plugins doing it automatically right now. So people will need to know about it in order to use it. And for plugin authors to implement it.

alixander (approved):

PR #308, opened 2022-12-01
Add markdown TOC generation using pandoc and shell

nhooyr:
See https://github.com/terrastruct/ci/blob/a875763eccbb51298807901be68c31e5b150dceb/lib/misc.sh#L28

The TOC is generated via pandoc and inserted via shell. `<!-- toc -->`
marks the insertion point, the TOC is inserted right below.

I'm using this in utils-go too. We should use it anywhere else we need
TOCs. I'm sure pandoc can handle the frontmatter in d2-docs too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ all: fmt gen lint build test
 
 .PHONY: fmt
 fmt:
-	prefix "$@" ./ci/sub/fmt/make.sh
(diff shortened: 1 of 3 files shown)


alixander (changes requested):
Screen Shot 2022-12-01 at 8 54 30 AM

alixander (approved):

PR #311, opened 2022-12-01
util-go migration

### LINKED ISSUE #175: cli: Avoid opening browser when using `--watch`
Is there a way to disable the browser opening when using `--watch`?
It is an annoyance when running in headless linux.

nhooyr:
closes #175


alixander (approved):
pulled and tested that install and tests work, tested browser=0 works.

PR #317, opened 2022-12-01
layout: sequence diagrams

### LINKED ISSUE #99: layout, syntax: Sequence diagrams

ejulio-ts:
fixes #99 

I'll followup with a PR to docs showcasing this.

Before any layout engine runs, sequence diagrams are processed.
Their content is replaced with a rectangle, of the same size of the laid out diagram, and then the layout engine just needs to place this object.
Later, the sequence diagram is moved around and added back to the graph.

Sequence diagram edges (messages) are also removed from the graph.

Renamed `edges`, in sequence diagrams, to `messages` as it is the definition in UML.

Added e2e test cases

Example of nested sequence diagrams with `TALA`
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (4)

```
this -> is -> orthogonal
is -> this
orthogonal -> this
with -> orthogonal
some -> orthogonal
nodes -> this
is -> some
this <-> and
and -> here
here -> you
here -> see
see -> a
see -> tree

some: {
    shape: sequence_diagram
    alice -> bob
    alice -> charlie
    charlie -> alice
    alice <- bob
}

here: {
    shape: sequence_diagram
    alice.a -> bob: a message to bob
    alice.a -> charlie.a
    charlie.a.b -> bob.a: another message to bob
    bob.a -> charlie.a.b
    charlie.a.b -> alice.a: {
        stroke-dash: 3
        stroke: green
    }
}
```

another example in tala
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (5)

```
this -> is -> a
this -> subgraph
is -> subgraph
a -> this

another: {
    shape: sequence_diagram
    alice.a.b. -> bob.c
    bob.c.d -> alice.a
    bob.c -> alice
    alice -> bob
}

what -> about -> hierarchies
they -> about
they -> what
should -> hierarchies
work -> should

what: {
    shape: sequence_diagram
    alice -> bob: talks with
}
```


d2layouts/d2sequence/layout.go
```diff
@@ -2,280 +2,119 @@ package d2sequence
...
-	maxActorHeight float64
-}
-
-func (sd *sequenceDiagram) init() {
-	sd.edges = make([]*d2graph.Edge, len(sd.graph.Edges))
-	copy(sd.edges, sd.graph.Edges)
-
-	queue := make([]*d2graph.Object, len(sd.graph.Root.ChildrenArray))
-	copy(queue, sd.graph.Root.ChildrenArray)
+// Layout identifies and layouts sequence diagrams within a graph
```
alixander, line 12: performs layout on

alixander (changes requested):
let's see some TALA runs

alixander (changes requested):
i don't think we should wrap to compose layout engines. i think just simple in-order run-through is best. what if we implement swimlanes then gantt, etc? starts to get into 4 nested funcs.

or like, maybe we can let users specify a different layout for each container.

how about we change the signature of a layout to be `func(ctx, g) (cleanup, err)`, and then for the core layouts, the cleanup is just a no-op.

then u can have a function that runs multiple

```
func RunLayouts() (err error) {
	layouts := [d2sequence.Layout, layout]
	for _, layout := range layouts {
		cleanup, err := layout(ctx, g)
		if err != nil { return err }
		defer func() { 
			if err != nil {
				err = cleanup()
			}
		}
	}
        return nil
}
```

ejulio-ts:
I don't see any issue with nesting and composition, that's the whole deal with making things extensible. If these calls were recursive, then I'd say it could become an issue
I just went with a simple `decorator` approach here. If you think that it needs to change the API, then ok..

Just note that your particular example leaves cleanup to the end of everything, while you might want to restore things on the go..

alixander:
wdym "restore things on the go"? at no point would one layout engine want to do anything before another layout has finished right?

alixander:
> that's the whole deal with making things extensible

i also don't get why it's "the whole deal with making things extensible". there's many different ways

>  leaves cleanup to the end of everything,

not of everything, just... until another layout engine finished

ejulio-ts:
> wdym "restore things on the go"? at no point would one layout engine want to do anything before another layout has finished right?

In your example, you defer `cleanup` and, as I understand, it only runs when `RunLayouts` ends. In this case, the layouts would only cleanup and restore the state after everything ran.
I get your idea, I see no problem with as I see no problem with this one.
The only thing with yours is that for `cleanup` you'll need either an object to store the state or create a closure to keep the state around.
While in function composition, just pass a single pointer around and the "closure" is kept by the call stack. I also get that the current function interface is not proper for this as `d2sequence` exposes a different interface than the plugins.

alixander:
> after everything ran.

isn't that the same as right now? your cleanup code only runs after dagre is run. 

> exposes a different interface than the plugins.

yes we would need a different interface. i'm okay doing this in another PR and we leave structure as is for this one

d2layouts/d2sequence/layout.go
```diff
@@ -2,280 +2,139 @@ package d2sequence
...
+		objectsToKeep[obj] = struct{}{}
+		if obj.Attributes.Shape.Value == d2target.ShapeSequenceDiagram {
+			// TODO: should update obj.References too?
+
+			// clean current children and keep a backup to restore them later
+			obj.Children = make(map[string]*d2graph.Object)
+			children := obj.ChildrenArray
+			obj.ChildrenArray = nil
+			// creates a mock rectangle so that layout considers this size
+			sdMock := obj.EnsureChild([]string{"sequence_diagram"})
```
alixander, line 45: oh interesting, a new object as a child of the sequence diagram...

d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+	}
+}
+
+func (sd *sequenceDiagram) getMessageY(rank int) float64 {
+	// +1 so that the first message has the top padding for its label
+	return ((float64(rank) + 1.) * sd.messageYStep) + sd.maxActorHeight
+}
+
+func (sd *sequenceDiagram) isActor(obj *d2graph.Object) bool {
```
alixander, line 275: save the root object on the sequenceDiagram. Then, if the obj.parent == the root object, then u know it's an actor

alixander (changes requested):
the sequence_diagram code is nice and clean, but i think the layout code needs some work. i think a bunch of these data structures are unnecessary or can be simplified

alixander (approved):

PR #331, opened 2022-12-03
sequence diagrams: beautify

ejulio-ts:
Follows @katwangy guidelines

The changes below are highlighted in the image
- [x] 1. Remove edges horizontal pad
- [x] 2. edge y pad is 16
- [x] 3. Set actors to constant size 150x100 (wxh)
  - Here we set only the min width to 150 and resize height keeping the spec ratio 
- [x] 4. actors horizontal distance is 250
- [x] 5. lifeline width is 2px dash 8
- [x] 6. spans have base 12x80, 80 is default size if there is only one arrow
- [x] 7. spans grow by 8, 4 each side, when nesting
- [x] 8. Edge step is 80
- [x] 9. person and image shape have bottom label

image

Title padding
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (6)

Actors font size
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (7)


alixander:
it seems like there's a bug that the font size is mismatching. not sure if related

## Before

sequence_before

## After

sequence_after

alixander (approved):

PR #334, opened 2022-12-03
prevent illegal sequence diagram connection

### LINKED ISSUE #324: sequence_diagram: prevent edges from going into sequence diagram inner object from outside

alixander:
closes #324

### DIFF
--- a/d2graph/d2graph.go
+++ b/d2graph/d2graph.go
@@ -711,6 +711,16 @@ func (e *Edge) AbsID() string {
 	return fmt.Sprintf("%s(%s %s %s)[%d]", commonKey, strings.Join(srcIDA, "."), e.ArrowString(), strings.Join(dstIDA, "."), e.Index)
 }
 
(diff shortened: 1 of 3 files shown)


d2graph/d2graph.go
```diff
@@ -732,6 +742,10 @@ func (obj *Object) Connect(srcID, dstID []string, srcArrow, dstArrow bool, label
 	src := srcObj.EnsureChild(srcID)
 	dst := dstObj.EnsureChild(dstID)
 
+	if src.outerSequenceDiagram() != dst.outerSequenceDiagram() {
+		return nil, errors.New("connections within sequence diagrams can connect only to other objects within the same sequence diagram")
```
ejulio-ts, line 746: it could also be that objects inside a sequence diagram can't receive connections from the outside

alixander: this error message would cover that. there is no such thing as "receive connection", as that implies directionality. an edge connects two objects.

PR #336, opened 2022-12-03
sequence diagram: self edges

ejulio-ts:
based on https://github.com/terrastruct/d2/pull/273
routes self edges and edges to descendants
added a new test case (shown below)
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (8)


alixander:
Screen Shot 2022-12-03 at 10 08 43 AM

@katwangy what do u think? should we have labels in middle or up top like sequencediagram.org does? i think on top to get more space.

@ejulio-ts ur PRs need to always either include test case (we have a todo_test.go specifically for these) or add a note or address it somehow. "a deeper descendant" looks like it's about to overflow, so if u make it longer, does it? or do u handle it?

ejulio-ts:
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (9)

something like this, @alixander ?

alixander:
yes. i think u should place these labels like sequencediagram.org does -- on top

d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -76,17 +76,27 @@ func newSequenceDiagram(actors []*d2graph.Object, messages []*d2graph.Edge) *seq
...
-		sd.setMinMaxMessageRank(message.Src, rank)
-		sd.setMinMaxMessageRank(message.Dst, rank)
-
 		// ensures that long labels, spanning over multiple actors, don't make for large gaps between actors
 		// by distributing the label length across the actors rank difference
 		rankDiff := math.Abs(float64(sd.objectRank[message.Src]) - float64(sd.objectRank[message.Dst]))
-		distributedLabelWidth := float64(message.LabelDimensions.Width) / rankDiff
-		sd.actorXStep = math.Max(sd.actorXStep, distributedLabelWidth+HORIZONTAL_PAD)
+		if rankDiff != 0 {
+			// rankDiff = 0 for self edges
```
alixander, line 86: self edges need extra padding too. for long self-edge labels.

d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -142,7 +146,8 @@ func (sd *sequenceDiagram) placeActors() {
 //        │
 //        │
 func (sd *sequenceDiagram) addLifelineEdges() {
-	endY := sd.getMessageY(len(sd.messages))
+	lastRoute := sd.messages[len(sd.messages)-1].Route
+	endY := lastRoute[len(lastRoute)-1].Y + MIN_MESSAGE_DISTANCE
```
alixander, line 150: u need to for loop the route, prepare for non-instantaneous msgs

alixander (changes requested):

alixander:
we can get the edge label in another PR, good to just have the ability to do self edges for now

PR #344, opened 2022-12-04
sequence diagram: notes

alixander:
syntax for specifying notes: a child of an actor with no edges connected to it
(a child of an actor with edges connected to it is a span)

my plan is that notes can be specified "right of" and "left of" by using `direction` keyword.

Screen Shot 2022-12-04 at 12 12 51 AM


d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -20,33 +20,71 @@ type sequenceDiagram struct {
...
 	spans     []*d2graph.Object
+	notes     []*d2graph.Object
 
 	// can be either actors or spans
 	// rank: left to right position of actors/spans (spans have the same rank as their parents)
 	objectRank map[*d2graph.Object]int
 
 	// keep track of the first and last message of a given actor/span
 	// the message rank is the order in which it appears from top to bottom
+	// TODO rank is used for y coordinate here and x coordinate above??
```
ejulio-ts, line 31: this is gone after the self edges PR.
Anyway, rank is just another word for `order`, so actors rank is left to right, while edges rank is top to bottom

d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -63,21 +101,34 @@ func newSequenceDiagram(actors []*d2graph.Object, messages []*d2graph.Edge) *seq
...
+			// notes are children of actors with no edges
+			if hasEdge(child) {
+				// spans have no labels
+				// TODO why not? Spans should be able to
+				child.Attributes.Label = d2graph.Scalar{Value: ""}
+				child.Attributes.Shape = d2graph.Scalar{Value: shape.SQUARE_TYPE}
+				sd.spans = append(sd.spans, child)
+				sd.objectRank[child] = rank
+			} else {
+				sd.verticalIndices[child.AbsID()] = getObjEarliestLineNum(child)
```
ejulio-ts, line 117: what about deeper descendants.
It seems that they would be drawn as notes too

d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -143,6 +195,11 @@ func (sd *sequenceDiagram) placeActors() {
 //        │
 func (sd *sequenceDiagram) addLifelineEdges() {
 	endY := sd.getMessageY(len(sd.messages))
+
+	for _, note := range sd.notes {
```
ejulio-ts, line 199: notes were already placed, and messages too.
```
endY := math.Max(lastMessage.Route[1].Y, lastNote.TopLeft.Y+lastNote.Height) + sd.yStep
```

d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -169,6 +226,34 @@ func (sd *sequenceDiagram) addLifelineEdges() {
...
+		verticalIndex := sd.verticalIndices[note.AbsID()]
+		y := sd.maxActorHeight + sd.yStep
+
+		for _, msg := range sd.messages {
+			if sd.verticalIndices[msg.AbsID()] < verticalIndex {
+				y += sd.yStep
+			}
+		}
+		for _, otherNote := range sd.notes {
```
ejulio-ts, line 244: here you can start at the the current `note` index as the previous ones were already processed.
In the previous loop, you could keep a `lastEdgeIndex` outside the loop to avoid going through all of them again (though, it may hurt readability a bit, so up to you)

d2layouts/d2sequence/sequence_diagram.go
alixander: i thought about this.

when we have "edge groups", notes should be able to be descendants of them.

for spans, there will be an edge reference.

d2layouts/d2sequence/sequence_diagram.go
alixander: can't use route[1] cuz later on there may be non-instantaneous msgs, but other than that, it's a good point

PR #349, opened 2022-12-04
Fix tests

alixander:
CI was broken, unaware of failing tests


ejulio-ts:
I thought about removing `layout_test` as it seems a bit convoluted and we could use `e2e` assertions to get the same benefit of sanity checking some conditions

PR #350, opened 2022-12-04
sequence diagram groups

alixander:
a group is an object in sequence diagrams that
- has no messages connecting to it
- contains a message or note

the style isn't correct, but i'm going to get all the styles in one go at the end

Screen Shot 2022-12-04 at 2 42 26 PM


d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -120,8 +163,9 @@ func newSequenceDiagram(actors []*d2graph.Object, messages []*d2graph.Edge) *seq
 			queue = queue[1:]
 
 			// spans are children of actors that have edges
-			// notes are children of actors with no edges and contains no messages
-			if hasEdge(child) && !sd.containsMessage(child) {
+			// notes are children of actors with no edges and no children
+			// edge groups are children of actors with no edges and children edges
```
ejulio-ts, line 167: they aren't children of actors, right?

alixander: oh right, yeah comment is wrong

PR #356, opened 2022-12-05
Fix: Update import path of textmeasure in examples

martinmbaya:


martinmbaya:
Installing the `textmeasure` package imported in the examples located in `docs/examples/lib` was throwing the error below:
`go: module oss.terrastruct.com/d2@upgrade found (v0.0.13), but does not contain package oss.terrastruct.com/d2/lib/textmeasure`

nhooyr (commented):
Sorry we're just about to do a new release this morning. If you run `go get oss.terrastruct.com/d2@master` then the existing import paths should work.

alixander:
should be fixed now with the new release (/lib is the right place). ty for taking the initiative anyway to correct it @martinmbaya !

martinmbaya:
You're welcome!
It's working fine now

PR #361, opened 2022-12-05
Mask group label

ejulio-ts:
Adds a mask to shape labels so that edges going through it don't make it hard to read the label

Didn't add a new test case because we already have ones considering it
- `sequence_diagram_groups`
- `child_to_child`
- many others

~~This is a global change to all diagrams, not scoped to sequence diagrams.~~
~~Some of them end up cropping a part of the edge, but it is related to the container label misplacement and not necessarily this change, so I assume that the correct solution for that is to fix the label position and not scope it to only sequence diagrams.~~
Shape labels are masked if `MaskLabel` is set to `true`


alixander (changes requested):
can u try with 0 padding for these masks (just for labels of shapes, keep labels of edges with padding)? i think that would avoid these

Screen Shot 2022-12-05 at 12 07 45 PM

e2etests/testdata/sanity/child_to_child/elk/sketch.exp.svg
```diff
@@ -14,7 +14,13 @@ width="464" height="862" viewBox="-88 -88 464 862"><style type="text/css">
...
+<rect x="0" y="0" width="464" height="862" fill="white"></rect>
+<rect x="136.000000" y="17.000000" width="17" height="36" fill="black"></rect>
```
gavin-ts, line 19: what if we tried something like semi-transparent edge lines when crossing a node label, but full opacity when it is  the edge label?
```suggestion
<rect x="136.000000" y="17.000000" width="17" height="36" fill="rgba(0,0,0,0.5)"></rect>
```

alixander: oh that's an interesting idea. yeah i'd like to see that version @ejulio-ts

ejulio-ts: something like this @gavin-ts @alixander 
image

alixander: can u do this for all the test cases? just like implement it for a commit

ejulio-ts: yeah, just wanted to confirm before working on it 😅

d2exporter/export.go
```diff
@@ -131,6 +131,9 @@ func toShape(obj *d2graph.Object, theme *d2themes.Theme) d2target.Shape {
 	if obj.LabelPosition != nil {
 		shape.LabelPosition = *obj.LabelPosition
 	}
+	if obj.MaskLabel != nil {
```
alixander, line 134: this shouldn't be in exporter. this is purely a render thing (e.g. if one day someone uses d2 diagrams rendered to Canvas or d3, this shouldn't appear).

i'm not sure how else to mark it though. Maybe you can create a `sequenceAttrs` (make sure to omitempty).

even currently, why are there a bunch of "false" in the testdata if u have omitempty?

alixander (changes requested):
reuse `blend`on d2target

d2renderers/d2svg/d2svg.go
```diff
@@ -991,14 +995,14 @@ func Render(diagram *d2target.Diagram) ([]byte, error) {
 	markers := map[string]struct{}{}
 	for _, obj := range allObjects {
 		if c, is := obj.(d2target.Connection); is {
-			labelMask := drawConnection(buf, c, markers, idToShape)
-			if labelMask != "" {
+			if labelMask := drawConnection(buf, c, markers, idToShape); labelMask != "" {
```
alixander, line 998: i don't actually prefer these even if u can do it. it makes the line too long and it's harder to read. can u change these back?

alixander (changes requested):

alixander:
looks good other than the nit!

alixander (approved):

PR #363, opened 2022-12-05
Fix actor padding for long notes

ejulio-ts:
note width was only considered on the actor right side, fixed to consider both.

image


alixander (approved):
nice

PR #364, opened 2022-12-05
fix sequence layout placements

alixander:
Screen Shot 2022-12-05 at 1 56 02 PM


d2exporter/export.go
```diff
@@ -99,6 +100,10 @@ func toShape(obj *d2graph.Object, theme *d2themes.Theme) d2target.Shape {
 	shape.Italic = text.IsItalic
 	shape.FontSize = text.FontSize
 
+	if strings.EqualFold(obj.Attributes.Shape.Value, d2target.ShapeSequenceDiagram) {
```
ejulio-ts, line 103: obj.IsSequenceDiagram

d2graph/d2graph.go
```diff
@@ -339,8 +327,26 @@ func (obj *Object) GetFill(theme *d2themes.Theme) string {
...
+			return theme.Colors.B6
+		} else if level == 3 {
+			return theme.Colors.Neutrals.N6
+		}
+		return theme.Colors.Neutrals.N7
+	}
+
 	shape := obj.Attributes.Shape.Value
 
+	if strings.EqualFold(shape, d2target.ShapeSequenceDiagram) {
```
ejulio-ts, line 346: same as above

d2layouts/d2sequence/layout.go
```diff
@@ -141,14 +141,26 @@ func cleanup(g *d2graph.Graph, sequenceDiagrams map[string]*sequenceDiagram, obj
...
+		)
 
 		obj.Children = make(map[string]*d2graph.Object)
+		obj.ChildrenArray = make([]*d2graph.Object, 0)
 		for _, child := range sd.actors {
 			obj.Children[child.ID] = child
+			obj.ChildrenArray = append(obj.ChildrenArray, child)
+		}
+		for _, child := range sd.groups {
+			if child.Parent == obj {
```
ejulio-ts, line 159: something tells me this won't work with TALA as the object references change in that call

alixander: changed to IDs. that'll work right

PR #369, opened 2022-12-06
Fix shape:note to resize the top right corner (fold) as the shape size grows

ejulio-ts:
Changes `shape: page` SVG path to keep the fold on the top right corner constant, regardless of the shape size.

Note that I had to change the base from 30.5x36.5 to 66x79 in `pageInnerPath` so they would match the values from `pageOuterPath` (see individual PRs for these).

Also changed `note` type in sequence diagrams

Before
image

After
image


alixander (approved):
nice, looks good

PR #372, opened 2022-12-06
install.sh: Improve default $PREFIX behaviour

nhooyr:
As before we default to /usr/local but now we switch to ~/.local if /usr/local if not accessible instead of prompting for sudo. Instead of having an exception for macOS/arm64.

Of course if a user wants to install in /usr/local anyway, they can still pass --prefix /usr/local to force installation into /usr/local and prompt for sudo.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -47,9 +47,9 @@ We're careful shell programmers and are aware of the many footguns of the Unix s
 script was written carefully and with detail. For example, it is not vulnerable to partial
 execution and the entire script runs with `set -eu` and very meticulous quoting.
 
-It follows the XDG standards, installs `d2` properly into a Unix hierarchy path (defaulting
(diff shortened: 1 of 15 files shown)


docs/INSTALL.md
```diff
@@ -126,8 +126,7 @@ The install script places the standalone release into `$PREFIX/lib/d2/d2-<versio
 and we recommend doing the same with manually installed releases so that you
 know where the release directory is for easy uninstall.
 
-> warn: Our binary releases aren't fully static like normal Go binaries due to the C
-> dependency on v8go for executing dagre. If you're on an older system with an old
+> warn: Our binary releases aren't fully static. If you're on an older system with an old
```
alixander, line 129: is this still needed?

nhooyr: Nope, will remove good catch.

alixander (approved):

PR #373, opened 2022-12-06
unique label mask

alixander:
fixes https://github.com/terrastruct/d2/pull/332

i was editing the docs, and saw weird masking. turns out it was grabbing the ID from other d2 svg's in the same DOM.

Screen Shot 2022-12-05 at 10 34 33 PM


pleshevskiy:
image

Seems like this commit related with #442

PR #377, opened 2022-12-06
install.sh: Add warning if d2 is shadowed on installation

### LINKED ISSUE #107: install.sh: Explain in summary that installed D2 might be shadowed by existing install that we could not uninstall
It's confusing for us as devs.

nhooyr:
Closes #107


nhooyr:
Screenshot 2022-12-06 at 1 30 04 AM

PR #394, opened 2022-12-06
d2chaos: fixes chaos tests

### LINKED ISSUE #376: d2chaos: respect sequence diagrams
https://github.com/terrastruct/d2/actions/runs/3627398645/jobs/6117297838

```
test: --- FAIL: TestD2Chaos (0.00s)
test:     d2chaos_test.go:49: writing generated files to out
test:     --- FAIL: TestD2Chaos/#00 (0.02s)
test:         d2chaos_test.go:82: failed to create "\"G\\n\x1c=f\x02>f)6\x17\x0ffk'np4\x18+W^-\x02>MC\x10\x1ew\" -> \"G\\n\x1c=f\x02>f)6\x17\x0ffk'np4\x18+W^-\x02>MC\x10\x1ew\".\"tk[TOmI(X\x05bv6K(6\x06 d@\"": failed to recompile:
test:             "G\n=f>f)6fk'np4+W^->MCw": {
test:               shape: sequence_diagram
test:               "tk[TOmI(Xbv6K(6 d@": {shape: page}
test:             }
test:             "G\n=f>f)6fk'np4+W^->MCw" -> "G\n=f>f)6fk'np4+W^->MCw"."tk[TOmI(Xbv6K(6 d@"
test:             
test:             5:1: connections within sequence diagrams can connect only to other objects within the same sequence diagram
```

cc @gavin-ts @ejulio-ts

ejulio-ts:
Fixed #376

Fixes the two issues in #376 

Each commit should be a separate fix.

- only add an edge between sequence diagrams if it is valid, otherwise skip it
- removed the check for self edges in chaos tests as we support it now
- sequence diagrams weren't expected in chaos tests, as images, but the `if`s there allowed it
  - as I fixed the sequence diagrams issues, they are now allowed in chaos tests
  - images are still not allowed because they need an `icon`. I thought about adding some code to generate random URLs, but as we lived without it so far, I don't think this is critical, so skipped
- fixed some other issues found with special chars in IDs with `dagre`
  - changed the test from `dagre_id_with_new_line` to `dagre_id_with_special_chars` and added the new examples to that test


alixander (approved):

d2layouts/d2dagrelayout/layout.go
```diff
@@ -257,7 +258,11 @@ func setGraphAttrs(attrs dagreGraphAttrs) string {
 }
 
 func escapeID(id string) string {
-	id = strings.ReplaceAll(id, `\n`, `\\n`)
+	// fixes \\
+	id = strings.ReplaceAll(id, "\\", `\\`)
+	// replaces \n with \\n whenever \n is not preceded by \ (does not replace \\n)
+	re := regexp.MustCompile(`[^\\](\n)`)
```
nhooyr, line 264: Why the paren around `\n`?

nhooyr (commented):
good stuff ✅

d2layouts/d2dagrelayout/layout.go
ejulio-ts: My bad, for a moment I thought I had to replace the group and ended forgetting it there 😅

PR #395, opened 2022-12-06
fix builds on 32bit systems

Vaelatern:
Lots of changes that sum up to "int() on 32bit systems is 32 bits, not 64, so you need to call out 64 bits to use it there".

I'll be running this personally on a 64 bit machine.

It seems that gofmt ran while I was saving these, so some of these changes may break asciiart.

With the go-ahead I'll fix the tests too.


alixander:
> It seems that gofmt ran while I was saving these, so some of these changes may break asciiart.

yeah idk what's going on with gofmt latest version. we are on go 1.18 to avoid that ascii-breaking behavior. can you please use 1.18's gofmt? hopefully the Go team will address this but I haven't found any issues yet. 

> you need to call out 64 bits to use it there

we don't afaik actually need the 64 bits, and I'd prefer not to have to convert everywhere/anywhere in the codebase.

Somebody a while ago made a PR but didn't want to sign the CLA, which I think is trying to solve the same issue here in a different manner. https://github.com/terrastruct/d2/pull/137

Vaelatern:
I signed the CLA, if that helps.

If we don't need the 64bit everywhere, then we can look at alternatives that pass the build....

alixander:
> I signed the CLA, if that helps.

yup i saw, thank you -- i was referring the previous PR, whom we couldn't accept their contribution due to lack of CLA.

> If we don't need the 64bit everywhere, then we can look at alternatives that pass the build....

I'm pretty sure we don't. This is the highest number we go to https://github.com/terrastruct/d2/blob/3f62eed2420b1765e45bc9d29045431e2163b515/lib/imgbundler/imgbundler.go#L25

Theme IDs are currently int64, but that's for sure not necessary.

PR #397, opened 2022-12-06
sequence_diagram: Fix span size

### LINKED ISSUE #391: sequence_diagram: span doesn't cover
```
How sequence diagrams are rendered in D2: {
  shape: sequence_diagram

  CLI; d2ast; d2compiler; d2layout;
  layout engine; d2themes; d2exporter; d2renderer

  errors parsed, if any: {
    CLI -> d2ast: "'How sequence diagrams are ...'"
    d2ast -> CLI: tokenized AST
  }
  CLI -> d2compiler: compile AST
  d2compiler."validate syntax"
  d2compiler."measurements of text take place,\nunless explicitly provided"
  d2compiler -> CLI: objects and edges
  CLI -> d2layout.layout: run layout engines
  majority of runtime: {
    d2layout.layout -> d2layout.layout: sequence_diagram layout
    d2layout.layout -> layout engine: run core engine on rest
  }
  d2layout.layout -> d2layout.layout: add back in sequence diagrams
  d2layout -> CLI: diagram with correct positions and dimensions
  CLI -> d2exporter: export diagram with chosen theme and renderer
  d2exporter.export -> d2themes: get theme styles
  d2exporter.export -> d2renderer: render to SVG
  d2exporter.export -> CLI: resulting SVG
}
```
Screen Shot 2022-12-06 at 11 25 00 AM

ejulio-ts:
Fixes #391 

Span size wasn't computed correctly for self edges. It was considering the first route point instead of the last one for the maximum edge Y.

Added a new regression case. Below is a more convolute one

Diagram of the issue
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (11)

Nested self edges test
_Users_juliobatista_Projects_terrastruct_d2_sequence svg (12)


d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -402,19 +402,18 @@ func (sd *sequenceDiagram) placeSpans() {
 		// finds the position if there are messages to this span
 		minMessageY := math.Inf(1)
 		if firstMessage, exists := sd.firstMessage[span]; exists {
-			// needs to check Src/Dst because of self-edges or edges to/from descendants
-			if span == firstMessage.Src {
+			if firstMessage.Src == firstMessage.Dst || span == firstMessage.Src {
```
alixander, line 405: why don't u just do like get min and max of all points instead of assuming src/dst? seems less robust

alixander (changes requested):

d2layouts/d2sequence/sequence_diagram.go
ejulio-ts: Edges to/from descendants need this, as well self edges.
In these cases, taking the min/max of the first/last points will make spans bigger than they they should be

This is what happens to `sequence_diagram_self_edges` if
```
minMessageY := math.Inf(1)
		if firstMessage, exists := sd.firstMessage[span]; exists {
			minMessageY = math.Min(firstMessage.Route[0].Y, firstMessage.Route[len(firstMessage.Route)-1].Y)
		}
		maxMessageY := math.Inf(-1)
		if lastMessage, exists := sd.lastMessage[span]; exists {
			maxMessageY = math.Max(lastMessage.Route[len(lastMessage.Route)-1].Y, lastMessage.Route[0].Y)
		}
```

Result
image

Expected
image

PR #398, opened 2022-12-07
fix builds on 32bit systems

Vaelatern:
This is to supersede #395 with an alternative that fixes the initial problem by ignoring it -- let's NOT do 64 bit integers!

That should be fine, right?

I also didn't include the go fmt changes for ascii art.

### DIFF
--- a/d2target/d2target.go
+++ b/d2target/d2target.go
@@ -47,10 +47,10 @@ func (diagram Diagram) BoundingBox() (topLeft, bottomRight Point) {
 	if len(diagram.Shapes) == 0 {
 		return Point{0, 0}, Point{0, 0}
 	}
-	x1 := int(math.MaxInt64)
-	y1 := int(math.MaxInt64)
-	x2 := int(-math.MaxInt64)
-	y2 := int(-math.MaxInt64)
+	x1 := int(math.MaxInt32)
+	y1 := int(math.MaxInt32)
+	x2 := int(math.MinInt32)
+	y2 := int(math.MinInt32)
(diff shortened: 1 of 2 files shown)


d2target/d2target.go
```diff
@@ -44,10 +44,10 @@ func (diagram Diagram) HashID() (string, error) {
...
 
 func (diagram Diagram) BoundingBox() (topLeft, bottomRight Point) {
-	x1 := int(math.MaxInt64)
-	y1 := int(math.MaxInt64)
-	x2 := int(-math.MaxInt64)
-	y2 := int(-math.MaxInt64)
+	x1 := int(math.MaxInt32)
+	y1 := int(math.MaxInt32)
+	x2 := int(-math.MaxInt32)
+	y2 := int(-math.MaxInt32)
```
alixander, line 50: looks like `MinInt32` is a thing. can u switch to that?

alixander (changes requested):

d2target/d2target.go
Vaelatern: Made the change

alixander:
can u rebase master in? that failing test is another issue i've just fixed: https://github.com/terrastruct/d2/pull/399

alixander (approved):

Vaelatern:
I'll learn about signing commits in the next two days.

Vaelatern:
@alixander Ready for the approval.

alixander:
thanks @Vaelatern !

PR #441, opened 2022-12-13
changed README to say that d2 supports png now.

andorsk:
d2 README showed only SVG exports accepted. Changed to add png as suggested by @alixander that this is currently supported now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ one, please see [./d2renderers/d2fonts](./d2renderers/d2fonts).
 
 ## Export file types
 
-D2 currently supports SVG exports. More coming soon.
+D2 currently supports SVG and PNG file exports. More coming soon. 
 
 ## Language tooling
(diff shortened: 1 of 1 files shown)


alixander:
included in another PR, ty! https://github.com/terrastruct/d2/pull/445/files

andorsk:
@alixander generally to encourage open source contributions, I would suggest giving the people that came up with the idea/PR credit to their contribution.

alixander:
@andorsk You're right. The reason for this is that we have a CLA for contributing to D2 (https://github.com/terrastruct/d2/blob/master/docs/CONTRIBUTING.md#logistics), but I didn't want to ask you to sign anything for a small doc edit. Thank you for the contribution and I apologize for the misunderstanding. We are in the process of evaluating whether we need a CLA, so this may not be a problem in the future.

PR #445, opened 2022-12-14
increase png export quality

berniexie:
We can very easily increase the png export quality inside the script. Here is a simple comparison where the before and after shows the difference. Multiplying the width and height by 1.5 is the most straightforward, but we can of course choose other values, or only when the height or width is under a certain size.

BEFORE:
before

AFTER
after


lib/png/generate_png.js
```diff
@@ -11,8 +11,8 @@ async (imgString) => {
   };
   const img = await loadImage();
   const canvas = document.createElement("canvas");
-  canvas.width = img.width;
-  canvas.height = img.height;
+  canvas.width = img.width * 1.5;
```
nhooyr, line 14: Let's go with `2x` to match retina displays. Maybe this should be a flag actually? `--png-scale`?

nhooyr (changes requested):

lib/png/generate_png.js
alixander: Javascript should actually have access to this info: https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio , but experimentally doesn't work for me. 

yeah i'm down for 2x being default and then flag to control

nhooyr: This code doesn't run in a real browser and so the devicePixelRatio will always be one as there is no real display.

alixander: sry should've specified, the experiment part was in d2 playground, where i copied some of this code and tried with devicepixel ratio: https://github.com/terrastruct/d2-playground/blob/master/src/js/modules/export.js#L46 . that does run in a real browser

nhooyr: was the browser window displayed on your external monitor or internal display? each would have different devicePixelRatios. external display probably just 1 whereas internal should be 2.

alixander: yea just macbook

nhooyr: What do you have your display scaling set to in system preferences? What does typing devicePixelRatio in the JS console print?

alixander: 2. i'll report back what it prints during the export run when i'm next working on that

alixander: it's also 2, and it works. i must've put it in the wrong spot before, thx for verifying

nhooyr: This isn't resolved, this should still be configurable. devicePixelRatio inside playwright will always be 1. Either `--png-scale` or `--png-width` (we always keep aspect ratio same)

alixander: https://github.com/terrastruct/d2/issues/463

PR #508, opened 2022-12-23
fix null key

### LINKED ISSUE #507: n is lack in svg
## Step to reproduce

1. Write below code as d2 file

```markdown
hello -> null
```

2. Run `d2 null.d2 null.svg`
3. Show null.svg

null

## Expected behavior

null is displayed.

## Actual behavior

ull is displayed.

## System configuration

**d2**: v0.1.2

alixander:
closes #507 

was it in early designs that `null` was special?

without this, IDVal was treating it as newline
Screen Shot 2022-12-22 at 7 08 17 PM


d2format/escape.go
```diff
@@ -47,10 +47,6 @@ func escapeUnquotedValue(s string, inKey bool) string {
 		return `""`
 	}
 
-	if strings.EqualFold(s, "null") {
-		return "\\null"
```
nhooyr, line 51: The right fix should be changing this to `"\\\\null"`

nhooyr: The idea is that this null string should be parsed not as the value null but as the string null.

You could also just make it \`'null'\` to make it more clear.

nhooyr (changes requested):

nhooyr:
@alixander null is still special, it's not parsed the string null. We just don't use this ability anywhere yet.

alixander:
hm ok, i guess when layers/scenarios come, it can be used for "unsetting" an inherited attribute

PR #530, opened 2022-12-26
docs: Add reference to community MdBook preprocessor

danieleades:


alixander (approved):
thanks @danieleades !

Would you kindly include it in the d2 docs as well? https://github.com/terrastruct/d2-docs/blob/master/docs/tour/extensions.md#community

PR #545, opened 2022-12-28
Tooltips

### LINKED ISSUE #367: Tooltips
### Discussed in https://github.com/terrastruct/d2/discussions/329

<div type='discussions-op-text'>

<sup>Originally posted by **azban** December  2, 2022</sup>
It would be nice to have tooltips for objects, so that by default the diagrams are concise, but more information can be encoded for someone looking to dig into specific elements.</div>

alixander:
On SVG, tooltips will be hover to show.

Screen Shot 2022-12-28 at 12 29 54 PM

Exported to PNG, the tooltips will turn into a numbered legend.

Screen Shot 2022-12-28 at 12 25 09 PM

closes #367


d2renderers/d2svg/d2svg.go
```diff
@@ -39,10 +39,15 @@ const (
 	DEFAULT_PADDING            = 100
 	MIN_ARROWHEAD_STROKE_WIDTH = 2
 	threeDeeOffset             = 15
+
+	tooltipIconLen = 32
```
ejulio-ts, line 43: `tooltipIconSize`?

main.go
```diff
@@ -247,7 +248,8 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, sketc
 
 	out := svg
 	if filepath.Ext(outputPath) == ".png" {
-		svg := svg
+		svg := appendix.AppendTooltips(diagram, ruler, svg)
```
ejulio-ts, line 251: Why not call `d2svg` with an extra argument like `withTooltipAppendix` instead of manipulating raw HTML/SVG?

alixander: more composable. d2svg.Render would get real messy with all the flags.

d2renderers/d2svg/d2svg.go
alixander: it's circular, and uses radius elsewhere, so length clarifies that

ejulio-ts: The radius usage is not related to this one.
maybe, diameter then

main.go
ejulio-ts: I won't enter a long discussion on the composition as it seems quite hackish to split the html and append back, so I'll assume this is ok for now and we can refactor later on if we need something else

ejulio-ts (approved):

main.go
alixander: yeah but that's two separate things. like it should not be an XML string until the very end. d2svg.Render should return a structured XML that postprocessing steps can manipulate.

ejulio-ts: it makes sense, though at the same time the interface could be actual composition/chain of functions where you don't need to parse, use regexes and change viewports after it was ready.
In this case, `main` would decide which functions to include in the pipeline
```
sag := Render([diagram, tooltips])

func diagram(buffer) {
    buffer.write(...)
}

func tooltips(buffer) {
    buffer.write(...)
}
```

alixander: > parse, use regexes and change viewports

it would still need to do all that in the proposed pseudocode

ejulio-ts: Why? if the buffer is always appending to the end.
`Render` opens `<svg>` and closes `</svg>` everything else happens in between

alixander: oh i see. yeah that'd be a good refactor to do.

d2renderers/d2svg/d2svg.go
alixander: it is, it's replacing this icon

PR #557, opened 2022-12-29
docs: add py-d2

MrBlenny:


alixander (approved):
very cool!

As an aside, I find it unfortunate that Github forever marks a repo when it uses a template. 
Screen Shot 2022-12-29 at 9 02 44 AM

This happened to another repo of D2, and I just nuked and re-made a repo, giving credit in a NOTICE instead (https://github.com/terrastruct/text-to-diagram-site/blob/master/NOTICE.txt). No action here from you, just sharing in case it bothered you too.

alixander:
Ah, do you mind setting up commit signing and re-committing?

Screen Shot 2022-12-29 at 9 06 12 AM

You'll see a "verified" badge in your commits when it's signed

Screen Shot 2022-12-29 at 9 07 05 AM

PR #561, opened 2022-12-30
docs: add py-d2

MrBlenny:
Now with a signed commit!


alixander (approved):
thanks for bearing with that. would you mind adding to docs as well (d2lang.com)?

https://github.com/terrastruct/d2-docs/blob/master/docs/tour/extensions.md

PR #577, opened 2022-12-31
add go version to contributing.md

alixander:

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -39,6 +39,7 @@ The simplified D2 flow at a package level looks like:
 
 ## Logistics
 
+- Use Go 1.18. Go 1.19's autofmt inexplicably strips spacing from ASCII art in comments.
 - D2 uses Issues as TODOs. No auto-closing on staleness.
 - Branch off `master`.
 - Prefix pull request titles with a short descriptor of the domain, e.g. `d2renderer: Add


docs/CONTRIBUTING.md
```diff
@@ -39,6 +39,7 @@ The simplified D2 flow at a package level looks like:
 
 ## Logistics
 
+- Use Go 1.18. Go 1.19's autofmt inexplicably strips spacing from ASCII art in comments.
```
nhooyr, line 42: Let's fix this instead. If we switch all instances to `/* */` comments, then we'll be fine I think. It's clear the Go team is sticking with this abhorrent and garbage change but we can't stay on Go 1.18 forever.

nhooyr (changes requested):

docs/CONTRIBUTING.md
alixander: what about we wait til go 1.20 (~1 month wait: https://tip.golang.org/doc/go1.20). if they still have this behavior, we'll jump to that version and change to block comments

nhooyr: I think they've made it very clear there will be no change and if they did want to change, it would have been a minor release months ago. But we have no urgency to upgrade either and so waiting is fine.

nhooyr (approved):

docs/CONTRIBUTING.md
alixander: i couldn't find any documentation that this change would happen in go 1.19. i don't think u can count on the lack of announcement as clarity.

nhooyr: > i couldn't find any documentation that this change would happen in go 1.19. i don't think u can count on the lack of announcement as clarity.

It's right here in the release notes: https://tip.golang.org/doc/go1.19#go-doc

And then if you dig through the github issues opened about it they don't consider it backwards incompatible because if you rendered your documentation with `godoc` previously, the comments wouldn't clearly match their render on `godoc` but now they do.

And so we were "wrong" for relying on 50 year old conventions in code comments and must be punished for not using godoc which is only really used in open source and even then is frankly complete garbage now with the redesign at https://pkg.go.dev

alixander: i see. argh, this also isn't something we can reliably do a one-off command to change all huh. since we don't know if the comment is ascii art or not. i suppose we can just slowly start migrating

alixander: the potential for ⭐ performance improvements ⭐ is enticing to warrant upgrading. i suppose D2 itself does not have thaat much ascii art compared to TALA

Issue #384, opened 2022-12-06
Add Crows Feet to Arrowhead options


Rjvs:
Hi, although documentation claims D2 is useful for ERDs, it would be useful to have more options for the arrowheads, such as [crows feet](https://vertabelo.com/blog/crow-s-foot-notation/).

suhr:
Crow's foot is a more compact and visual notation than labels.

alixander:
closed by https://github.com/terrastruct/d2/pull/578

@suhr you had an opinion on rendering, how's it look to you?

suhr:
Proportions are rather unusual. Compare with this images:

![entity-relationship-diagram-1536x781](https://user-images.githubusercontent.com/65870/210457549-c5421dbf-27d8-44b5-9f78-0cadb2f1635e.png)
![php7wjEYw](https://user-images.githubusercontent.com/65870/210457568-89a9a0c5-eb4c-4c1b-9785-17be66e3e72d.png)

The first picture is from a database design tool: https://www.datensen.com/blog/er-diagram/crows-foot-notation-in-erd/

suhr:
There's an another example: ![schema](https://user-images.githubusercontent.com/65870/210458301-1ee13f69-0414-4bb1-99f8-e84dec1074d4.svg)

martinjirku:
I was trying to replicate it from the [blog](https://vertabelo.com/blog/crow-s-foot-notation/) mentioned in the original post.
If there is a need, it can be definitely improved.

alixander:
yeah i'm interested to see what it looks like in practice. 

i don't think the existing implementations out there are the gold standard by any means:

<img width="114" alt="Screen Shot 2023-01-04 at 11 27 40 AM" src="https://user-images.githubusercontent.com/3120367/210633934-182a4694-bfc5-42e1-bceb-68d7b15af240.png">

PR #578, opened 2022-12-31
d2renderer: #384 The arrowhead crow feet variants

martinjirku:
Introduce new styles for arrowheads #384:
  - crows-foot-many-required
  - crows-foot-many-optional
  - crows-foot-one-required
  - crows-foot-one-optional

# Open questions

I do not like `crows-foot-*` repeated 4 times, but it seems the simplest way to implement this, though d2 DSL for crows-foot is not the best DX.

# Screenshots
This SVG:
foot

is generated by this d2:
```d2
a <-> b: crows-foot-many-optional {
  source-arrowhead: {
    shape: crows-foot-many-optional
  }
  target-arrowhead: {
    shape: crows-foot-many-optional
  }
}

c <--> d: crows-foot-many-required {
  source-arrowhead: {
    shape: crows-foot-many-required
  }
  target-arrowhead: {
    shape: crows-foot-many-required
  }
}

e <--> f: crows-foot-one-required {
  source-arrowhead: {
    shape: crows-foot-one-required
  }
  target-arrowhead: {
    shape: crows-foot-one-required
  }
}

g <--> h: crows-foot-one-optional {
  source-arrowhead: {
    shape: crows-foot-one-optional
  }
  target-arrowhead: {
    shape: crows-foot-one-optional
  }
}
```


alixander (changes requested):
nice! these look great!

can you an e2e test to https://github.com/terrastruct/d2/blob/master/e2etests/stable_test.go?

i wonder if we can just remove the "crows-foot" to these names. and also the omission of "required" to signify optional.
```
many
many-required
one
one-required
```

any thoughts? i need to do some research to see if this is ambiguous, like if there's other symbols that people would expect with these keywords.

alixander:
or maybe `cf` for shorthand? `cf-one`

martinjirku:
`cf-one | cf-many | cf-many-required | cf-one-required` is definitely less to write. I am always afraid of abbreviations, but if it will be well documented, then it could be OK.

alixander:
let's go with that then

alixander:
@martinjirku could you sign your commits please? we have this requirement turned on in our org: https://github.com/terrastruct/d2/pull/557#issuecomment-1367468730 .

d2renderers/d2svg/d2svg.go
```diff
@@ -113,6 +113,15 @@ func arrowheadDimensions(arrowhead d2target.Arrowhead, strokeWidth float64) (wid
 	case d2target.DiamondArrowhead:
 		widthMultiplier = 11
 		heightMultiplier = 9
+	case d2target.CfOne:
+		fallthrough
+	case d2target.CfMany:
+		fallthrough
+	case d2target.CfOneRequired:
+		fallthrough
+	case d2target.CfManyRequired:
```
alixander, line 122: let's use commas instead of fallthroughs, little less words

alixander (changes requested):
can you add a line to changelog? (https://github.com/terrastruct/d2/blob/master/ci/release/changelogs/next.md)

you can follow the format of previous changelogs: https://github.com/terrastruct/d2/blob/master/ci/release/changelogs/v0.1.4.md?plain=1

d2renderers/d2svg/d2svg.go
martinjirku: done.

martinjirku:
Finally, I've successfully signed my previous commits. Can you re-check?

alixander (changes requested):
```
g <--> h: {
  source-arrowhead: {
    shape: cf-one
  }
  target-arrowhead: {
    shape: cf-one
  }
}
h -> e -> a: {
  target-arrowhead: {
    shape:cf-many
  }
}
g -> e -> a: {
  target-arrowhead: {
    shape:cf-many-required
  }
}
```
Screen Shot 2022-12-31 at 10 23 52 AM


1. The `cf-many` looks slightly off (black circle). 
2. Also when the arrows approach at an angle, the crows feet don't match up to the border (red). 

(1) should be fixed here, but i'm not sure what to do about (2). To extend these at an angle would also look weird. I'd say it's acceptable for now.

alixander:
Screen Shot 2022-12-31 at 10 28 09 AM

hmm actually i just noticed it's also off for other shapes too. this might be tricky to solve, you'd have to basically curve the crows foot too. i'll just put an issue to track

https://github.com/terrastruct/d2/issues/581

d2renderers/d2svg/d2svg.go
```diff
@@ -220,6 +223,47 @@ func arrowheadMarker(isTarget bool, id string, connection d2target.Connection) s
 				width*0.6, height*7/8,
 			)
 		}
+	case d2target.CfOne, d2target.CfMany, d2target.CfOneRequired, d2target.CfManyRequired:
+		attrs := fmt.Sprintf(`class="connection" stroke="%s" stroke-width="%d" fill="white"`, connection.Stroke, connection.StrokeWidth)
```
alixander, line 227: i think you should fill with connection.Stroke instead of just white. like how `FilledDiamondArrowhead` does

d2renderers/d2svg/d2svg.go
```diff
@@ -220,6 +223,47 @@ func arrowheadMarker(isTarget bool, id string, connection d2target.Connection) s
...
+			)
+		} else {
+			modifier = fmt.Sprintf(`<circle %s cx="%f" cy="%f" r="%f" />`,
+				attrs,
+				offset/2.0+1.0, height/2.0,
+				offset/2.0,
+			)
+		}
+		if !isTarget {
+			attrs = fmt.Sprintf(`%s transform="scale(-1) translate(-%f, -%f)"`, attrs, width, height)
```
alixander, line 246: nice method of flipping

d2renderers/d2svg/d2svg.go
```diff
@@ -220,6 +223,47 @@ func arrowheadMarker(isTarget bool, id string, connection d2target.Connection) s
...
+				attrs, modifier,
+				offset+2.0, height/2.0,
+				width+offset, height/2.0,
+				offset+2.0, height/2.0,
+				width+offset, 0.,
+				offset+2.0, height/2.0,
+				width+offset, height,
+			)
+		} else {
+			path = fmt.Sprintf(`<g %s>%s<path d="M%f,%f %f,%f M%f,%f %f,%f" /></g>`,
```
alixander, line 259: get rid of the extra space in `%f" />` plz (and in other places in this PR too)

e2etests/stable_test.go
```diff
@@ -1724,6 +1724,42 @@ code.height: 512
 package.height: 512
 `,
 		},
+		{
+			name: "crow_foot_arrowhead",
```
alixander, line 1728: can you add some styles on these too? e.g. stroke-width and stroke

alixander (changes requested):

d2renderers/d2svg/d2svg.go
alixander: oh wait it's supposed to be hollow. in that case just "transparent"

martinjirku: The problem with transparent is this:
image

martinjirku:
> 1. The `cf-many` looks slightly off (black circle).
> 2. Also when the arrows approach at an angle, the crows feet don't match up to the border (red).
> 
> (1) should be fixed here, but i'm not sure what to do about (2). To extend these at an angle would also look weird. I'd say it's acceptable for now.

Is this acceptable as solution for (1)?
image

alixander:
nice, yeah that's a good solution for now

d2renderers/d2svg/d2svg.go
alixander: oh interesting. the correct thing to do would be to make a mask then (like we do for labels so that connections don't overlap). but i'm okay with doing that later and keeping white for now.

e2etests/stable_test.go
martinjirku: I've added styles on some of those, and I did introduce a slightly complicated connection. Is it what you expected?

alixander (approved):
awesome! is this ready to merge?

martinjirku:
> awesome! is this ready to merge?

Yes, ready to go 🚀

PR #601, opened 2023-01-03
export: add fit to screen mode to exported svg

berniexie:
Also fixes the issue where it refits to screen on every watch refresh.

scripts in svgs won't run when they're embeded in an html document so the script won't run twice in watch mode.


d2renderers/d2svg/d2svg.go
```diff
@@ -1124,6 +1127,8 @@ func Render(diagram *d2target.Diagram, opts *RenderOpts) ([]byte, error) {
 ]]>
 </style>`, styleCSS, styleCSS2))
 
+	buf.WriteString(fmt.Sprintf(`<script type="application/javascript"><![CDATA[%s]]></script>`, fitToScreenScript))
```
alixander, line 1130: i don't think this should be default. there might be many SVGs on a single page, we don't want all of them running JS. This should be a flag, `--fit`. even then, i'm not so sure it's that useful. it just doesn't feel like something SVG should be doing. @nhooyr wdyt

nhooyr: This JS only runs if the SVG is opened as the root document. In the case of multiple SVGs on a single page, it will not run.

I'm personally in favour as I want to see the whole diagram when I first open the SVG.

alixander: > This JS only runs if the SVG is opened as the root document

i missed this, what specifies this behavior?

nhooyr: idk exactly but that's just how it works.

alixander: it looks like it won't run as ``, but it will if you embed, which is still a common use case that this would break if it's default (we embed in d2lang.com).

nhooyr: Interesting. @berniexie I thought you tried it in `--watch` and it didn't run twice?

alixander: i didn't verify this. it's info i got from http://thenewcode.com/1094/Using-JavaScript-in-SVG

berniexie: @alixander locally when i tried to run it with it embeded, it did not run. Where does it say it will run?

alixander: Screen Shot 2023-01-03 at 1 58 02 PM

nhooyr: Perhaps it did run @berniexie but only on the initial load. Then when the SVG is dynamically updated, the embedded JS is ignored.

nhooyr: What if we made it so that this script only runs in `--watch` and when you open it standalone?

For `--watch`, we could set a data attribute on the HTML element to indicate to the script that it's running in `--watch`. And for standalone, we can add an ID to our svg and then you can check whether the SVG matched by the ID is the same as `document.rootElement`.

alixander: that'd work

berniexie: Fascinating. I thought the script didn't run because of what Anmol said, but digging deeper,  I think it actually doesn't run because we use innerHTML to embed the svg in watch mode. https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

Screen Shot 2023-01-03 at 2 15 31 PM

If this is behavior we want, I don't think there's another way to do this besides adding this behavior only on a flag.

berniexie: @nhooyr oooh ur standalone idea works for me

berniexie:
@alixander what do we do about the sketch tests

d2renderers/d2svg/fitToScreen.js
```diff
@@ -1,5 +1,10 @@
 window.addEventListener("DOMContentLoaded", () => {
   const svgEl = document.querySelector("svg");
+  const svgID = "d2-svg";
```
nhooyr, line 3: I meant do this when generating the svg. Otherwise, our svg could be embedded inside another svg and this code will still run.

nhooyr (changes requested):

nhooyr (approved):
nice

nhooyr (commented):
maybe add a comment about why it doesn't run again with `--watch`.

alixander (approved):

PR #652, opened 2023-01-12
implement `animated` keyword

alixander:
animation speed based on stroke width and dash gap size

sketch exp (10)

there's an issue there with the crows foot, but it's edge case, i'll defer in an issue: https://github.com/terrastruct/d2/issues/653

it'd be cool for bidirectional connection to animate in both directions: https://github.com/terrastruct/d2/issues/654

sketch looks imperfect: https://github.com/terrastruct/d2/issues/655

sketch exp (11)


bo-ku-ra:
scrolling;
<-> : From the center to each end
--  :  From each end to the center

difficult?
(when editing, top and bottom, right and left are switched frequently.)

alixander:
@bo-ku-ra I have an issue (see PR description) for the first one (`<->`), but it's not clear to me the second one (`--`) means end to center. It's also unclear what each end to the center animation means.

> (when editing, top and bottom, right and left are switched frequently.)

Sorry didn't understand this line, can you rephrase?

bo-ku-ra:
sorry, i made a mistake in writing issue. (It was #654)
what should I do?

alixander:
no worries, up to you =)

PR #668, opened 2023-01-15
cli: Make bundle option default to false

### LINKED ISSUE #667: Seems like `-b`/`--bundle` cannot be unset
The CLI option `-b`/`--bundle` is documented as defaulting to true. However, it seems to me that it's not possible to unset it.

Trying to set it to false with an argument results in an error:
```sh
$ d2 --bundle false input.d2 output.svg
err: bad usage: too many arguments passed
```

OTOH, the `-d`/`--debug` option defaults to false, and setting it (`d2 --debug input.d2 output.svg`) enables debug output.

So I guess the bundle option should either accept an argument or should default to false to turn it into an actual option that can be influenced.

homeworkprod:
If it would default to true (like it did), it could not be unset.

Should fix #667.

### DIFF
--- a/ci/release/changelogs/next.md
+++ b/ci/release/changelogs/next.md
@@ -16,3 +16,4 @@
 - Fixes arrowheads sometimes appearing broken with sketch on. [#656](https://github.com/terrastruct/d2/pull/656)
 - Fixes code snippets not being tall enough with leading newlines. [#664](https://github.com/terrastruct/d2/pull/664)
 - Icon URLs that needed escaping (e.g. with ampersands) are handled correctly by CLI. [#666](https://github.com/terrastruct/d2/pull/666)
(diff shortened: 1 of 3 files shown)


alixander:
@homeworkprod thank you! Looks good. Would you mind signing your commit (https://github.com/terrastruct/d2/blob/master/docs/CONTRIBUTING.md#logistics)?

homeworkprod:
I don't usually sign commits and as such don't have a key.

Is signing a hard requirement?

alixander:
Yes, it's an org-wide setting.

homeworkprod:
The commit should be verified now.

alixander:
sorry my mistake @homeworkprod , we should go with the documentation change instead of changing this flag as discussed in the issue.

homeworkprod:
Makes sense, no worries.

PR #677, opened 2023-01-19
fix opacity for labels

alixander:
applies opacity to group instead of everything inside the group.

I feel like there was a reason we didn't do this originally, but can't remember. @gavin-ts do u remember anything?

Screen Shot 2023-01-19 at 12 26 45 AM


gavin-ts (approved):
looks good to me, will this make tooltips transparent too though? there was related #246 which still seems good. I think originally we might have wanted the labels to always be opacity 1 and have the style just apply to the shape or connection itself?

alixander:
> will this make tooltips transparent too though?

i think it will, which seems like what ppl would intend

PR #690, opened 2023-01-19
Add xdg-open along with open and friendly error output on REPORT_OUTPUT failure

exapsy:


alixander:
could you set up commit signing please? https://github.com/terrastruct/d2/pull/557#issuecomment-1367468730

ci/e2ereport.sh
```diff
@@ -9,7 +9,9 @@ FORCE_COLOR=1 DEBUG=1 go run ./e2etests/report/main.go "$@";
 
 if [ -z "${NO_OPEN:-}" ]; then
   if [ -s "$REPORT_OUTPUT" ]; then
-    open "$REPORT_OUTPUT"
+    open "$REPORT_OUTPUT" 2>/dev/null || \
```
nhooyr, line 12: Use `command -v` to check if they exist before opening them.

With `2>/dev/null` you won't get any error messages even if they're legit.

nhooyr: Grep other scripts for `command -v` to see how it's used.

nhooyr (changes requested):

ci/e2ereport.sh
exapsy: Addressed on e8906ef :D

exapsy:
Closing to ensure signed commits from my side

Issue #659, opened 2023-01-13
resizing script makes SVGs un-embeddable


alixander:
To get interactivity for SVGs within web pages, you need <object> embed. When you use <object> embed, it's a separate document model created for the SVG. So the script in the SVG thinks its root and tries to resize incorrectly.

we should make it an opt-in flag

alixander:
fyi @nhooyr

nhooyr:
Does it think that the whole web page is its width/height? If so, why don't we just change it to use the root element width/height? Then that should work fine even if embedded.

alixander:
no it thinks its containing element is. i set <object width="100%"... and it turned out like

<img width="1276" alt="Screen Shot 2023-01-13 at 12 26 22 PM" src="https://user-images.githubusercontent.com/3120367/212412984-2f059975-78da-472c-9f57-0aebe7b2eddb.png">

but actually i realized i can just set height and it works.

nvm then, we should just make a note of it in our docs, like in the troubleshooting section

homeworkprod:
Is the embedded resize script really necessary at all? I was actually very surprised to find it in the SVG output.

IIRC, specifying `viewBox` but not `width` and `height` on an SVG makes it auto-scale in current browsers while preserving aspect ratio. Constraints can then be applied to its container (`<object>`, `<img>`, wrapper `<div>`, etc.) as necessary.

This article seems to cover it quite well: https://css-tricks.com/scale-svg/

alixander:
it's a long article, can you reference the part where it says it auto-scales to fit to browser?

The use case here is giving users a pleasant experience when they open the SVG file standalone in the browser. AFAIK, without a script, there's no behavior where it auto-fits.

homeworkprod:
> it's a long article, can you reference the part where it says it auto-scales to fit to browser?

The last paragraph of https://css-tricks.com/scale-svg/#aa-the-height-and-width-attributes mostly sums it up for me.

However, there'll probably still be cases this won't fully solve. It's a complicated mess, even if we don't have to care for MSIE anymore.

> The use case here is giving users a pleasant experience when they open the SVG file standalone in the browser. AFAIK, without a script, there's no behavior where it auto-fits.

I see. Yeah, that's my experience as well, unfortunately. And in addition, at least Firefox's zoom functionality doesn't work for me with standalone SVGs.


[...]

PR #721, opened 2023-01-26
Add option for fit to screen

bluemir:
Add CLI option for disable `fit to screen` script. ( `--disable-fit-to-screen` )

It might be solution for this issue https://github.com/terrastruct/d2/issues/659#issue-1532664265

Test failed when I run `./ci/test.sh` in my laptop, but it looks like not related to this change. 
(I have no idea to fix it)

### DIFF
--- a/d2cli/main.go
+++ b/d2cli/main.go
@@ -63,7 +63,7 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
 	if err != nil {
 		return err
 	}
-	disableFitToScreenFlag, err := ms.Opts.Bool("D2_DISABLE_FIT_TO_SCREEN", "disable-fit-to-screen", "", false, "disable fit to screen")
+	fitToScreenFlag, err := ms.Opts.Bool("D2_FIT_TO_SCREEN", "fit-to-screen", "", true, "fit to screen, to disable it use`fit-to-screen=false`")
 	if err != nil {
 		return err
 	}
@@ -240,20 +240,20 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
 			return xmain.UsageErrorf("-w[atch] cannot be combined with reading input from stdin")
 		}
 		w, err := newWatcher(ctx, ms, watcherOpts{
-			layoutPlugin:       plugin,
-			sketch:             *sketchFlag,
-			center:             *centerFlag,
-			themeID:            *themeFlag,
-			darkThemeID:        darkThemeFlag,
-			pad:                *padFlag,
-			host:               *hostFlag,
-			port:               *portFlag,
-			inputPath:          inputPath,
-			outputPath:         outputPath,
-			bundle:             *bundleFlag,
-			forceAppendix:      *forceAppendixFlag,
-			disableFitToScreen: *disableFitToScreenFlag,
-			pw:                 pw,
+			layoutPlugin:  plugin,
+			sketch:        *sketchFlag,
+			center:        *centerFlag,
+			themeID:       *themeFlag,
+			darkThemeID:   darkThemeFlag,
+			pad:           *padFlag,
+			host:          *hostFlag,
+			port:          *portFlag,
+			inputPath:     inputPath,
+			outputPath:    outputPath,
+			bundle:        *bundleFlag,
+			forceAppendix: *forceAppendixFlag,
+			fitToScreen:   *fitToScreenFlag,
+			pw:            pw,
 		})
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


gavin-ts (commented):
this looks good to me but I wonder if we should have it inverted (`--fit-to-screen` / `--fit`) to enable as mentioned in #659. related discussion https://github.com/terrastruct/d2/pull/601#discussion_r1060938685 
I also wonder if we could do the fitting with just css instead of js

bluemir:
I think the '--fit-to-screen' option is better, too. 
so I change it.

gavin-ts (approved):

gavin-ts:
Since the default behavior no longer adds the js, the test svgs need to be updated with `TESTDATA_ACCEPT=1 ./ci/test.sh`

alixander:
hold off on this for a sec, need to think about this whole fit to screen thing

alixander:
now that the resizing is done with CSS, i wonder if anyone would still want to turn this off. i'll leave this up for a little longer to wait and see, but i suspect that https://github.com/terrastruct/d2/pull/725 solves all resizing woes.

gavin-ts:
#1080

alixander:
hey @bluemir sorry for the looong delay. had to see if anyone was still interested in this after a bunch of fixes. https://github.com/terrastruct/d2/issues/1080 is motivating enough. do you still have interest in making this change? if so, perhaps it'd be easier to just start a new PR given all the merge conflicts.

d2renderers/d2svg/d2svg.go
```diff
@@ -1838,18 +1839,21 @@ func Render(diagram *d2target.Diagram, opts *RenderOpts) ([]byte, error) {
...
+	fitScreenWrapper := fmt.Sprintf(`<svg %s d2Version="%s" preserveAspectRatio="%s meet" viewBox="0 0 %d %d"%s>`,
 		`xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"`,
 		version.Version,
 		alignment,
 		w, h,
 		dimensions,
 	)
+	if !opts.FitScreen {
+		fitScreenWrapper = ""
+	}
```
gavin-ts, line 1851: I think this should be the same as `setDimensions` above now. setting dimensions disables the fit to screen behavior.

gavin-ts (changes requested):

d2renderers/d2svg/d2svg.go
gavin-ts: I think we also want to rename the option to  `--no-fit` or `--sized` since we want the fit behavior as default now

d2cli/main.go
```diff
@@ -422,7 +422,7 @@ func _render(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, sketc
 		ThemeID:       themeID,
 		DarkThemeID:   darkThemeID,
 		SetDimensions: toPNG,
-		FitScreen:     fitScreen,
+		NoFit:         noFit,
```
gavin-ts: Since NoFit does the same thing as `SetDimensions`, I think it is better to just have `SetDimensions: toPNG || noFit,` without another RenderOpt

d2renderers/d2svg/d2svg.go
```diff
@@ -1830,15 +1831,15 @@ func Render(diagram *d2target.Diagram, opts *RenderOpts) ([]byte, error) {
 	}
 
 	var dimensions string
-	if setDimensions {
+	if setDimensions && !opts.NoFit {
 		dimensions = fmt.Sprintf(` width="%d" height="%d"`, w, h)
 	}
```
pangloss, line 1836: I experimented with applying this change in a diagram and learned that it does not do what I expected. Please see the note here: https://github.com/terrastruct/d2/issues/1080#issuecomment-1482221132

d2cli/watch.go
```diff
@@ -360,7 +361,7 @@ func (w *watcher) compileLoop(ctx context.Context) error {
 			w.pw = newPW
 		}
 
-		svg, _, err := compile(ctx, w.ms, w.layoutPlugin, w.sketch, w.center, w.pad, w.themeID, w.darkThemeID, w.inputPath, w.outputPath, w.bundle, w.forceAppendix, w.pw.Page)
+		svg, _, err := compile(ctx, w.ms, w.layoutPlugin, w.sketch, w.center, w.pad, w.themeID, w.darkThemeID, w.inputPath, w.outputPath, w.bundle, w.forceAppendix, w.noFit, w.pw.Page)
```
gavin-ts, line 364: if we want this to affect watch mode we will need to update the js code, but that should be a follow up PR that addresses #1080.

alixander (changes requested):
@bluemir if you want to fix the merge conflicts, we can get this merged, thank you!

PR #733, opened 2023-01-27
fmt: Accept multiple files to be formatted

### LINKED ISSUE #718: d2 fmt: Accept multiple files to be formatted

nhooyr:
Closes #718

### DIFF
--- a/fmt.go
+++ b/fmt.go
@@ -17,26 +17,24 @@ func fmtCmd(ctx context.Context, ms *xmain.State) (err error) {
 
 	ms.Opts = xmain.NewOpts(ms.Env, ms.Log, ms.Opts.Flags.Args()[1:])
 	if len(ms.Opts.Args) == 0 {
-		return xmain.UsageErrorf("fmt must be passed the file to be formatted")
-	} else if len(ms.Opts.Args) > 1 {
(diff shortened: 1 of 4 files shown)


fmt.go
```diff
@@ -17,26 +17,24 @@ func fmtCmd(ctx context.Context, ms *xmain.State) (err error) {
...
-		return xmain.UsageErrorf("fmt accepts only one argument for the file to be formatted")
+		return xmain.UsageErrorf("fmt must be passed at least one file to be formatted")
 	}
 
-	inputPath := ms.Opts.Args[0]
-	input, err := ms.ReadPath(inputPath)
-	if err != nil {
-		return err
+	for _, inputPath := range ms.Opts.Args {
+		input, err := ms.ReadPath(inputPath)
```
alixander, line 24: should we check that they're d2 extensions? that way someone can just do `d2 fmt docs/*` and it'd ignore the svgs (idk if globs actually expand like that, might be wrong)

nhooyr: Nah, file extensions are nominal on UNIX unlike Windows/macOS.

Someone could just glob like this instead `d2 fmt docs/*.d2`.

alixander (approved):

PR #763, opened 2023-02-05
Fix padding

alixander:
fix regression made in https://github.com/terrastruct/d2/pull/702 where setting just width/height would cause padding in the unset dimension to be lost.

Screen Shot 2023-02-05 at 2 00 07 AM

Screen Shot 2023-02-05 at 2 01 51 AM


e2etests/testdata/stable/people/dagre/board.exp.json
```diff
@@ -294,10 +294,10 @@
       "type": "person",
       "pos": {
         "x": 0,
-        "y": 280
+        "y": 260
       },
-      "width": 18,
-      "height": 21,
+      "width": 41,
+      "height": 61,
```
gavin-ts, line 300: the padding isn't set for either dimensions when a dimension is set because even the opposite dimension can affect the minimum size of a shape. See this person shape which had a desiredWidth of 16 but now is much further from that due to the padding on the y-axis.

gavin-ts:
changes make sense

> fix regression made in #702 where setting just width/height would cause padding in the unset dimension to be lost.

but this was intentional for when a small width/height is desired (see comment above)

PR #772, opened 2023-02-06
classes

### LINKED ISSUE #102: syntax: classes
classes should aggregate styles so you don't have to apply the same set of styles in a bunch of places

### LINKED ISSUE #566: Proposal: Add free-form 'tagging' of objects and relationships for plugin interpretation
_(Based on some brainstorms in discord, [#feature-requests, 29 december](https://discord.com/channels/1039184639652265985/1039185500088569907/1057973370727178251))_

As a plugin developer, I would like to define extra attributes for objects and relationships.
I could develop a plugin that renders d2 as a BMPN graph. I would then like to know which objects are a activity, which are a gateway, event, etc. This information is not encoded in d2's grammar. It's probably good to keep d2 agnostic about graph specifications such as UML and BPMN. But it would be nice if d2 could pass 'hints' to plugins, for which the interpretation is completely up to the plugin developer.

The way to do this currently would be to use comments. Ofcourse this is not very clean. And since it's not formally specified, prone to breaking changes.

An inspiration for this system may be how in the Go language struct fields are tagged. The Go languages doesn't know anything about json, but the `encoding/json` package can obtain information from a tag. `json:"my_field"`

```d2
customer `bpmn:"pool"` {
  purchase: Purchase item `bpmn:"task"`
}

cashier `bpmn:"pool"` {
  handle: Handle purchase `bmpn:"task"`
}


customer.purchase -> cashier.handle: Purchase an item `bpmn:"sequence"`
```

alixander:
```
classes: {
  dragon_ball: {
    label: ""
    shape: circle
    style.fill: orange
    style.stroke-width: 0
    width: 50
  }
}
nostar: { class: dragon_ball }
1star: { label: "*"; class: dragon_ball }
2star: { label: "**"; class: dragon_ball }
```

Screen Shot 2023-04-03 at 8 47 16 PM

```
classes: {
  orb: {
    shape: circle
    style.fill: red
    style.stroke-width: 0
    label: ""
  }
}

a.class: orb
b.class: orb
c.class: orb
d.class: orb

steps: {
  1: {
    classes.orb.style.fill: blue
  }
  2: {
    classes.orb.style.fill: green
  }
  3: {
    classes.orb.style.fill: yellow
  }
}
```

https://user-images.githubusercontent.com/3120367/230550664-f4e5a4a3-3c1e-4289-a12d-9bfa73bbbc49.mp4

Also writes classes to SVGs, which can be a workflow to tag things in D2 for post processing. closes #566  

Followups:
- https://github.com/terrastruct/d2/issues/1135

closes #102


alixander:
@nhooyr this isn't quite ready for merging yet but would like a review of this initial implementation

alixander:
@nhooyr bump, i wanna release tmr morning

d2compiler/compile.go
```diff
@@ -159,7 +170,24 @@ func (c *compiler) compileField(obj *d2graph.Object, f *d2ir.Field) {
 		return
 	}
 	_, isReserved := d2graph.SimpleReservedKeywords[keyword]
-	if isReserved {
+	if f.Name == "classes" {
+		if f.Map() != nil {
+			if len(f.Map().Edges) > 0 {
+				c.errorf(f.LastRef().AST(), "classes cannot contain an edge")
```
nhooyr, line 176:
```suggestion
				c.errorf(f.Map().Edges[0].LastRef().AST(), "classes cannot contain an edge")
```

d2compiler/compile.go
```diff
@@ -116,15 +116,26 @@ func (c *compiler) compileBoardsField(g *d2graph.Graph, ir *d2ir.Map, fieldName
...
 func (c *compiler) compileMap(obj *d2graph.Object, m *d2ir.Map) {
+	class := m.GetField("class")
+	if class != nil {
+		className := class.Primary()
+		if className == nil {
+			c.errorf(class.LastRef().AST(), "class missing value")
+		} else {
+			classMap := m.GetClosestClassMap(className.String())
+			if classMap != nil {
+				c.compileMap(obj, classMap)
```
nhooyr, line 135: This isn't ideal. If you make a change to the class after assigning the class to a shape, I'd expect the adjusted class to apply to the shape. I think we should use overlay instead here and overlay shape maps onto their classes at the very end.

d2compiler/compile.go
```diff
@@ -504,6 +528,29 @@ func (c *compiler) compileEdge(obj *d2graph.Object, e *d2ir.Edge) {
...
+func (c *compiler) compileEdgeMap(edge *d2graph.Edge, m *d2ir.Map) {
+	class := m.GetField("class")
+	if class != nil {
+		className := class.Primary()
+		if className == nil {
+			c.errorf(class.LastRef().AST(), "class missing value")
+		} else {
+			classMap := m.GetClosestClassMap(className.String())
+			if classMap != nil {
+				c.compileEdgeMap(edge, classMap)
```
nhooyr, line 540: What if a class contains `class: X` where X is the ID of the class itself?

d2ir/compile.go
```diff
@@ -22,12 +22,50 @@ func Compile(ast *d2ast.Map) (*Map, error) {
...
+		if lf.Map() == nil || lf.Primary() != nil {
+			c.errorf(lf.References[0].Context.Key, "invalid layer")
+			continue
+		}
+		l := lf.Map()
+		lClasses := l.GetField("classes")
+
+		if lClasses == nil {
+			lClasses = classes.Copy(l).(*Field)
+			l.Fields = append(l.Fields, lClasses)
```
nhooyr, line 57: Isn't this too late? By this time we've already interpreted all the classes in the layer?

d2ir/d2ir.go
```diff
@@ -601,6 +601,18 @@ func (m *Map) EdgeCountRecursive() int {
 	return acc
 }
 
+func (m *Map) GetClosestClassMap(name string) *Map {
```
nhooyr, line 604:
```suggestion
func (m *Map) GetClassMap(name string) *Map {
```

This doesn't recurse up so Closest isn't accurate.

nhooyr (changes requested):

d2compiler/compile.go
alixander: oh good call, will add a test for that

d2compiler/compile.go
alixander: i don't understand. can you say again?

nhooyr:
```
classes: {
  first: {
    class: first
  }
}

my shape: {
  class: first
}
```

Will that cause an infinite loop?

d2ir/compile.go
alixander: oh hm, i guess a test like this is what u mean
```
classes: {
  orb: {
    style.fill: yellow
  }
}
layers: {
  x: {
    classes.orb.style.stroke: red
  }
}
```

d2compiler/compile.go
alixander: ohh hm i think we should prevent `class` from appearing in `classes`. 

i thought maybe they'd want to use it to construct new classes off a template, but that's not how classes are used. you'd just apply two classes to one shape to achieve that

nhooyr: Sounds good to me.

d2ir/compile.go
nhooyr: Nah I mean this:

```
classes: {
  orb: {
    style.fill: yellow
  }
}
layers: {
  x: {
    class: orb
  }
}
```

Wouldn't `orb` be undefined in `layers.x` as we don't add it until after compilation?

alixander: no, d2ir's compile actually doesn't look at the `class` keyword at all. the interpretation happens later

nhooyr: Oh right!

d2compiler/compile.go
alixander: i thought this would fail, but this actually passes:
https://github.com/terrastruct/d2/pull/772/commits/5dd273da18edf9b53be8c9165e5b171b0f6dbcc9

i guess d2ir must already merge

nhooyr: Yea makes sense. I didn't realize some of the code changes were in `d2compiler` for some reason.

d2ir/compile.go
alixander: (the example i gave of what i thought u meant remains an issue though)

nhooyr: Why? Don't you overlay if classes already exists just below that code?

alixander: ah fuck, was just a typo in the test case i had. https://github.com/terrastruct/d2/pull/772/commits/352a6a36d6043deefebf54b3b38c704030150e32

you're right. 

d2ir is 👨‍🍳

nhooyr (approved):
🚀

PR #783, opened 2023-02-09
fix steps panic

alixander:
just added the panicking test. i know it's not valid, but it shouldn't panic


nhooyr (approved):
oh yes definitely should not should be an error. I guess somewhere I'm assuming each child of a steps is a map.

alixander:
> I guess somewhere I'm assuming each child of a steps is a map.

your hunch is correct

d2ir/compile.go
```diff
@@ -66,6 +66,9 @@ func (c *compiler) compileSteps(m *Map) {
 		if i == 0 {
 			base = m.CopyBase(sf)
 		} else {
+			if steps.Fields[i-1].Map() == nil {
```
nhooyr, line 69: Check if there's a scalar/array too and if so, error.

nhooyr (changes requested):

d2ir/compile.go
nhooyr: None of the compiler functions return an error. You use `c.errorf()` to report it and skip the offending section.

nhooyr (changes requested):

d2ir/compile.go
nhooyr: my main point though is that a step *must* be either a map or not set to any value.

d2ir/compile.go
```diff
@@ -59,13 +60,18 @@ func (c *compiler) compileSteps(m *Map) {
...
-		if sf.Map() == nil {
+		if sf.Map() == nil || sf.Primary() != nil {
+			c.errorf(sf.References[0].Context.Key, "invalid step")
 			continue
 		}
 		var base *Map
 		if i == 0 {
 			base = m.CopyBase(sf)
 		} else {
+			if steps.Fields[i-1].Map() == nil || steps.Fields[i-1].Primary() != nil {
```
nhooyr, line 71: Remove this and break above instead. This error isn't accurate anyway and reports the wrong step as being invalid.

nhooyr (approved):
✅

Issue #594, opened 2023-01-02
docker: playwright not installed 


avelino:
In the official docker image, playwright is not installed by default and fails to install automatically when in use

```
$ /usr/local/bin/d2 --sketch -l elk docs/tech/orgchart/orgchart.d2 orgchart.png
2023-01-02 17:50:02.672 [INFO]	(default.stdlib)	<./../../github.com/playwright-community/playwright-go/run.go:107>	(*PlaywrightDriver).DownloadDriver	Downloading driver to /root/.cache/ms-playwright-go/1.20.0-beta-1647057403000
err: failed to install Playwright: could not install driver: could not install driver: could not download driver: Get "https://playwright.azureedge.net/builds/driver/next/playwright-1.20.0-beta-1647057403000-linux.zip": x509: certificate signed by unknown authority
ERROR: Job failed: command terminated with exit code 1
```

**installation ([ref](https://github.com/playwright-community/playwright-go#installation)):**
```
go run github.com/playwright-community/playwright-go/cmd/playwright install --with-deps
```

avelino:
@nhooyr is there any prediction when we will release this update?

alixander:
@avelino in a few days after https://github.com/terrastruct/d2/pull/784 is merged! (which fixes our release process)

PR #784, opened 2023-02-10
docker: Run image build on amd64 instance to avoid QEMU errors on arm64

nhooyr:
Playwright won't install in a amd64 docker container running under QEMU
on arm64.


alixander:
this is for fixing our build right? (which was broken when we fixed https://github.com/terrastruct/d2/issues/594)

alixander (approved):

PR #786, opened 2023-02-10
Tune spacing

alixander:
i want to do one for sequence diagrams too, but it requires some more major refactoring that i'll save for another PR

make things slightly more compact:

dagre example: (dagre is annoying we can't go down too much b/c it has the problem of not being able to adjust container size)

Screen Shot 2023-02-09 at 7 27 04 PM

Additionally:

- elk has `thoroughness` turned down. this was making it unnecessarily slow for large diagrams for marginal gain.
- crows feet is dialed down
Screen Shot 2023-02-10 at 12 59 20 AM
- fixes a minor rendering bug for sequence diagrams


ejulio-ts (approved):
quickly checked the examples, looks good

gavin-ts (approved):
maybe also EdgeEdgeBetweenLayersSpacing 50->40?

alixander:
> maybe also EdgeEdgeBetweenLayersSpacing 50->40?

pretty sure i tried that value and liked 50 more

PR #788, opened 2023-02-10
move dagre container labels to outside position

### LINKED ISSUE #19: dagre container label overflows
```d2
aws_s3: AWS S3 California {
    Monitoring -> California
}
aws_s3: "AWS S3 San Francisco, California" {
    California.San Francisco
}
```

197423135-141668fd-6f3f-47a4-a680-dffc83872e21

alixander:
screencapture-file-Users-alexanderwang-dev-alixander-d2-e2etests-out-e2e-report-html-2023-02-10-20_44_25

we lost a bit of this visual symmetry, but the new version is still symmetrical, just doesn't look as much like it
Screen Shot 2023-02-10 at 8 45 22 PM
Screen Shot 2023-02-10 at 8 46 02 PM

fixes #19
Screen Shot 2023-02-10 at 2 06 47 PM


gavin-ts (approved):
cool I wonder if we should try reducing the label padding when outside so it is a little closer to the container shapes (e.g. 2 instead of 5?); also curious about an outside top left label placement instead of the center one; 


> we lost a bit of this visual symmetry, but the new version is still symmetrical, just doesn't look as much like it Screen Shot 2023-02-10 at 8 45 22 PM Screen Shot 2023-02-10 at 8 46 02 PM

for the asymmetry I wonder if we can see if (d's) top child (h) is also a container and then remove the inner top padding too (moving everything up?) since the label itself will cover the padding?

alixander:
i'm not 100% sure the asymmetry is 100% bad, the difference in padding makes it clear that `h` is label of child, not container.

> also curious about an outside top left label placement instead of the center one;

i was actually thinking something on the side, since dagre is so hierarchical that arrows are not likely to cross (inverse if direction changes ofc). but i also don't want long labels to go to neighbors. so, something to noodle on more.

PR #800, opened 2023-02-12
wip: Reserved keywords only unquoted

### LINKED ISSUE #477: compiler: escape reserved keywords
E.g. width and height should be usable as sql table column keys.

https://play.d2lang.com/?script=AvnbSiE4sypVoZpLQQHsaiuF4sKc-JLEpJxULgUFJbA6JSuFvNLcpNQikAgkoBBCtVyAAAAA__8%3D

alixander:
@nhooyr spot check before i go too far on this. there's a number of places where we iterate over `ida` and check reserved. those will be annoying, as we now have to change them to carry info about their original AST quoting. E.g. there's a lot of ScalarString() conversions made too early now

closes #477


d2ir/d2ir.go
```diff
@@ -687,8 +690,9 @@ func (m *Map) ensureField(i int, kp *d2ast.KeyPath, refctx *RefContext) (*Field,
 	}
 
 	f := &Field{
-		parent: m,
-		Name:   head,
+		parent:     m,
+		Name:       headString,
+		IsUnquoted: headUnquoted,
```
nhooyr, line 695: You also need to check that there are no escapes. Use `len(headString) == len(head.(*d2ast.UnquotedString).RawString)`. i.e there are no escapes if the length of the string value is equal to it's length in the AST.

nhooyr (commented):
looks reasonable otherwise to me

d2ir/d2ir.go
alixander: IsUnquoted could be confusing then, but i can't think of a word that encapsulates "is unquoted or is escaped". Raw already means something else. will leave as IsUnquoted for now unless u have a word

nhooyr: Yea just add a comment on `IsUnquoted`. I have no better word in mind.

alixander:
https://github.com/terrastruct/d2/pull/2226

PR #824, opened 2023-02-15
fix 2 bugs: compiler: underscores in edges; oracle: edge ref

alixander:
1. edge case where creating a new edge with underscores AND common prefix was panicking
2. moving an object where the most nested reference is an edge

### DIFF
--- a/d2ir/d2ir.go
+++ b/d2ir/d2ir.go
@@ -379,7 +379,9 @@ func (eid *EdgeID) Match(eid2 *EdgeID) bool {
 	return true
 }
 
-func (eid *EdgeID) resolveToCommon(m *Map) (*d2ast.KeyPath, *EdgeID, *Map, error) {
+// resolve resolves both underscores and commons in eid.
+// It returns the new eid, containing map adjusted for underscores and common ida.
+func (eid *EdgeID) resolve(m *Map) (_ *EdgeID, _ *Map, common []string, _ error) {
 	eid = eid.Copy()
 	maxUnderscores := go2.Max(countUnderscores(eid.SrcPath), countUnderscores(eid.DstPath))
 	for i := 0; i < maxUnderscores; i++ {
@@ -401,17 +403,16 @@ func (eid *EdgeID) resolveToCommon(m *Map) (*d2ast.KeyPath, *EdgeID, *Map, error
 		}
 	}
 
-	common := &d2ast.KeyPath{}
 	for len(eid.SrcPath) > 1 && len(eid.DstPath) > 1 {
 		if !strings.EqualFold(eid.SrcPath[0], eid.DstPath[0]) {
-			return common, eid, m, nil
+			return eid, m, common, nil
 		}
-		common.Path = append(common.Path, d2ast.MakeValueBox(d2ast.RawString(eid.SrcPath[0], true)).StringBox())
+		common = append(common, eid.SrcPath[0])
 		eid.SrcPath = eid.SrcPath[1:]
 		eid.DstPath = eid.DstPath[1:]
 	}
 
-	return common, eid, m, nil
+	return eid, m, common, nil
 }
 
 type Edge struct {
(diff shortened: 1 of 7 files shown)


d2ir/d2ir.go
```diff
@@ -770,7 +770,8 @@ func (m *Map) CreateEdge(eid *EdgeID, refctx *RefContext) (*Edge, error) {
 	if len(common) > 0 {
 		tmp := *refctx.Edge.Src
 		kp := &tmp
-		kp.Path = kp.Path[:len(common)]
+		underscores := countUnderscores(kp.IDA())
```
nhooyr, line 773: ooo true. Though this should instead be in `trimCommon`. or actually can you combine `resolveUnderscores` and `trimCommon`? This code should be in there.

d2oracle/edit_test.go
```diff
@@ -2001,6 +2001,32 @@ c: {
 				assert.JSON(t, len(g.Objects), 3)
 			},
 		},
+		{
+			name: "underscore-connection",
+
+			text: `a: {
+  b
```
nhooyr, line 2008: For the test case, I don't see how your fix applies. Wouldn't this have been the first reference anyway?

nhooyr (changes requested):

nhooyr:
> moving an object where the most nested reference is an edge

How is bug 2 related to underscores?

d2oracle/edit_test.go
alixander: i guess not. idk how they're ordered

alixander: i add this line

Screen Shot 2023-02-14 at 5 06 11 PM

and get this
Screen Shot 2023-02-14 at 5 06 05 PM

nhooyr: Ohh that's cause this function is being called for `c` not `a.b` lol.

I don't think it's related to underscores though, you could remove the underscores and still reproduce this I believe.

d2ir/d2ir.go
```diff
@@ -762,16 +759,12 @@ func (m *Map) CreateEdge(eid *EdgeID, refctx *RefContext) (*Edge, error) {
...
 	}
 
-	eid, m, err := eid.resolveUnderscores(m)
+	common, eid, m, err := eid.resolveToCommon(m)
 	if err != nil {
 		return nil, d2parser.Errorf(refctx.Edge, err.Error())
 	}
-	common, eid := eid.trimCommon()
-	if len(common) > 0 {
-		tmp := *refctx.Edge.Src
```
nhooyr, line 771: This was a hack to reuse `EnsureField`, common should be reverted back to `[]string`. It doesn't reflect something in the syntax. Ideally later we add a `ensureFieldIDA` that accepts `[]string` directly.

d2ir/d2ir.go
```diff
@@ -379,7 +379,7 @@ func (eid *EdgeID) Match(eid2 *EdgeID) bool {
 	return true
 }
 
-func (eid *EdgeID) resolveUnderscores(m *Map) (*EdgeID, *Map, error) {
+func (eid *EdgeID) resolveToCommon(m *Map) (*d2ast.KeyPath, *EdgeID, *Map, error) {
```
nhooyr, line 382:
```suggestion
// resolve resolves both underscores and commons in eid.
// It returns the new eid, containing map adjusted for underscores and common ida.
func (eid *EdgeID) resolve(m *Map) (_ *EdgeID, _ *Map, common []string, _ error) {
```

nhooyr (changes requested):

nhooyr:
@alixander So do you agree the second bug isn't related to underscores?

alixander:
yeah i guess not. an edge can just be the most nested. surprised this hasn't come up earlier though

d2ir/d2ir.go
```diff
@@ -762,16 +760,12 @@ func (m *Map) CreateEdge(eid *EdgeID, refctx *RefContext) (*Edge, error) {
...
 	if err != nil {
 		return nil, d2parser.Errorf(refctx.Edge, err.Error())
 	}
-	common, eid := eid.trimCommon()
 	if len(common) > 0 {
-		tmp := *refctx.Edge.Src
-		kp := &tmp
-		kp.Path = kp.Path[:len(common)]
-		f, err := m.EnsureField(kp, nil)
+		f, err := m.EnsureField(d2ast.MakeKeyPath(common), nil)
```
nhooyr, line 768: lol i forgot i added MakeKeyPath at some point nice.

nhooyr (approved):
✅

PR #825, opened 2023-02-15
tooling: Add .PDF exports

### LINKED ISSUE #120: export: PDF
straightforward with gopdf, will dedicate time to do it when there's enough demand

berniexie:
Uses GoFPDF library to generate pdf exports of d2 diagrams. (I couldn't find a newer library that is able to dynamically generate all the different parts of a page). The pdf will have a header on each export that indicates the exact path to that layer/scenario/step.

closes #120


main.go
```diff
@@ -349,6 +356,87 @@ func _render(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, sketc
...
+		ext := filepath.Ext(outputPath)
+		trimmedPath := strings.TrimSuffix(outputPath, ext)
+		splitPath := strings.Split(trimmedPath, "/")
+		rootName := splitPath[len(splitPath)-1]
+		currBoardPath = append(boardPath, rootName)
+	} else {
+		currBoardPath = append(boardPath, diagram.Name)
+	}
+
+	if !isScenario {
```
berniexie, line 378: In Terrastruct, Scenarios are folders that can only contain steps, so I chose to ignore the generation step for scenarios. 

However, now we can have scenarios that are a diagram? If that's the case what's the point of having scenarios vs layers?

nhooyr: > However, now we can have scenarios that are a diagram? If that's the case what's the point of having scenarios vs layers?

Code reuse. Scenarios are layers where the base of the layer is another diagram. So you have all this code creating a diagram and then a scenario of said diagram where you change just a few things.

berniexie:
[out2.pdf](https://github.com/terrastruct/d2/files/10738473/out2.pdf)

Example PDF

nhooyr:
Beautiful output!

alixander:
we can use the diagram name as the title page

nhooyr:
Is there a way we could make the text highlightable like normal PDF text?

berniexie:
@alixander the diagram name is already the root board though.

berniexie:
@nhooyr the text in the shapes? Unfortuntately not since those are rendered from PNGs

main.go
berniexie: If a scenario exists but only contains steps, is the expectation that it has a "page" within the pdf?

nhooyr: Maybe like a section heading. Though I don't think right now we have a way to detect that which is unfortunate.

Even for general output, if a scenario only contains steps and no modifications of its own, it shouldn't render.

nhooyr:
> @nhooyr the text in the shapes? Unfortuntately not since those are rendered from PNGs

Could we overlay text somehow?

alixander:
> Could we overlay text somehow?

yeah we can. i've done that before for Terrastruct looooooooong ago.

bo-ku-ra:
on windows (installed inkscape : microsoft store - free)

1.open svg-file by inkscape
2.save as foobar.pdf

ex.
[t.pdf](https://github.com/terrastruct/d2/files/10738567/t.pdf)

lib/pdf/pdf.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+type GoFPDF struct {
+	pdf *gofpdf.Fpdf
+}
+
+func Init() *GoFPDF {
+	newGofPDF := gofpdf.NewCustom(&gofpdf.InitType{
+		UnitStr:    "in",
+		FontDirStr: "./d2renderers/d2fonts/ttf",
```
alixander, line 18: this won't work

alixander (changes requested):
Screen Shot 2023-02-22 at 3 12 20 PM

lib/pdf/pdf.go
berniexie: hmmm so far this is the only path that works for me lol.

alixander: ??

i don't understand. this is a string. at runtime this'll look for this path from wherever they're running d2

main.go
```diff
@@ -285,11 +286,22 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, sketc
 		return nil, false, err
 	}
 
-	compileDir := time.Since(start)
-	svg, err := render(ctx, ms, compileDir, plugin, sketch, pad, themeID, darkThemeID, inputPath, outputPath, bundle, forceAppendix, page, ruler, diagram)
+	var svg []byte
+	if filepath.Ext(outputPath) == ".pdf" {
+		svg, err = renderPDF(ctx, ms, plugin, sketch, pad, outputPath, page, ruler, diagram, nil, nil)
+	} else {
+		compileDir := time.Since(start)
```
alixander, line 293: this had to have been a typo lol. `compileDur`

alixander (approved):
siiick
Screen Shot 2023-02-24 at 1 23 56 PM

PR #826, opened 2023-02-15
renderer: update icon size

ejulio-ts:
This is a followup from a PR in TALA where we decreased the icon size as it was overlapping with container borders.

I'll open an issue here as the ideal solution would be to either, precompute the icon size in D2 and send it to the layout engines, or get the icon size in D2 from the layout engines and compute it here only if the layout engine doesn't give back the size.

here is how it looks like now (the other 3 tests look the same)
image


alixander:
i don't want to decrease the icon size. how is that the solution here

PR #833, opened 2023-02-16
d2plugin: Have each plugin return the correct Type

nhooyr:
A plugin from FindPlugin will now return the correct Type and Path when
calling info on it. Before, they would only be set on ListPluginInfos
which makes no sense.


nhooyr:
`d2 layout tala` now correctly shows at the top:

```
tala (executable plugin at ~/.local/share/gopath/bin/d2plugin-tala):
```

alixander (approved):

PR #837, opened 2023-02-17
serde: fiz deserialize

ejulio-ts:
when rendering the following diagram with `TALA`, it errored when adding `no collision.near: game loop.collision detected`

```
direction: down
first input -> start game -> game loop

game loop: {
  direction: down
  input -> increase bird top velocity

  move bird -> move pipes -> render

  render -> no collision -> wait 16 milliseconds -> move bird
  render -> collision detected -> game over
  no collision.near: game loop.collision detected
}
```

The reason was that when deserializing the graph, `game loop.Children` had the full path `game loop.collision detection` instead of just `collision detection`. Then, TALA failed to find the `near` object and resulted in a `panic`.

### DIFF
--- a/ci/release/changelogs/next.md
+++ b/ci/release/changelogs/next.md
@@ -13,3 +13,4 @@
 - Fixes rare compiler bug when using underscores in edges to create objects across containers. [#824](https://github.com/terrastruct/d2/pull/824)
 - Fixes rare possibility of rendered connections being hidden or cut off. [#828](https://github.com/terrastruct/d2/pull/828)
 - Creating nested children within `sql_table` and `class` shapes are now prevented (caused confusion when accidentally done). [#834](https://github.com/terrastruct/d2/pull/834)
+- Fixes graph deserialization. [#837](https://github.com/terrastruct/d2/pull/837)
--- a/ci/release/changelogs/next.md
+++ b/ci/release/changelogs/next.md
@@ -13,4 +13,4 @@
 - Fixes rare compiler bug when using underscores in edges to create objects across containers. [#824](https://github.com/terrastruct/d2/pull/824)
 - Fixes rare possibility of rendered connections being hidden or cut off. [#828](https://github.com/terrastruct/d2/pull/828)
 - Creating nested children within `sql_table` and `class` shapes are now prevented (caused confusion when accidentally done). [#834](https://github.com/terrastruct/d2/pull/834)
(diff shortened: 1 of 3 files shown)


ci/release/changelogs/next.md
```diff
@@ -13,3 +13,4 @@
 - Fixes rare compiler bug when using underscores in edges to create objects across containers. [#824](https://github.com/terrastruct/d2/pull/824)
 - Fixes rare possibility of rendered connections being hidden or cut off. [#828](https://github.com/terrastruct/d2/pull/828)
 - Creating nested children within `sql_table` and `class` shapes are now prevented (caused confusion when accidentally done). [#834](https://github.com/terrastruct/d2/pull/834)
+- Fixes graph deserialization. [#837](https://github.com/terrastruct/d2/pull/837)
```
nhooyr, line 16:
```suggestion
- Fixes graph deserialization bug. [#837](https://github.com/terrastruct/d2/pull/837)
```

d2graph/serde.go
```diff
@@ -49,7 +49,7 @@ func DeserializeGraph(bytes []byte, g *Graph) error {
 			for _, id := range so["ChildrenArray"].([]interface{}) {
 				o := idToObj[id.(string)]
 				childrenArray = append(childrenArray, o)
-				children[strings.ToLower(id.(string))] = o
+				children[strings.ToLower(o.IDVal)] = o
```
nhooyr, line 52:
```suggestion
				children[strings.ToLower(o.ID)] = o
```

nhooyr (changes requested):
Could you move the test alex added to the monorepo here into e2etests?

ejulio-ts:
The e2etest? If so, I can add it, but I don't see a good reason for it. It's not testing anything new and we're not looking for its aesthetics. For example, here it wouldn't catch this bug because it was specific to TALA.

nhooyr:
> The e2etest? If so, I can add it, but I don't see a good reason for it. It's not testing anything new and we're not looking for its aesthetics. For example, here it wouldn't catch this bug because it was specific to TALA.

Well it's not specific to TALA but that's the only part where we currently use the serialization code and the tests don't have an option to use external dagre yet so yea it's fine for now then.

nhooyr (approved):
✅

ejulio-ts:
Maybe, it makes sense to keep it in TALA so we detect any changes there ASAP

PR #840, opened 2023-02-18
tests: compare graphs after serialize/deserialize

ejulio-ts:
followup from https://github.com/terrastruct/d2/pull/837

We were exercising the serialization routine, but weren't testing if the resulting graph was the same

- Added `Compare` to graph, object, edge
- Fixed `root` serialization
- Fixed `near` removal
- No e2e changes


d2graph/d2graph.go
```diff
@@ -1582,3 +1582,316 @@ func (g *Graph) SortEdgesByAST() {
 	})
 	g.Edges = edges
 }
+
+func (g *Graph) Compare(other *Graph) error {
```
alixander, line 1586: I think these should go in `serde`, everywhere else I wouldn't compare this way, I'd just do equality check or compare IDs.

d2layouts/d2near/layout.go
```diff
@@ -98,7 +98,7 @@ func WithoutConstantNears(ctx context.Context, g *d2graph.Graph) (nears []*d2gra
 			nears = append(nears, obj)
 			g.Objects = append(g.Objects[:i], g.Objects[i+1:]...)
 			i--
-			delete(obj.Parent.Children, obj.ID)
+			delete(obj.Parent.Children, strings.ToLower(obj.ID))
```
alixander, line 101: which test was this catching?

alixander (changes requested):

d2layouts/d2near/layout.go
ejulio-ts: `stable/constant_near_stress`

`len(obj.Children) = 5` and `len(other.Children) = 2` because `Joe`, `Donald`, `The ..` were not properly deleted with `d2near.WithoutConstantNears`

```
x -> y
The top of the mountain: { shape: text; near: top-center }
Joe: { shape: person; near: center-left }
Donald: { shape: person; near: center-right }
bottom: |md
	# Cats, no less liquid than their shadows, offer no angles to the wind.

  If we can't fix it, it ain't broke.

  Dieters live life in the fasting lane.
| { near: bottom-center }
i am top left: { shape: text; near: top-left }
i am top right: { shape: text; near: top-right }
i am bottom left: { shape: text; near: bottom-left }
i am bottom right: { shape: text; near: bottom-right }
```

alixander (approved):

Issue #723, opened 2023-01-26
add error for dagre's broken container edges


gavin-ts:
dagre currently cannot render self edges on containers correctly, or container-child edges. As mentioned below we should give an error or warning when attempting to run d2 on a diagram containing these when layout is set to dagre.

> if this problem occurs on d2(dagre), d2 should be abend (svg file is not output) and report this problem to me.
> I will revise this d2 source.

_Originally posted by @bo-ku-ra in https://github.com/terrastruct/d2/issues/707#issuecomment-1404570829_

PR #845, opened 2023-02-19
container dimensions, check layout feature support

### LINKED ISSUE #844: width and height should be usable on containers
just not for dagre

alixander:
1. Removes restriction of adding dimensions on containers
2. Adds a check of compatibility between feature and layout. There's issue reports and confusion when we let features be used when it's not supported by the layout engine. E.g. near key being set to another object doesn't work. They think it's a bug on our side.

closes #844


d2layouts/d2elklayout/layout.go
```diff
@@ -192,13 +198,18 @@ func Layout(ctx context.Context, g *d2graph.Graph, opts *ConfigurableOpts) (err
 				EdgeEdgeBetweenLayersSpacing: 50,
 				HierarchyHandling:            "INCLUDE_CHILDREN",
 				ConsiderModelOrder:           "NODES_AND_EDGES",
+				// Why is it (height, width)?
```
alixander, line 201: i have no clue. but it works.

alixander:
(@nhooyr tagged for second point of PR -- CLI & plugin changes)

d2layouts/d2elklayout/layout.go
```diff
@@ -192,13 +198,21 @@ func Layout(ctx context.Context, g *d2graph.Graph, opts *ConfigurableOpts) (err
 				EdgeEdgeBetweenLayersSpacing: 50,
 				HierarchyHandling:            "INCLUDE_CHILDREN",
 				ConsiderModelOrder:           "NODES_AND_EDGES",
+				// Why is it (height, width)? I have no clue, but it works.
```
nhooyr, line 201: 😂

d2plugin/plugin_elk.go
```diff
@@ -85,8 +85,11 @@ func (p elkPlugin) Info(ctx context.Context) (*PluginInfo, error) {
 		f.AddToOpts(opts)
 	}
 	return &PluginInfo{
-		Name:      "elk",
-		Type:      "bundled",
+		Name: "elk",
+		Type: "bundled",
+		Features: map[PluginFeature]struct{}{
```
nhooyr, line 90: This should be an array not a map, even if that makes checking for a feature a little harder.

nhooyr (changes requested):

alixander:
oh i can add https://github.com/terrastruct/d2/issues/723 to this

d2plugin/plugin_elk.go
alixander: why? the ordering does not matter.

alixander: i guess if they print help and the order differs each time b/c of map iteration not being deterministic, that's bad.

nhooyr (commented):
✅

PR #857, opened 2023-02-20
delete id deltas reserved

alixander:
We considered this for edge keys, but i guess mistakenly omitted for regular keys
https://github.com/terrastruct/d2/blob/95eccae9c1a4a59037b930332d93962235541c7f/d2oracle/edit.go#L1908

### DIFF
--- a/d2oracle/edit.go
+++ b/d2oracle/edit.go
@@ -1864,6 +1864,12 @@ func DeleteIDDeltas(g *d2graph.Graph, key string) (deltas map[string]string, err
 	conflictNewIDs := make(map[*d2graph.Object]string)
 	conflictOldIDs := make(map[*d2graph.Object]string)
 	if mk.Key != nil {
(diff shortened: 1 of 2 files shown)


d2oracle/edit.go
```diff
@@ -1864,6 +1864,12 @@ func DeleteIDDeltas(g *d2graph.Graph, key string) (deltas map[string]string, err
 	conflictNewIDs := make(map[*d2graph.Object]string)
 	conflictOldIDs := make(map[*d2graph.Object]string)
 	if mk.Key != nil {
+		ida := d2graph.Key(mk.Key)
+		// Deleting a reserved field cannot possibly have any deltas
```
nhooyr, line 1868: Because reserved fields have no depth right?

alixander: they can, `style.fill`. it just won't affect object/edge IDs

PR #876, opened 2023-02-23
compiler: Fix underscore reference

alixander:


d2compiler/compile.go
```diff
@@ -189,7 +189,7 @@ func (c *compiler) compileField(obj *d2graph.Object, f *d2ir.Field) {
 			}
 		}
 		scopeObjIDA := d2ir.IDA(fr.Context.ScopeMap)
-		scopeObj, _ := obj.Graph.Root.HasChildIDVal(scopeObjIDA)
+		scopeObj := obj.Graph.Root.EnsureChildIDVal(scopeObjIDA)
```
nhooyr, line 192: How did this fix the problem? Wouldn't the scopeObj be guaranteed to exist?

nhooyr (changes requested):

d2compiler/compile.go
alixander: i msged u in slack:  "i think cuz u compile depth-first, the outer object referenced by an underscore key may not exist yet, so scopeobj is nil when compiling its references"

nhooyr: > i msged u in slack: "i think cuz u compile depth-first, the outer object referenced by an underscore key may not exist yet, so scopeobj is nil when compiling its references"

That doesn't make sense to me. We create all parent objects with `obj = obj.EnsureChild(d2graphIDA([]string{f.Name}))` first on line 174.

nhooyr (approved):
Ok I investigated and turns out this is the correct fix. The real reason though is that we compile references as soon as we see the object for the first time. So it's possible that one of the reference's ScopeObj's hasn't yet been created if underscores have been used and the object is referenced from a more nested scope.

alixander:
@nhooyr i'm not sure why it changed the JSON for the layers example. it uses no underscores. can you take a look and verify?

nhooyr (changes requested):
oh yea that's not supposed to happen. It's because of id's like `scenarios.storm`. I'll have to change the code here to compile references only at the end.

alixander:
kk will let u make that change then.

Issue #831, opened 2023-02-16
Automatic dark mode


nhooyr:
It would be really nice if the SVGs could be embedded into blogs/documentation and automatically adjust their theme to become dark if the user is on dark mode.

nhooyr:
Example of what I'd like to avoid:

<img width="802" alt="image" src="https://user-images.githubusercontent.com/10180857/219399632-0d0921bd-0638-4074-9314-7aa1ddcb9daa.png">

From https://github.com/nhooyr/flappy-bird

alixander:
https://github.com/terrastruct/d2/pull/613 implements this. it replaces the hardcoded theme colors with css classes like `THEME_NEUTRAL_1`, and then uses media queries to determine which css color to load. it's nearly there, but the author's back in school or something so i gotta take it over one of these days

vfosnar:
this can be closed @alixander

nhooyr:
@alixander Just tested latest master and it doesn't seem to work for me. Remains in light even if I change system to dark mode. Not sure if we didn't add a `@media (prefers-color-scheme: dark ) {` somewhere.

vfosnar:
@nhooyr have you tried using these flags
```bash
d2 --theme 0 --dark-theme 200 ...
```

nhooyr:
Works now, beautiful!

<img width="1512" alt="Screenshot 2023-02-24 at 3 43 58 AM" src="https://user-images.githubusercontent.com/10180857/221171155-1b3b8e82-4bff-4690-a518-f88d9b1c07a0.png">

Is there any reason that `--dark-theme` isn't enabled by default? If we're using meda queries to determine which to render it seems reasonable to have it on by default @alixander

nhooyr:
> I think the only thing is that it is not an expected behavior for many users.

It's only unexpected because we didn't do it from the start. But now it seems like a strictly better behaviour.

And we should get rid of that warning unless they explicitly pass `--dark-theme` in which case pngs should be dark too.

alixander:
> But now it seems like a strictly better behaviour.

nah i don't think so. some diagrams make style changes based on the theme staying the way it is. e.g. a color that pairs with the color of their theme. once it switches it dark, it'll look ugly. we can invert the color, as suggested here, https://github.com/terrastruct/d2/issues/885. but i'm not sure how that'll end up looking without testing.

the user should explicitly pass so they don't get surprised their diagram doesn't look a certain way for dark mode users.

[...]

PR #894, opened 2023-02-25
main: Enable dark mode by default

nhooyr:
Updates #831


alixander (changes requested):
we can't just do this without anything else. any set styles will be up for incompatability. we have to do that !important thing discussed

nhooyr:
Ok I'm going to modify the release notes to indicate the release dark mode is beta.

nhooyr:
Reverted as we discussed on slack.

Issue #869, opened 2023-02-22
3d on hexagon shape


alixander:
<img width="455" alt="220769755-8f1bda9d-ff20-459f-bb5a-7fae11a90aae" src="https://user-images.githubusercontent.com/3120367/220771571-96efebb7-dea0-45bc-909a-48dd88ffbce5.png">

related: https://github.com/terrastruct/d2/issues/868

JettChenT:
Hi! I'm willing to work on this issue.
Is there an API for drawing dotted lines in the SVG library to create the 3D perspective effect, or should I write a loop to achieve this?
Thanks!

JettChenT:
Here's my current progress:
<img width="1442" alt="CleanShot 2023-02-25 at 00 02 15@2x" src="https://user-images.githubusercontent.com/45888395/221226822-61e5d8c2-c62f-498f-8e43-e21346e884e1.png">

alixander:
awesome! no documentation unfortunately.

> API for drawing dotted lines in the SVG library to create the 3D perspective effect.

ohh very interesting. there isn't. i'm guessing you'll want to split it up into multiple paths or lines, and the ones behind will get lower opacity and dashed properties. 

i wonder if this exact image will be the best rendering of 3d for hexagon. it doesn't look like a lot of space for a label/icon. i'll let our designer @katwangy take a look this morning and chime in

alixander:
<img width="617" alt="Screen Shot 2023-02-24 at 10 41 14 AM" src="https://user-images.githubusercontent.com/3120367/221264057-8728ae73-f9b4-45a7-9ff0-b809bbb4b4b7.png">

I would agree with that, the different perspective could look weird. can you change to doing this please? Sorry for that, I know you already started @JettChenT

JettChenT:
Thanks a lot! I'll change this to the new design right away.

PR #895, opened 2023-02-25
Render: Support 3d hexagons

JettChenT:
#869 

This adds support for 3d hexagons. For example, the following code:
```
hexagon: {shape: "hexagon"}
hexagon.style.3d: true
```
will produce:
CleanShot 2023-02-25 at 11 33 15@2x

This example can be ran via `./ci/e2ereport.sh -test-case=hexagon_3d`


alixander:
@katwangy can you check if the dimensions of the 3d stuff look right to u?

alixander:
@JettChenT could you sign your commits please? https://github.com/terrastruct/d2/blob/master/docs/CONTRIBUTING.md#logistics

alixander (changes requested):
ty @JettChenT ! code looks good to me, and quite reusable for future 3d implementations.

i pulled and ran `go install .`, and played around with it.

the default themed color looks a bit different (this one in sketched mode, but i don't think it made a difference)

Screen Shot 2023-02-25 at 8 04 01 AM

```
hexagon: {
  shape: hexagon
  style.3d: true
}


rect: {
  shape: rectangle
  style.3d: true
}

hexagon -> rect
```

and then when i add a fill, it seems to be ignored (can you add an e2e test that has all 3d shapes fill and stroke set?):

Screen Shot 2023-02-25 at 8 05 03 AM

```
hexagon: {
  shape: hexagon
  style.3d: true
  style.fill: honeydew
}


rect: {
  shape: rectangle
  style.3d: true
  style.fill: honeydew
}

hexagon -> rect
```

JettChenT:
Hi! I've set up commit signing, and fill should now work as expected(e2e test accesible via `3d_fill_and_stroke`):
sketch exp

However, the default 3d appearance of 3d Hexagon is still different in color from the 3d rectangle.
sketch exp
CleanShot 2023-02-26 at 10 49 45@2x

This is most likely due to the default fill color for them being different:
CleanShot 2023-02-26 at 10 52 36@2x

CleanShot 2023-02-26 at 10 51 41@2x

d2renderers/d2svg/d2svg.go
```diff
@@ -746,6 +746,128 @@ func render3dRect(targetShape d2target.Shape) string {
...
+	}
+	halfYFactor := 43.6 / 87.3
+
+	// draw border all in one path to prevent overlapping sections
+	var borderSegments []string
+	borderSegments = append(borderSegments,
+		moveTo(d2target.Point{X: scale(targetShape.Width, 0.25), Y: 0}),
+	)
+	Y_OFFSET := d2target.THREE_DEE_OFFSET / 2
+	for _, v := range []d2target.Point{
```
alixander, line 767: maybe leave a comment on which part of the 3d this is drawing. e.g. the top side vs the right side

d2renderers/d2svg/d2svg.go
```diff
@@ -746,6 +746,128 @@ func render3dRect(targetShape d2target.Shape) string {
...
+		{X: scale(targetShape.Width, 0.75), Y: targetShape.Height},
+		{X: targetShape.Width, Y: scale(targetShape.Height, halfYFactor)},
+		{X: scale(targetShape.Width, 0.75), Y: 0},
+		{X: scale(targetShape.Width, 0.25), Y: 0},
+	} {
+		sidePoints = append(sidePoints,
+			fmt.Sprintf("%d,%d", v.X+targetShape.Pos.X, v.Y+targetShape.Pos.Y),
+		)
+	}
+	// TODO make darker color part of the theme? or just keep this bypass
```
alixander, line 856: bit of an ongoing debate hah, i think this is fine for now. https://github.com/terrastruct/d2/issues/896

d2renderers/d2svg/d2svg.go
```diff
@@ -958,6 +1080,39 @@ func drawShape(writer io.Writer, targetShape d2target.Shape, sketchRunner *d2ske
 			}
 		}
 	case d2target.ShapeText, d2target.ShapeCode:
+	case d2target.ShapeHexagon:
```
alixander, line 1083: can u put this above the line above? i think slightly nicer to have the "do nothing" cases next to default and not sandwiched between the "do something" cases

alixander (changes requested):
excellent! looks great.

Screen Shot 2023-02-25 at 8 06 42 PM

i did notice that the more square hexagon looks better than the wide ones. separate issue from this though, so made another to track: https://github.com/terrastruct/d2/issues/904

Could you squash your commits or re-commit to sign those previous ones please? then i think it's good to go! thank you!

katwangy:
Possible to make it thicker so the depth is more on par with the cube? The top part of the hexagon is looking a little thin

d2renderers/d2svg/d2svg.go
JettChenT: got it. I've added the comments

d2renderers/d2svg/d2svg.go
JettChenT: Alright! I've changed the order

d2renderers/d2svg/d2svg.go
JettChenT: got it.

JettChenT:
Hi, I've squashed the commits and made the corresponding changes. Not quite sure why the CI fails on `COLOR=1 ./make.sh` though, as it works on my local machine.

@katwangy would the following be better?(adjusting Y offset from 1/2 of 3D offset to 2/3 of 3D offset)
CleanShot 2023-02-27 at 09 58 48@2x

alixander (approved):
perfect, thanks @JettChenT !!

PR #913, opened 2023-02-27
Elk border radius

donglixiaoche:
#688 

add support for customized border-radius of elk connection, the following script:
```
a -> b
a -> c {
  style: {
    border-radius: 0
  }
}
a -> e {
  style: {
    border-radius: 5
  }
}
a -> f {
  style: {
    border-radius: 10
  }
}
a -> g {
  style: {
    border-radius: 20
  }
}
```
will produce:
image

you can get this report by run:
`./ci/e2ereport.sh -test-case=elk_border_radius`


donglixiaoche:
@alixander rasied a pull request, pls help to take a look, and i will complete the related docs once code review is done

alixander (changes requested):
@donglixiaoche great! please add some e2e tests and screenshots. see https://github.com/terrastruct/d2/pull/895 for an ongoing pull request example

alixander:
also please sign your commits https://github.com/terrastruct/d2/blob/master/docs/CONTRIBUTING.md#logistics

donglixiaoche:
@alixander i got one question to discuss with you

well, previously we got a fixed `10` border-radius for every elk connection, i think we should keep this as defalut behavior right? if so, then i think there are two choices:
1.  inside the `pathData` function, when draw an elk connection, if the border-radius equals 0, then set to 10, but this will prevent user from set border-radius to 0 too
image
so i add a default `-1` if no `border-radius` is set, but this result in a lot of failed of our test cases

2. add a default `BorderRadius: 10` for `BaseConnection`, but this will cause the same issue above
image

do you have any good idea?

alixander:
@donglixiaoche ohh i see.

> add a default BorderRadius: 10 for BaseConnection

i think we should do this. that's the most accurate anyway. all the tests will need to be regenerated (`TESTDATA_ACCEPT=1 ./ci/test.sh`), but that's okay.

donglixiaoche:
@alixander already signed my commits, added an e2e testcase and regenerated testcases

d2target/d2target.go
```diff
@@ -293,11 +293,12 @@ type Connection struct {
...
-	StrokeDash  float64 `json:"strokeDash"`
-	StrokeWidth int     `json:"strokeWidth"`
-	Stroke      string  `json:"stroke"`
-	Fill        string  `json:"fill,omitempty"`
+	Opacity      float64 `json:"opacity"`
+	StrokeDash   float64 `json:"strokeDash"`
+	StrokeWidth  int     `json:"strokeWidth"`
+	Stroke       string  `json:"stroke"`
+	Fill         string  `json:"fill,omitempty"`
+	BorderRadius float64 `json:"borderRadius,omitempty"`
```
alixander, line 301: this should be an int. people won't need this much precision for border radius.

alixander (changes requested):

e2etests/stable_test.go
```diff
@@ -12,6 +12,32 @@ var testMarkdown string
 
 func testStable(t *testing.T) {
 	tcs := []testCase{
+		{
+			name: "elk_border_radius",
+			script: `
+				a -> b
+				a -> c {
```
alixander, line 19: this script is malformed. should be colons before maps

e.g.

```
a -> c: {
  ...
}
```

i guess it still works because we format it beforehand or something, but still the tests should use correct syntax

alixander (changes requested):

d2target/d2target.go
donglixiaoche: but finally before render, we need to convert it to floast64 anyway
image

is it necessary to define BorderRadius as an int?

donglixiaoche: ahh, or maybe we can convert `dist/2` to an int

alixander: no the distance should still be precise i think. in fact that code should probably be `dist/2.0`.

> is it necessary to define BorderRadius as an int?

hmmmm actually this just made me think of something. i was thinking users wouldn't need it, but maybe we can designate 0-1 values as percentages. then `0.5` means 50% border-radius! 

this was one of the things that D2 couldn't do in recreating this diagram: https://github.com/terrastruct/d2/issues/920

want to include that change in here? @donglixiaoche . if not, no worries, can defer to another PR

donglixiaoche: sure, let me think about it

alixander: i guess it doesn't make sense for connections because they don't have dimensions (https://stackoverflow.com/a/29966500/20901387). would only be applicable for shapes. anyway, still worth keeping as float64

e2etests/stable_test.go
donglixiaoche: fixed

d2target/d2target.go
donglixiaoche: okok

alixander: will track here: https://github.com/terrastruct/d2/issues/947

alixander (approved):
tested locally. works well. 💯

alixander:
oops forgot to ask you to update changelog. no big deal, i just made a commit to do it: https://github.com/terrastruct/d2/commit/91e0bbdf54784572b15d60cc0c1bc5527f1f607d . thanks @donglixiaoche !

PR #923, opened 2023-02-27
cli: Improve multiboard output

nhooyr:
- Boards with subboards are now selfcontained folders with index.svg/png as
  the root board render.
- Boards that are only containers of other boards are not rendered. For example a
  scenario with no modifications and only steps only has its steps
  rendered.
- Boards with sibling boards of another type are rendered under a
  subdirectory indicating their type to separate them. For example a
  board with layers and scenarios has its layers rendered into subfolder
  layers and scenarios into subfolder scenarios.

cc @berniexie see BoardContainer field on d2target.Board for the field
you were looking for to skip renders for PDFs too.


.gitignore
```diff
@@ -6,3 +6,4 @@
 e2e_report.html
 bin
 out
+d2
```
alixander, line 9: why ignore these? we want to use it for examples/docs, tests from files

d2renderers/d2sketch/testdata/animated_dark/sketch.exp.svg
```diff
@@ -30,7 +30,7 @@
...
..., 148.547838 142.592714" fill="none" class="connection animated-connection stroke-B1" style="stroke-width:2;stroke-dasharray:10.000000,9.865639;stroke-dashoffset:-198.656393;animation: dashdraw 4.932820s linear infinite;" mask="url(#1960707820)" /><path d="M-8.527627 -3.097061 L1.749550 0.558791 L-8.562935 4.521533" stroke="none" class="connection fill-B1" style="stroke-width:0;" transform="translate(148.877606 140.778378) rotate(-136.00508600525419)" /> <path d="M-10.153731 -4.038897 C-7.293657 -2.964754, -5.552453 -3.126871, 0.222305 -0.654474 M-10.160117 -4.253535 C-7.616436 -2.677663, -5.569656 -2.320404, -0.086565 0.272291 M0.578048 -0.807164 C-2.240460 1.133634, -3.845699 1.135504, -9.579367 4.140709 M-0.217907 -0.322328 C-3.660571 0.941126, -7.003142 2.167050, -10.100296 3.840861 M-9.957758 4.629247 C-9.937438 2.794817, -10.508655 0.509238, -9.330834 -3.522818 M-10.354741 4.285014 C-9.712366 0.996453, -9.805329 -1.235319, -9.648840 -4.366524" fill="none" class="connection stroke-B1" style="stroke-width:2;" transform="translate(148.877606 140.778378) rotate(-136.00508600525419)" /></g><mask id="1960707820" maskUnits="userSpaceOnUse" x="-101" y="-100" width="429" height="799">
```
alixander, line 33: revert these plz

alixander (changes requested):

.gitignore
nhooyr: that's the d2 binary being ignored if you run `go build`, not `*.d2`.

alixander: oh lol my b

d2compiler/compile.go
```diff
@@ -73,7 +73,7 @@ func (c *compiler) compileBoard(g *d2graph.Graph, ir *d2ir.Map) *d2graph.Graph {
 	c.compileBoardsField(g, ir, "steps")
 	if d2ir.ParentMap(ir).CopyBase(nil).Equal(ir.CopyBase(nil)) {
 		if len(g.Layers) > 0 || len(g.Scenarios) > 0 || len(g.Steps) > 0 {
-			g.BoardContainer = true
+			g.IsContainerOnly = true
```
alixander, line 76: I actually don't like the use of Container here as it has meaning for diagrams. how about just `IsFolder`?

nhooyr: Will go with `IsFolderOnly`.

alixander (approved):

Issue #903, opened 2023-02-25
cli testing framework


alixander:
regressions like https://github.com/terrastruct/d2/issues/889 are bad and will continue to happen unless we have tests for the CLI.

PR #927, opened 2023-02-27
d2cli: Move main code into d2cli so it can be imported from tests

nhooyr:
Updates #903


nhooyr:
@alixander I want to merge this in ASAP to avoid conflicts with any other changes to main.go

alixander (approved):

PR #929, opened 2023-02-28
pdf: Support diagram fill in pdf exports

berniexie:
We have to make some tweaks to fill the background of pdf exports to match the diagram fill color.

Tested using
```
this is orange

style.fill: "#FF4500"

layers: {
  cyan: {
    style.fill: cyan
    this is cyan
  }
  black: {
    style.fill: "#000000"
    this is black
  }
}
steps: {
  purple: {
    style.fill: purple
    this is purple
  }
}
```

which generated
[orange.pdf](https://github.com/terrastruct/d2/files/10845402/orange.pdf)


lib/color/color.go
```diff
@@ -143,3 +145,204 @@ const (
...
+	rgb = RGB{
+		Red:   uint8(values >> 16),
+		Green: uint8((values >> 8) & 0xFF),
+		Blue:  uint8(values & 0xFF),
+	}
+
+	return rgb, nil
+}
+
+// https://www.w3schools.com/cssref/css_colors.php
```
alixander, line 189: w3schools is ancient. use a more recent source. e.g. https://147colors.com/rebeccapurple/

alixander (changes requested):

lib/color/color.go
nhooyr: https://developer.mozilla.org/en-US/docs/Web/CSS/named-color is probably best

berniexie: Changed it to w3 official source

nhooyr: nice that's even better.

lib/color/color.go
```diff
@@ -143,3 +146,205 @@ const (
...
+
+	hsp := math.Sqrt(0.299*math.Pow(r, 2) + 0.587*math.Pow(g, 2) + 0.114*math.Pow(b, 2))
+
+	return hsp > 130
+}
+
+// https://gist.github.com/CraigChilds94/6514edbc6a2db5e434a245487c525c75
+func Hex2RGB(hex string) (RGB, error) {
+	var rgb RGB
+	if hex[0:1] == "#" {
```
alixander, line 170: `if len(hex) > 3 && hex[0] == "#"`

alixander (approved):

PR #934, opened 2023-02-28
pdf: Fix pdf fill when using theme colors

berniexie:
Needed to be able to parse theme colors like "N7" when filling the background of PDF exports.


lib/pdf/pdf.go
```diff
@@ -47,10 +49,15 @@ func (g *GoFPDF) GetFillRGB(fill string) (color.RGB, error) {
 		return rgb, nil
 	}
 
+	if color.IsThemeColor(fill) {
```
alixander, line 52: put the above clause in an else

alixander (approved):

PR #940, opened 2023-02-28
exports: pdf exports link overlay

berniexie:
I had to convert the pdf export units to "pt" to support X and Y.

Here's an example:
[links.pdf](https://github.com/terrastruct/d2/files/10855196/links.pdf)


lib/pdf/pdf.go
```diff
@@ -117,17 +119,38 @@ func (g *GoFPDF) AddPDFPage(png []byte, boardPath []string, themeID int64, fill
...
+	for _, shape := range shapes {
+		if shape.Link != "" {
+			viewboxX, err := strconv.ParseFloat(viewboxSlice[0], 64)
+			if err != nil {
+				return err
+			}
+			viewboxY, err := strconv.ParseFloat(viewboxSlice[1], 64)
+			if err != nil {
+				return err
+			}
```
gavin-ts, line 136: these are the same for all shapes so this can be outside the loop, or passed in as params instead of the slice

alixander (approved):
tested 🚀 

[cat.pdf](https://github.com/terrastruct/d2/files/10855554/cat.pdf)

PR #951, opened 2023-03-01
build: `make` bails on the wrong go version

wycats:
This code:

1. Extracts the required go minor from `go.mod`
2. Gets the current go version from `go version` and normalizes it
3. Checks whether the required go minor is a version prefix for the current go
version, and bails if it isn't.

The conditional is more complicated than it would be if this script supported bash extensions to the POSIX shell standard, but changing to `#!/bin/bash` seemed out of scope for this change.

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -1,6 +1,5 @@
 #!/bin/sh
 set -eu
-
 if [ ! -e "$(dirname "$0")/ci/sub/.git" ]; then
   set -x
   git submodule update --init
--- a/make.sh
+++ b/make.sh
@@ -1,19 +1,6 @@
 #!/bin/sh
 set -eu
 
-REQUIRED_GO_MINOR=$(sed -En 's/^go[[:space:]]+([[:digit:].]+)$/\1/p' go.mod)
-ACTUAL_GO_VERSION=$(go version | sed -n 's/^go version go\([0-9]*\.[0-9]*\.[0-9]*\)\(.*\)/\1/p')
-
-# We use 'case' instead of '[' to match values against complex patterns, because POSIX
-# does not guarantee that '[' supports advanced features like globs and regex.
-if case "$ACTUAL_GO_VERSION" in "$REQUIRED_GO_MINOR".*) false ;; *) true ;; esac then
-  red="\e[0;91m"
-  reset="\e[0m"
-
-  printf "${red}PROBLEM: You need go %s to build d2, but you have %s installed.${reset}\n" "$REQUIRED_GO_MINOR" "$ACTUAL_GO_VERSION"
-  exit 1
-fi
-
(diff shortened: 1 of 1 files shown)


nhooyr (changes requested):
I appreciate the shellfu but we don't need all that. Just error if `if ! go version | grep -q 1.18`.

And do it below `cd -- "$(dirname "$0")"`.

And instead of using printf, use `printferr` which is sourced in this script. It works just like `printf`.

wycats:
> I appreciate the shellfu but we don't need all that. Just error if `if ! go version | grep -qF 1.18`.

It seems useful to leave in the code that gets the version out of `go.mod`, right? Otherwise we end up with two sources of truth that have to be kept in sync.

wycats:
>

nhooyr:
> It seems useful to leave in the code that gets the version out of go.mod, right? Otherwise we end up with two sources of truth that have to be kept in sync.

They don't need to be kept in sync because they don't mean the same thing. The `go.mod` version is the earliest version that our code compiles on. This check is for CI where newer go versions screw up our code formatting.

nhooyr (changes requested):

nhooyr:
It's also not something we're going to bump. As soon as we can build on Go 1.19, the restriction will be removed.

make.sh
```diff
@@ -9,4 +9,11 @@ fi
 PATH="$(cd -- "$(dirname "$0")" && pwd)/ci/sub/bin:$PATH"
 cd -- "$(dirname "$0")"
 
+GO_VERSION=$(sed -En 's/^go[[:space:]]+([[:digit:].]+)$/\1/p' go.mod)
```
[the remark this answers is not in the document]
wycats: okey dokey!

make.sh
```diff
@@ -9,4 +9,9 @@ fi
 PATH="$(cd -- "$(dirname "$0")" && pwd)/ci/sub/bin:$PATH"
 cd -- "$(dirname "$0")"
 
+if ! $(go version | grep -qF "1.18."); then
```
nhooyr, line 12:
```suggestion
if ! go version | grep -qF '1.18'; then
```

make.sh
```diff
@@ -9,4 +9,9 @@ fi
...
+  printferr "You need go 1.18 to build d2.\n"
```
nhooyr, line 13:
```suggestion
  echoerr "You need go 1.18 to build d2."
```
Just realized you don't need printf anymore.

nhooyr (changes requested):

nhooyr (approved):
nice ✅

alixander:
thank you @wycats ! (tbh i originally thought matching with go.mod was right too, appreciate the forethought anyway)

Issue #972, opened 2023-03-04
theme-css needs to append unique id


alixander:
otherwise two diagrams on the same HTML page could overwrite each others css

just like we do for masking. shoudl use exact same method

PR #982, opened 2023-03-06
Class and sqltable border radius

### LINKED ISSUE #689: render: border-radius should work with class and sql-table
currently it's implemented for rectangles, but class and sql-table are rectangular in shape, so should apply to those too

donglixiaoche:
closes #689 
this pr add support for apply border-radius to class and sql_table

the image below can be generate by run `./ci/e2ereport.sh -test-case=class_and_sqlTable_border_radius`:
image


alixander:
hey @donglixiaoche i responded in Discord but copying here: 
```
i think this is an issue with something in your PATH, like you have gem in your PATH pointing to a wrong directory
```

can you double check your PATH?

e2etests/stable_test.go
```diff
@@ -12,6 +12,35 @@ var testMarkdown string
...
+				b: {
+					shape: class
+
+					field: "[]string"
+					method(a uint64): (x, y int)
+					
+					style: {
+						border-radius: 10
+					}
+				}
```
alixander, line 41: can you also add 
```
c: { shape: class }
d: { shape: sql_table }
```

people often use these without bodies, e.g. https://github.com/Greenroom-Robotics/ros-d2#example-usage

I thinkk the behavior should be the same as you have it now, which is the top corners are rounded but the bottom are not. but i'm curious how it looks

donglixiaoche: @alixander fixed the logic for class and table without bodies with testcases. still trying to fix the env issue i got, do you know why in our ci process we need to our `/usr/bin/gem` ?

alixander: The CI process doesn't. This is the leading theory: https://github.com/terrastruct/d2/issues/991#issuecomment-1458415718

alixander (changes requested):
these look great. sorry can you add this test to d2sketch_test as well? I just want to make sure it works there too, then we can merge

e2etests/stable_test.go
alixander: okay found the issue. fixed here: https://github.com/terrastruct/d2/pull/1002 . once you rebase with master, error should go away.

d2renderers/d2svg/table.go
```diff
@@ -91,6 +121,8 @@ func drawTable(writer io.Writer, targetShape d2target.Shape) {
 	rectEl.Fill, rectEl.Stroke = d2themes.ShapeTheme(targetShape)
 	rectEl.ClassName = "shape"
 	rectEl.Style = targetShape.CSSStyle()
+	rectEl.Rx = float64(targetShape.BorderRadius)
```
donglixiaoche, line 124: @alixander after add `Rx` and `Ry` attributes for sql_table and class rect, lots of testcases fails, i think regenerate these testcase should be the correct way right? or do i need to add these attributes when `Shape.BorderRadius` not equal to zero?

donglixiaoche:
@alixander already regenerated some tests. seems like these changes have no effect on d2Sketch, do i need to fix it?

alixander:
> @alixander already regenerated some tests. seems like these changes have no effect on d2Sketch, do i need to fix it?

no it's not a big deal, as long as not broken

donglixiaoche:
@alixander one thing need to discuss with is, althought these changes work well, but the logic for rendering border-radius is distributed between different files. isn't it hard to maintain?

alixander:
> @alixander one thing need to discuss with you is, althought these changes work well, but the logic for rendering border-radius is distributed between different files. isn't it hard to maintain?

yeah it's unfortunate. i didn't expect class and table to deviate so much from d2svg's main rendering. eventually we should consolidate in a refactor

donglixiaoche:
@alixander still some little changes even i just add `rx` and `ry` for rect with non-zero border-radius, is that normal?
image

alixander:
> but i regenerate those testcases anyway, do i need revert it?

yes please revert it. sorry about that. 

as well as the CLI tests: `e2etests-cli/testdata/TestCLI_E2E/hello_world_png_sketch.exp.png`

donglixiaoche:
> > but i regenerate those testcases anyway, do i need revert it?
> 
> @donglixiaoche yes please revert it. sorry about that.
> 
> as well as the CLI tests: `e2etests-cli/testdata/TestCLI_E2E/hello_world_png_sketch.exp.png`

@alixander done, `./make.sh` still passed, sry, my bad, i thought it's part of our ci process

d2renderers/d2svg/table.go
```diff
@@ -12,13 +13,42 @@ import (
 	"oss.terrastruct.com/util-go/go2"
 )
 
+func tableHeaderBorderRadius(shape d2target.Shape) string {
```
alixander, line 16: can you leave a comment that this is also for `class` shapes, despite the name

d2renderers/d2svg/table.go
```diff
@@ -12,13 +13,42 @@ import (
...
+	box := geo.NewBox(
+		geo.NewPoint(float64(shape.Pos.X), float64(shape.Pos.Y)),
+		float64(shape.Width),
+		float64(shape.Height),
+	)
+	topX, topY := box.TopLeft.X+box.Width, box.TopLeft.Y
+	out := fmt.Sprintf(`<clipPath id="%v">`, shape.ID)
```
alixander, line 23: so, there's this kind of annoying thing:

when you have multiple SVGs in the same page, they can reference each other's IDs.

So mask IDs, for example we have to use the diagram hash (https://github.com/donglixiaoche/d2/blob/833967a296c5abb8b2b0d69ff898ca2fe520bbdf/d2renderers/d2svg/d2svg.go#L1650).

That way, when you embed two D2 diagrams in the same page, they don't conflict.

i think you have to do this for clip paths too. So on top of the shape.ID, you should append the diagram hash too.

This is actually a problem in our theme CSS right now (if you embed two diagrams rendered with different themes, the second one will overwrite the first ones): https://github.com/terrastruct/d2/issues/972

d2renderers/d2svg/table.go
```diff
@@ -12,13 +13,42 @@ import (
...
+
 func tableHeader(shape d2target.Shape, box *geo.Box, text string, textWidth, textHeight, fontSize float64) string {
 	rectEl := d2themes.NewThemableElement("rect")
 	rectEl.X, rectEl.Y = box.TopLeft.X, box.TopLeft.Y
 	rectEl.Width, rectEl.Height = box.Width, box.Height
 	rectEl.Fill = shape.Fill
 	rectEl.ClassName = "class_header"
 	str := rectEl.Render()
+	if shape.BorderRadius != 0 {
+		str = strings.Replace(str, "/>", fmt.Sprintf(`clip-path="url(#%v)" />`, shape.ID), 1)
```
alixander, line 50: same comment as the one in `class`

alixander (changes requested):
tested locally and works well! thanks for taking such a tedious one. classes and sql_tables are some of the most used shapes people use D2 for, so i'm sure we'll start seeing some usages of these in no time

d2renderers/d2svg/table.go
donglixiaoche: fixed, clipPath id now is made up with `labelMaskID-shapeID` form

d2renderers/d2svg/table.go
donglixiaoche: fixed as the one in `class`

d2renderers/d2svg/table.go
donglixiaoche: fixed, renamed this func, and add a comment above

d2renderers/d2svg/table.go
alixander: so, masks used here is just invisible areas. for connection labels, we put a mask so that connections don't overlap with the text.

but imagine in this doc, where there's two SVGs, if we called both masks ID "123". In the first SVG, mask 123 is defined as one rectangle, and in the second, it's defined as another rectangle. Well then because they're on the same DOM, the second mask overrides the first mask, and both connections from unrelated diagrams would be using the same mask! 

Screen Shot 2023-03-09 at 2 43 45 PM

alixander (changes requested):

d2renderers/d2svg/class.go
```diff
@@ -18,6 +19,9 @@ func classHeader(shape d2target.Shape, box *geo.Box, text string, textWidth, tex
 	rectEl.Fill = shape.Fill
 	rectEl.ClassName = "class_header"
 	str := rectEl.Render()
+	if shape.BorderRadius != 0 {
+		str = strings.Replace(str, "/>", fmt.Sprintf(`clip-path="url(#%v)" />`, shape.ID), 1)
```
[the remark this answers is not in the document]
alixander: it doesn't matter that it's used by lots of files, it won't affect them. 

Just add a `ClipPath` field to element, and during render, if it's not empty, use it.

d2renderers/d2svg/class.go
```diff
@@ -11,13 +11,13 @@ import (
 	"oss.terrastruct.com/d2/lib/svg"
 )
 
-func classHeader(shape d2target.Shape, box *geo.Box, text string, textWidth, textHeight, fontSize float64) string {
+func classHeader(labelMaskID string, shape d2target.Shape, box *geo.Box, text string, textWidth, textHeight, fontSize float64) string {
```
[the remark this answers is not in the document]
donglixiaoche: fixed

d2renderers/d2svg/class.go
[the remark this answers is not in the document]
donglixiaoche: fixed

d2renderers/d2svg/table.go
donglixiaoche: copy, thx

alixander (approved):
beautiful (code, commit msgs, and resulting diagrams)!

PR #997, opened 2023-03-07
Fix cli race

alixander:
didn't deeply inspect, just fixing the reported errors with what seems like a harmless straightfwd way


nhooyr (commented):
looks fine except for the internal pdf being committed

nhooyr (approved):

PR #1006, opened 2023-03-09
Accept higher value in border-radius to be able to render a pill shaped rectangle

### LINKED ISSUE #947: use 0-1 decimal value as percentages for border-radius
border-radius people may want either percentage (namely, 50% for the "pill" shape), or a px value. 

https://github.com/terrastruct/d2/pull/913#discussion_r1121144213

### LINKED ISSUE #948: increase range for border-radius
now that we can use it for connections, values above 20 are reasonable. let's increase to like 40 or whatever the ceiling still looks reasonable

Poivey:
# Summary
Remove the higher limit of `border-radius`, to be able to render rectangle shape as "pill".

# Details
* Computing the value of `rx` and `ry` from the shape `width` and `height` to render the same way CSS does
* Limiting `rx` and `ry` values to half of the minumum value between `width` and `height` to mimic CSS (see 1st and 2nd example)

# Examples

## CSS example for reference

https://user-images.githubusercontent.com/15009534/231796752-7cfe75ac-b883-4f96-b14f-296fe27bd376.mov

^ Note that `border-radius: 50px` is the same render as `border-radius: 999px`

## Rendered SVG example

https://user-images.githubusercontent.com/15009534/231796759-b80a8e5d-fc27-48b2-8d75-26a267c9b7b2.mov

^ Note that `rx="50"` is NOT the same render as `rx="999"` -> This is why we need the method `calculateAxisRadius` to limit the value `rx` and  `ry` can have.

```
ThisIsElement.style.border-radius: 999
ThisIsElement.width: 200
ThisIsElement.height: 100
```

## Also works vertically

Screenshot 2023-04-13 at 16 13 15

^ When only `rx` is set, it the same as if `ry` had the same value.

## Does not need `width` and `height` to be set manually to work

Screenshot 2023-04-13 at 16 55 20

```
ThisIsElement.style.border-radius: 999
```

### TODO
- [x] Remove "percentage border-radius" tests
- [x] Add new tests

<details>
<summary>[Deprecated] Old description percentage border-radius solution</summary>

# Summary
Considers `border-radius` value between (0,1) as a percentage, applied to [`rx`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/rx) and [`ry`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/ry) shape attributes.

# Details
* Computing the value of `rx` and `ry` from the shape `width` and `height` to rendre the same way CSS does
* Add tests to check `border-radius` value validation
* Add a test with a percentage border-radius
* Some test were updated because `ry` was added to shape having a `rx`. The rendesr are the same. `rx` and `ry` will have the same value if a "number" value is given to `border-radius`, and might have a different one if a "percent" value is given.

# Examples

> Multiple shape with border-radius 10%

> Html/css with border-radius 10%, it is the same shape 

> A shape with border-radius 35% and the same shape using html/css with border-radius 35%

> Border-radius 50%

</details>

<details>
<summary>[Deprecated] Old description with first questions </summary>

Here are example of how it renders, shaped are named according the the applied `border-radius` value (eg. `Example0,05` has `border-radius: 0.05`, `rx` will be `"5%"`)
Screenshot
Screenshot
Screenshot
Screenshot
Screenshot
## Question
I feel like the first 4 examples are rendering as expected, but the last one highlight an issue present when the image is too wide -> small percentages are rendering a more rounded border radius as expected. The difference is easily visible on both `Element0,05` and `ElementWithALongerName0,05` in exemple 2 and 5.
So I think the percentage value might depend on the image width. I'm not used to this so I plan to look for a solution next time I can focus on this issue (likely next week).
Do you have an idea about how I could fix this issue ? Maybe compute an integer `rx` value from the percentage value and the shape width instead of setting with the percentage value directly ?
</details>

~FIX #947~ -> Canceled as render renders were not satisfying
FIX #948


d2graph/d2graph.go
```diff
@@ -216,10 +216,13 @@ func (s *Style) Apply(key, value string) error {
...
 			break
 		}
-		f, err := strconv.Atoi(value)
+		f, err := strconv.ParseFloat(value, 64)
 		if err != nil || (f < 0 || f > 20) {
 			return errors.New(`expected "border-radius" to be a number between 0 and 20`)
 		}
+		if hasDecimalValue := math.Mod(f, 1) != 0; f > 1 && hasDecimalValue {
+			return errors.New(`expected "border-radius" to be an integer if superior to 1`)
+		}
```
alixander, line 225: check if `f > 1` first, then check if decimal

alixander: copy should be: `...if greater than 1`

alixander (changes requested):
@Poivey it should follow CSS conventions as much as possible: https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius

you might need to introduce `ry` to rects too.

so that `0.25` would be 25% of x axis for rx and 25% of y axis for ry. at least i think  that's how CSS is doing it

d2themes/element.go
```diff
@@ -227,3 +227,13 @@ func (el *ThemableElement) Render() string {
 	}
 	return out
 }
+
+func calculateAxisRadius(borderRadius float64, sideLength float64) float64 {
```
Poivey, line 231: I've found calculating a value to use for `rx`/`ry` instead of passing it a percentage value makes it easier to render as is done in CSS. This way we avoid the case where having multiple rectangles with percentage rx/ry will render differently 

> Exemple of what was wrong with % values in `rx`, 5% border-radius behaved differently if it was rendered with other shapes.
example2

example4


The following exemples are rendered with this new function, calculating the border radius to set it with a number, not a percentage

> Multiple shape with border-radius 10%


> Html/css with border-radius 10%, it is the same shape 👍 


> A shape with border-radius 35% and the same shape using html/css with border-radius 35%



> Border-radius 50%



## Questions ❓ 
* Do you think this is a good way approach this feature ?

If yes -> What bothers me a bit is that `calculateAxisRadius` now depends on either `Height` or `Width` when dealing with percentages values, and I did not see any other example of a computed value like this in this file. So we need to check if the "side length" value equality to `math.MaxFloat64` one more time here.
The code itself could be refactored to be more clear (maybe something like `out += el.getRx()` instead of `if el.Rx != math.MaxFloat64 {out += fmt.Sprintf(' rx="%f"', calculateAxisRadius(el.Rx, el.Width))}` to handle necessary check on Rx and the conditional check on Width in the same function `el.getRx()`, instead of doing validity check outside and inside the "calculate" function)


* Does this issue includes pill shapes ?

Another point of attention is that with only one border-radius percentage value, it is not possible to have pill shape like this.


To achieve that there is multiple solutions, but none with % in rx or  % border-radius in CSS. For exemple one solution is to set `rx` to 50% of the height of the rectangle (and give either no value or the same value to `ry`).

So the current implementation is good at replicating CSS when given a % value in `border-radius`, but cannot handle pill shape (CSS can't either with a  % `border-radius`).
Does this PR have to handle pill shape or do you think this is another issue ? If it has to, how do you think it should be implemented ? Add a `pill: true` property to the rectangle shape or create a new pill shape (that could be derived from rectangle shape in code if necessary)?

d2graph/d2graph.go
Poivey: ✅ Done

d2exporter/export.go
Poivey, line 1: There is no test yet, I would like your input on this solution before I make them since I didn't have more time & I'm not sure if that is satisfactory. I should have few hours to work on this PR next week to make/update tests if all is good.

NB : This comment does not refers to this file in particular, it's just to make a comment that can be resolved.

d2themes/element.go
alixander: apologies for the late response here. 

> This way we avoid the case where having multiple rectangles with percentage rx/ry will render differently

> What bothers me a bit is that calculateAxisRadius now depends on either Height or Width when dealing with percentages value

I don't understand, isn't that the point of percentages? To adapt to dimensions?

> if it was rendered with other shapes.

i also don't understand. what shapes are you using? are those all rectangles or are some ovals?

> Does this issue includes pill shapes ?

Firstly, really appreciate you thinking of motivating use cases like this. I wonder if this would work https://stackoverflow.com/a/18795153/20901387 . Border radius currently has a ceiling, but we could remove that for this use case (https://github.com/terrastruct/d2/issues/948).

Poivey: >This way we avoid the case where having multiple rectangles with percentage rx/ry will render differently

I wanted to highlight the issue that two rectangles with the same percentage `rx` did not render the same way if rendered in svg of different size. Eg. In the first two images of this thread, https://user-images.githubusercontent.com/15009534/224089772-b5101095-1357-4060-aaa2-226591cf452e.png https://user-images.githubusercontent.com/15009534/224089786-5ba28b39-4552-4e9e-a055-dea6e1a753c4.png, the two rectangle "ElementWithALongerName0,05" have a border-radius of 5% (rx) but are not rendered the same. That is why I explored the solution of computing a "number" value for rx/ry instead of simply passing a percentage value.

>I don't understand, isn't that the point of percentages? To adapt to dimensions?

Sure I agree 😄. I think this solution can work. I'm not familiar enough with the project to know if we will always have a `height` & `width` values when passing a `border-radius`, if we don't then it's impossible in those cases to compute the value and we would have to pass the raw percentage value, or nothing. What do you think about it, would it be acceptable to "compute" a rx/ry when given a % border-radius? Or would we prefer passing the raw % value to our shape and find another solution to make sure the way they render is consistant (solve the issue of the "ElementWithALongerName0,05" example)?

> I also don't understand. what shapes are you using? are those all rectangles or are some ovals?

What I meant by "if it was rendered with other shapes." is that "ElementWithALongerName0,05" is not rendered the same in the two exemple because the second example has other shapes (rect) in it, that is somehow causing a different render of its shape even if they have the same border-radius % value. I wanted to illustrate the issue I had with % rx/ry values.
That is why I tried to compute a non-percentage rx/ry value from a % border-radius value, that solution is illustrated with the other examples.
Every example you see are svg rect shapes with different border-radius values. The orange one are html/css with border-radius value to compare the results with.

> Firstly, really appreciate you thinking of motivating use cases like this. I wonder if this would work https://stackoverflow.com/a/18795153/20901387 . Border radius currently has a ceiling, but we could remove that for this use case (https://github.com/terrastruct/d2/issues/948).

I tried to do this (passing a huge value to rx/ry inside a rect), but that just renders rectangles as oval (like having %50 border-radius). The best solution I found for making a pill is to set `rx` to half of the `height` value. So we could always compute the `rx` value using the given percentage and the smallest value between `width` and `height`, and 50% would be a pill shape, but that would make us diverge a lot from CSS.
So if you consider the pill shape to be part of this issue, we have to find a solution (maybe creating a pill shape, of maybe a props to rect shape, I'm not sure which is the best). But if you don't, this PR can focus on replicating the CSS `border-radius` with a percentage value

alixander: > I wanted to highlight the issue that two rectangles with the same percentage rx did not render the same way if rendered in svg of different size

wait are you saying that it's dependent of the whole diagram size? not just the object size?

can you copy the .d2 code for those examples? or just include them as tests in `todo_test` and generate the svgs

Poivey: Exactly, that's what I have found but I might miss something as I'm kinda new on this kind of subject.
Here is a gist to illustrate it https://gist.github.com/Poivey/d965e66456925e41c19a3f179b18a0ee

As you can see here, both `percent-5percent-*.svg` files are containing a `rect` with `rx` and `ry` with `5%` value. But `ElementWith5PercentBorder` is not rendered the same way in the two file, when it is the only shape in the svg ([percentage-5percent-alone.svg](https://gist.github.com/Poivey/d965e66456925e41c19a3f179b18a0ee#file-percentage-5percent-alone-svg)) its border radius is rendered a different way than when it is rendered with other elements in the svg ([percentage-5percent-with-other-elements.svg](https://gist.github.com/Poivey/d965e66456925e41c19a3f179b18a0ee#file-percentage-5percent-with-other-elements-svg)).

I also added two file [numeric-5percent-with-other-elements.svg](https://gist.github.com/Poivey/d965e66456925e41c19a3f179b18a0ee#file-numeric-5percent-with-other-elements-svg) and [numeric-5percent-alone.svg](https://gist.github.com/Poivey/d965e66456925e41c19a3f179b18a0ee#file-numeric-5percent-alone-svg) that show that when computing a numeric value (and not a %) to set `rx` and  `ry` with, the `rect` is rendered the same way every time AND matches the shape of a css element with `border-radius: 50%` -> This is the current state of this PR

d2 souce is in the gist, it's just this. `AnotherElementX` are commented when rendering the rectangle with border-radius alone.
```
ElementWith5PercentBorder.style.border-radius: 0.05
AnotherElement1
AnotherElement2
AnotherElement3
AnotherElement4
```

alixander (commented):
Screen Shot 2023-04-06 at 3 27 49 PM

Screen Shot 2023-04-06 at 3 30 25 PM

hm yeah these just look weird, i don't think anyone would want to use that. i guess what i really wanted was a way to create the "pill" shape effect. sorry about this dragged-out experience @Poivey . i rly appreciate this high quality investigating and work and how thorough you've been in describing the issues, but i think the end result just didn't turn out to be something desirable, aesthetically. 

i also don't want to have a separate `pill: true`, or `shape: pill` though. i kind of wish the `border-radius: 999` would work, like it does in CSS. up to you if you want to keep going to find the implementation for that, otherwise i can just create an issue for it separately.

alixander:
> For exemple one solution is to set rx to 50% of the height of the rectangle

oh, what if % only applies to rx then?

Poivey:
> up to you if you want to keep going to find the implementation for that

Sure I would like to continue, I'll have time to work on it next week.

As I understand : the goal is to have a rectangle ressemble a pill shape if given a `border-radius` between 0 and 1. The rectangle will be more or less "pill-like" according to the value of `border-radius`.

Can you confirm this is what needs to be done ?

> what if % only applies to rx then, and just not affect the ry? that'd be a clear use case for %, and i think we don't lose anything since % just doesn't look good when applied to both rx and ry

Yes, from what I saw when I experimented, it's possible to have a pill shape like this. For example, if a rectangle has a `height` of `50` and a `width` of `100`, to be a pill shape it would need to have `rx` set to `25` (50% of the height).

It's not the same as setting `rx` to 50%, as it is calculated on the width by default 😄
Setting only `rx` is the same as setting `rx` and `ry` to the same value, so in this case only `rx` will be needed

alixander:
> the goal is to have a rectangle ressemble a pill shape if given a border-radius between 0 and 1.

No I think we want to just emulate CSS's handling. Users can specify a large value for "pill". you'll probably have to increase the allowed value, which is a todo anyway (https://github.com/terrastruct/d2/issues/948)

https://stackoverflow.com/questions/71023962/how-to-make-a-pill-shape-button-with-html-and-css

https://stackoverflow.com/questions/31617136/avoid-elliptical-shape-in-css-border-radius
"If the value exceeds half of the shortest side's length, the browser will use the minimum as its border-radius in both directions, producing a perfect pill shape on rectangular elements."

d2themes/element.go
```diff
@@ -227,3 +227,9 @@ func (el *ThemableElement) Render() string {
 	}
 	return out
 }
+
+func calculateAxisRadius(borderRadius float64, width float64, height float64) float64 {
+	var minimumSideSize = math.Min(width, height)
+	var maximumBorderRadiusValue = minimumSideSize / 2.0
```
alixander, line 233: initialize with `:=`

alixander (approved):
💯 beautiful docs, tests, and code. minor nitpicks, but other than that + merge conflicts, good to go!

alixander:
friendly ping @Poivey , would love to get this in

d2themes/element.go
```diff
@@ -227,3 +227,9 @@ func (el *ThemableElement) Render() string {
 	}
 	return out
 }
+
+func calculateAxisRadius(borderRadius float64, width float64, height float64) float64 {
```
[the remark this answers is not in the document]
Poivey: ✅

d2themes/element.go
Poivey: ✅

Poivey:
@alixander It's ready :)

alixander (approved):

alixander:
thanks again!

Mapaor4:
Does it visually render connections properly tho?
The following code:
```
Hello: {style.border-radius: 999}
Hello -> John
Hello -> Tim
Hello -> George
```
Produces:

Which is not bad but also not as expected. Shouldn't connections start and end in contact with the nodes?
I think it looks worse with grid diagrams:

PR #1011, opened 2023-03-10
dagre: improve inter-container connections

alixander:
before, we were setting all connections between containers to be the first child of the container.

But this often led to awkward placements on the side, since the rank would be calculated to equal a child.

Instead, if the source is a container, we get the tail of the longest edge chain. if the dst is a container, we get the first head.

Screen Shot 2023-03-09 at 7 41 01 PM


alixander:
cc @gavin-ts the dagre man

Issue #955, opened 2023-03-02
Uncaught syntax error leads to bug in render


BarryNolte:
This;
```x <= y```

Leads to:

![image](https://user-images.githubusercontent.com/4474173/222462820-6570e0a2-caba-4129-b609-4522d7a32b9f.png)

Completes successfully, no errors.

**Comment by nhooyr** [comment]
None of these are syntax errors. This is completely expected behaviour and if you run `d2 fmt` on the input you'll see that the non edge syntax characters have a space added between them and the edge syntax.

So

```d2
x <= y
```

Becomes:

```d2
x <- = y
```

As expected. I understand it's confusing but that's why `d2` always formats your input files with `d2 fmt` when you compile too. To make it clear what's going on.

Issue #322, opened 2022-12-02
Unexpectedly merging shapes


pleshevskiy:
Tested on dagre and elk layouts

```d2
a-
b-
c-
```

# Expected

![image](https://user-images.githubusercontent.com/7839004/205332558-07e6a4da-2664-4a1f-b138-546e79d01168.png)

# Actual

![image](https://user-images.githubusercontent.com/7839004/205332594-5feb0454-c360-4acd-961e-c38d2f53e421.png)

nhooyr:
Can confirm this is a bug.

PR #1021, opened 2023-03-13
testdata: Add a few parse tests and cleanup testdata

nhooyr:
Updates #955
Updates #322


nhooyr:
@alixander all the testdata deletions are from tests that no longer exist.

nhooyr:
Or were moved around I should say.

alixander (approved):

Issue #1027, opened 2023-03-14
e2etests-cli: got.pdf after running tests


gavin-ts:
Running tests leaves this untracked (but not git ignored file)
![Screen Shot 2023-03-14 at 11 21 13 AM](https://user-images.githubusercontent.com/85081687/225101775-519155b0-b1d9-4c93-a9af-511ebb87b1d5.png)

PR #1028, opened 2023-03-14
elk: favor balance over straight edges, margins

alixander:
why the fuck are some of the edges unnecessarily windy? idk.

but i do think it's generally better

Screen Shot 2023-03-14 at 12 52 06 PM

also adds some margins between edges and nodes.

Screen Shot 2023-03-14 at 1 05 21 PM


e2etests-cli/testdata/TestCLI_E2E/internal_linked_pdf.exp.pdf
gavin-ts, line 1: #1027

PR #1052, opened 2023-03-18
browser flag

### LINKED ISSUE #1049: cli:browser=0
\>d2.exe --watch ```--browser=0``` (#175)

alixander:
Screen Shot 2023-03-17 at 8 09 24 PM

Screen Shot 2023-03-17 at 8 09 56 PM

closes #1049


d2cli/main.go
```diff
@@ -83,6 +83,10 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
 	if err != nil {
 		return err
 	}
+	browserFlag, err := ms.Opts.Bool("BROWSER", "browser", "", true, "open the browser and go to the address that watch streams updates to")
```
nhooyr, line 86: Could you make this a string and interpet exactly like `$BROWSER`? So you'd just `ms.Env.Setenv("BROWSER", * browserFlag)` if the flag was set.

nhooyr (changes requested):

d2cli/main.go
alixander: all this is doing is validation though. it'll take 1 too. it's just helping users make less mistakes

nhooyr: But browser can be used to point to a specific executable too. So --browser and BROWSER aren't equal which doesn't seem good to me.

nhooyr (changes requested):
🔴

d2cli/main.go
alixander: oh, i didn't know that. okay

nhooyr (approved):

Issue #848, opened 2023-02-19
constant near keys should be able to be set on containers


alixander:
as long as nothing within is connected to anything outside

donglixiaoche:
@alixander yep,a hard one,exactly what I want

alixander:
@donglixiaoche 
so, for context, you can set `near` to constant values on individual shapes. it's implemented here: https://github.com/terrastruct/d2/pull/525

but people commonly want to make a legend, which is a container of stuff, and put it in the top-center.

setting constant nears works roughly like:
1. identify which objects are constant nears
2. remove them before the main layout engine runs
3. run main layout engine
4. add them back in to the appropriate spot

for containers, i think you'll have to
1. validate in the d2compiler that nothing in the container or descendant of container is connected to anything outside, (because if it is, it has to run in the main layout engine)
2. remove that container and all descendants of that container, including connections between children of that container
3. add all those things back in after main layout engine runs

donglixiaoche:
@alixander  just so you know, i'm still working on this issue...

as you mentioned, for shapes with near keyword, before main layout,  i need to remove that container and all descendants of that container. 
but if having descendants is allowed for container with near keyword, these descandants also need to be layouted, which can't be done if we remove them from the graph.
so i was thinking reolve this problem by contrusting a new graph, and then run another main layout? am i correct? or do you have any ideas?

alixander:
@donglixiaoche oh that's a good point, i didn't think about that. 

1. It looks like the core layout is coupled with sequence diagram layouts. that should probably be separated: https://github.com/terrastruct/d2/blob/7ba463c690d5771d42fdf808c1c33adb3b0d35ad/d2lib/d2.go#L73
2. The descendants would also need to undergo layout, but they themselves might also have `near` containers, or sequence diagrams. So it should be recursive. Which means this whole portion of layout code should probably be separated so that it can be run recursively: https://github.com/terrastruct/d2/blob/7ba463c690d5771d42fdf808c1c33adb3b0d35ad/d2lib/d2.go#L66-L81 . By doing this, perhaps (1) won't be necessary.

> contrusting a new graph, and then run another main layout?

yes, each constant-near-container would have to be its own graph. so you remove all the constant-near-containers, run layout on all of them (recursively), then run on the main graph, then re-insert all the objects and edges back into the main graph.

quite complicated, let me know when you run into anything at all, always available to help

donglixiaoche:
[...]

Issue #1031, opened 2023-03-14
shim elk: reattach to side ports


alixander:
no need to wind around

<img width="841" alt="Screen Shot 2023-03-14 at 3 02 32 PM" src="https://user-images.githubusercontent.com/3120367/225151639-2f5acf8b-7bbb-4f7a-9537-d4f4651fba96.png">

Issue #1001, opened 2023-03-08
Multi-colored text


Filip-K:
Hi! I'm trying to create a legend which explains the colors used in my diagram, but I'm a bit at loss with creating colored text that does not overlap. Here's as far as I could get ([link to the playground](https://play.d2lang.com/?l=&script=pJExCsMwDEV3neJfwIV21NC71K3aqRYIDXZL7h5kg5MhW0bD4z_zNJszLMIQVaQ7GuNPQHdchprxMZFCC1EL5HeEmLwC6O2vjLfqgCIKw6U6AVGSkdVdv-kpxcXm0Ha6na-_b4z8sBNz429rAAAA__8%3D&))
![d2](https://user-images.githubusercontent.com/16614296/223767096-93dbda60-2c10-4b84-ba78-1cfb2416dfc2.png)

Close enough, but the legend entries `foo` and `bar` overlap. Any ideas?

alixander:
@Filip-K ah, yeah multiple `near` set to the same constant will overlap.

The way to do this will be to put both items in a container and then set the `near` on just that container. But that's currently a TODO. We'll get to this in the next week or so. https://github.com/terrastruct/d2/issues/848

https://play.d2lang.com/?l=&script=jJExDsIwDEV3n-Ij5iLB6IGJvUsvQKlpK0EipR5aUO6O7KiwMHRM9Oxvvf91zkgmhmhGdcbCeBPgGYcSzeiTSKBMtBjy-ock6Qx4SC-hK8AezTBOGCfoIGjqS73zX_PJaKNqfFY3CSqJAG_tyLjH6NOAC2WozFreHvireL0LyLQuODHaa9q8wK628UyfAAAA__8%3D&

Filip-K:
I see, thanks! The trick with putting a legend into a container is already pretty good, even if it cannot be set `near` yet. Feel free to close this issue now or when #848 is done.

alixander:
great to hear! I'll close this then in favor of that one

PR #1071, opened 2023-03-21
Near keys for container

donglixiaoche:
#848 

now descendants are allowed for contianer with `near` attribute

the screenshot can be generated by run script:
```
./ci/e2ereport.sh -test-case=near_keys_for_container
```
image


donglixiaoche:
https://github.com/terrastruct/d2/blob/9ab0d0f279e2e5ba86c22c1282ed19a62b4cf530/d2layouts/d2near/layout.go#L43-L52
@alixander  these codes that set labelPosition of `near` shape should be removed once we allow descendant, but if outside label is allowed, does that mean these containers should also be included during the core layout? we shouldn't let label position affect the main layout

alixander:
> these codes that set labelPosition of near shape should be removed once we allow descendant, but if outside label is allowed, does that mean these containers should also be included during the core layout? we shouldn't let label position affect the main layout

why should it be removed? we need to place the labels somewhere. maybe we could place it depending on direction, so that it'll never collide with core layout. e.g. the `top-x` near objects get labels placed on outside_top, the `bottom-x` objects get labels to outside_bottom, etc

d2compiler/compile_test.go
```diff
@@ -1524,26 +1524,6 @@ d2/testdata/d2compiler/TestCompile/near-invalid.d2:14:9: near keys cannot be set
...
-
-			text: `x: {
-  near: top-center
-  y
-}
-`,
-			expErr: `d2/testdata/d2compiler/TestCompile/near_bad_container.d2:2:9: constant near keys cannot be set on shapes with children`,
-		},
-		{
-			name: "near_bad_connected",
```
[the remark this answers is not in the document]
donglixiaoche: is there a command that can auto genereate `exp.json` file?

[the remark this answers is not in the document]
alixander: if you just run `TA=1` before any test, it'll regenerate anything in testdata.

(stands for TESTDATA_ACCEPT)

`TA=1 ./ci/test.sh ./d2compiler` for this one

d2layouts/d2near/layout.go
```diff
@@ -77,12 +107,16 @@ func place(obj *d2graph.Object) (float64, float64) {
 	case "bottom-right":
 		return br.X + pad, br.Y + pad
 	}
+
```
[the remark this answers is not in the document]
donglixiaoche: fixed

d2layouts/d2near/layout.go
```diff
@@ -36,27 +55,38 @@ func Layout(ctx context.Context, g *d2graph.Graph, constantNears []*d2graph.Obje
...
+			g.Objects = append(g.Objects, child)
+			attachChildren(g, child)
+		}
+	}
+}
+
 // place returns the position of obj, taking into consideration its near value and the diagram
 func place(obj *d2graph.Object) (float64, float64) {
 	tl, br := boundingBox(obj.Graph)
+
```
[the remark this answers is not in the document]
donglixiaoche: fixed

d2compiler/compile_test.go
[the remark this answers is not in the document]
donglixiaoche: a connection that connect to a child of a `near` container alos shoulldn't be allowed, right?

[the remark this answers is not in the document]
alixander: if it connects to something outside that near container, correct. if it's inside, it should be allowed of course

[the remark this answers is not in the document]
donglixiaoche: fixed, thx for pointing this out, i forgot to validate this logic

e2etests/stable_test.go
```diff
@@ -12,6 +12,26 @@ var testMarkdown string
 
 func testStable(t *testing.T) {
 	tcs := []testCase{
+		{
+			name: "near_keys_for_container",
```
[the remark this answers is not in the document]
donglixiaoche: done

donglixiaoche:
> > these codes that set labelPosition of near shape should be removed once we allow descendant, but if outside label is allowed, does that mean these containers should also be included during the core layout? we shouldn't let label position affect the main layout
> 
> why should it be removed? we need to place the labels somewhere. maybe we could place it depending on direction, so that it'll never collide with core layout. e.g. the `top-x` near objects get labels placed on outside_top, the `bottom-x` objects get labels to outside_bottom, etc

well, i may misunderstood a little bit, but after run core layout for every `near` container, the label placement of near container still be skipped? what i thought is now descendant is allowed, so the label may be placed outside, so i need to consider the dimension of label as well when put these near container inside main layout

alixander:
@donglixiaoche 
> well, i may misunderstood a little bit, but after run core layout for every near container, the label placement of near container will still be skipped? what i thought is now descendant is allowed, so the label may be placed outside, so i need to consider the dimension of label as well when put these near container inside main layout

actually... i think you're right. if you run core layout on the near object, the label placement should happen too.

ohh and so what you're calculating now is this new x, y right? the position considering the label, otherwise, the label might collide with the main diagram?

Screen Shot 2023-03-22 at 10 15 52 PM

donglixiaoche:
> @donglixiaoche
> 
> > well, i may misunderstood a little bit, but after run core layout for every near container, the label placement of near container will still be skipped? what i thought is now descendant is allowed, so the label may be placed outside, so i need to consider the dimension of label as well when put these near container inside main layout
> 
> actually... i think you're right. if you run core layout on the near object, the label placement should happen too.
> 
> ohh and so what you're calculating now is this new x, y right? the position considering the label, otherwise, the label might collide with the main diagram?
> 
> Screen Shot 2023-03-22 at 10 15 52 PM

yep, that's what i tried to do, but i found there are still some little bugs to fix, working on it, sorry for deleted for comments 🤣

alixander:
@donglixiaoche no. i can't remember what the exceptions were, but it can't be assumed to always be 0, 0

donglixiaoche:
@alixander  you can now continue the code review process

e2etests/stable_test.go
```diff
@@ -1894,576 +1929,6 @@ i am bottom left: { shape: text; near: bottom-left }
 i am bottom right: { shape: text; near: bottom-right }
 `,
 		},
-		{
```
alixander, line 1897: why were so many cases removed?

d2lib/d2.go
```diff
@@ -68,14 +68,27 @@ func compile(ctx context.Context, g *d2graph.Graph, opts *CompileOptions) (*d2ta
 			return nil, err
 		}
 
-		constantNears := d2near.WithoutConstantNears(ctx, g)
+		constantNears, constantNearGraphs := d2near.WithoutConstantNears(ctx, g)
```
alixander, line 71: do you need two variables? wouldn't constantNears just be the roots of the constantNearGraphs?

d2graph/d2graph.go
```diff
@@ -90,11 +90,12 @@ type Object struct {
...
-	LabelPosition *string `json:"labelPosition,omitempty"`
-	LabelWidth    *int    `json:"labelWidth,omitempty"`
-	LabelHeight   *int    `json:"labelHeight,omitempty"`
-	IconPosition  *string `json:"iconPosition,omitempty"`
+	*geo.Box              `json:"box,omitempty"`
+	LabelPosition         *string `json:"labelPosition,omitempty"`
+	LabelWidth            *int    `json:"labelWidth,omitempty"`
+	LabelHeight           *int    `json:"labelHeight,omitempty"`
+	IconPosition          *string `json:"iconPosition,omitempty"`
+	IsInsideNearContainer bool    `json:"isInsideNearContainer,omitempty"`
```
alixander, line 98: shouldn't need to have this variable. if necessary, you can define something similar to this method:
https://github.com/terrastruct/d2/blob/718daee97c688af1d22605139ffb4c3d71e27c13/d2graph/seqdiagram.go#L9

alixander (changes requested):

e2etests/stable_test.go
donglixiaoche: delete by mistake, reverted

d2graph/d2graph.go
donglixiaoche: isn't it more efficient if we store this info? only one time iteration is enough.

d2lib/d2.go
donglixiaoche: fixed, and regerated some testcases

d2graph/d2graph.go
alixander: it is, but it's negligible. better to keep data structure minimal in this case

d2lib/d2.go
donglixiaoche: @alixander i found that the sequence of keys of golang map is uncertain, which may need an extrat sorting, do you think it's better than two variables?

alixander: can `constantNearGraphs` be an array then?

donglixiaoche: yep, quite strightforward..., just need to use `tempGraph.Root.ChildrenArray[0]` to retrieve the near object. fixed

d2graph/d2graph.go
donglixiaoche: fixed, defined as method  `IsInsideNearContainer ` instead

alixander (changes requested):
can you add this test? label still off. (got it from an issue with the motivating use case for this feature: https://github.com/terrastruct/d2/issues/1001 )

```
direction: right

x -> y: {
  style.stroke: green
}

y -> z: {
  style.stroke: red
}

legend: {
  # This is the TODO!
  near: bottom-center
  color1: foo {
    shape: text
    style.font-color: green
  }

  color2: bar {
    shape: text
    style.font-color: red
  }
}
```

Screen Shot 2023-03-26 at 8 45 28 PM

donglixiaoche:
@alixander previously there's bug when calc new x,y, fixed, and added a new testcase

donglixiaoche:
to clarify the logic of recalculating new x,y:
if a near container has an outside label, and the container is
1. placed on the top, and have a label on the bottom: shift upward the height of label
2. placed on the bottom, and have a label on the top: shift downward the height of label
3. placed on the left, and have a label on the right: shift leftward the width of label
4. placed on the right and have a label on the left: shift rightward the width of label

e2etests/testdata/regression/unconnected/elk/board.exp.json
```diff
@@ -531,7 +531,6 @@
       "underline": false,
       "labelWidth": 639,
       "labelHeight": 51,
-      "labelPosition": "INSIDE_MIDDLE_CENTER",
```
alixander, line 534: why did this go away?

d2graph/d2graph.go
```diff
@@ -968,6 +968,18 @@ func (obj *Object) GetDefaultSize(mtexts []*d2target.MText, ruler *textmeasure.R
 	return &dims, nil
 }
 
+func (obj *Object) IsInsideNearContainer() bool {
```
alixander, line 971: have this return the near container object and call it `OuterNearContainer()` or `NearAncestor()`

d2compiler/compile.go
```diff
@@ -735,8 +735,18 @@ func (c *compiler) validateNear(g *d2graph.Graph) {
 					c.errorf(obj.Attributes.NearKey, "constant near keys can only be set on root level shapes")
 					continue
 				}
-				if len(obj.ChildrenArray) > 0 {
-					c.errorf(obj.Attributes.NearKey, "constant near keys cannot be set on shapes with children")
+
+				descendantsMap := getNearDescendants(obj)
+				connectToOutside := false
+				for _, edge := range g.Edges {
```
alixander, line 741: simplify. loop through edges and make sure that the near ancestor of both endpoints match. no need for the extra map. 

i know it's less efficient, but it's negligible and simplifies code

d2compiler/compile.go
```diff
@@ -747,6 +757,24 @@ func (c *compiler) validateNear(g *d2graph.Graph) {
 	}
 }
 
+func getNearDescendants(nearObj *d2graph.Object) map[*d2graph.Object]bool {
```
alixander, line 760: don't need

d2lib/d2.go
```diff
@@ -68,14 +68,28 @@ func compile(ctx context.Context, g *d2graph.Graph, opts *CompileOptions) (*d2ta
 			return nil, err
 		}
 
-		constantNears := d2near.WithoutConstantNears(ctx, g)
+		constantNearGraphs := d2near.WithoutConstantNears(ctx, g)
+
+		// run core layout for constantNears
+		for _, tempGraph := range constantNearGraphs {
+			if tempGraph == nil {
```
alixander, line 75: why would it be nil?

d2lib/d2.go
```diff
@@ -68,14 +68,28 @@ func compile(ctx context.Context, g *d2graph.Graph, opts *CompileOptions) (*d2ta
...
+				continue
+			}
+
+			nearObject := tempGraph.Root.ChildrenArray[0]
+			nearObject.Parent = tempGraph.Root
```
alixander, line 80: is this doing anything?

d2layouts/d2near/layout.go
```diff
@@ -94,8 +150,16 @@ func WithoutConstantNears(ctx context.Context, g *d2graph.Graph) (nears []*d2gra
 		}
 		_, isConst := d2graph.NearConstants[d2graph.Key(obj.Attributes.NearKey)[0]]
 		if isConst {
-			nears = append(nears, obj)
-			g.Objects = append(g.Objects[:i], g.Objects[i+1:]...)
+			descendantObjects, edges := pluckOutNearObjectAndEdges(g, obj)
+
+			tempGraph := d2graph.NewGraph()
+			tempGraph.Root.ChildrenArray = []*d2graph.Object{obj}
+			tempGraph.Root.Children[obj.ID] = obj
```
alixander, line 157: would it work to set the Root as obj directly? then all the children and childrenarray are already correctly set

d2layouts/d2near/layout.go
```diff
@@ -106,7 +170,34 @@ func WithoutConstantNears(ctx context.Context, g *d2graph.Graph) (nears []*d2gra
...
 		}
 	}
-	return nears
+	return constantNearGraphs
+}
+
+func pluckOutNearObjectAndEdges(g *d2graph.Graph, obj *d2graph.Object) (descendantsObjects []*d2graph.Object, edges []*d2graph.Edge) {
+	for i := 0; i < len(g.Edges); i++ {
+		edge := g.Edges[i]
+		if edge.Src == obj || edge.Dst == obj {
```
alixander, line 179: i think this might double append. can you add this as a test. i think the edge would double append in this case

```
n: {
  near: top-center
  x
  inner: {
    y
  }
  x -> y
}
a; b; c
```

alixander (changes requested):

alixander:
> is there a way to mock label positon? how can i create label of different positions?

there is, something like https://github.com/donglixiaoche/d2/blob/66f620f57b7e2af5102d5f23e1d869a11cc6a010/d2target/d2target.go#L204

e2etests/testdata/regression/unconnected/elk/board.exp.json
donglixiaoche: i this it's becasue i removed this code snippet

image

d2lib/d2.go
donglixiaoche: yep, seems like `dagre` need this to do the layout work

d2compiler/compile.go
donglixiaoche: deleted

d2lib/d2.go
donglixiaoche: fixed

d2layouts/d2near/layout.go
alixander: ohh i see.

d2layouts/d2near/layout.go
```diff
@@ -94,8 +150,16 @@ func WithoutConstantNears(ctx context.Context, g *d2graph.Graph) (nears []*d2gra
...
 		_, isConst := d2graph.NearConstants[d2graph.Key(obj.Attributes.NearKey)[0]]
 		if isConst {
-			nears = append(nears, obj)
-			g.Objects = append(g.Objects[:i], g.Objects[i+1:]...)
+			descendantObjects, edges := pluckOutNearObjectAndEdges(g, obj)
+
+			tempGraph := d2graph.NewGraph()
+			tempGraph.Root.ChildrenArray = []*d2graph.Object{obj}
+			tempGraph.Root.Children[obj.ID] = obj
+			tempGraph.Objects = descendantObjects
```
alixander, line 158: i think the splitting of graphs has to be more robust. currently it's the bare minimum to get a layout working, but a layout code may change and depend on other parts of the graph structure that are not properly split.

specifically,
- each descendant object has to have all their references changed. the graph reference has to be updated: https://github.com/donglixiaoche/d2/blob/b9dd247c75013d0581d15ae2bb3490ca86e0ba4f/d2graph/d2graph.go#L79

- The parent of the top-level object has to be updated.

- they should be removed from the previous root's children and childrenarray.


and then when it's merged back in, all those should be updated again to be the same as before.

put another way: when the core layout is running on each of these graphs, it should have no way of reaching an object in a separate graph.

donglixiaoche: > i know that's a little ambiguous, but having a clean way to do this will be useful for more than just `near`. later when D2 supports other graph types, we'll want to run different layout algorithms for those too, requiring a clean split

totally argee with you, fixed

d2compiler/compile.go
```diff
@@ -735,8 +735,20 @@ func (c *compiler) validateNear(g *d2graph.Graph) {
...
+					dstNearContainer := edge.Dst.OuterNearContainer()
+
+					if srcNearContainer != dstNearContainer {
+						connectToOutside = true
+						break
+					}
+				}
+
+				if connectToOutside {
+					c.errorf(obj.Attributes.NearKey, "a child of a near container cannot connect to outside")
```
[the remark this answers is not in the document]
donglixiaoche: fixed

donglixiaoche:
@alixander  seems like there are some conflicts, what is the correct way to resolve them? how about i just use these files from master branh, and then rerun `TA=1 ./ci/test.sh`?

e2etests/testdata/regression/unconnected/elk/board.exp.json
donglixiaoche: any issues to remove this code snippet?

alixander:
> @alixander seems like there are some conflicts, what is the correct way to resolve them? how about i just use these files from master branh, and then rerun TA=1 ./ci/test.sh?

yeah exactly, what i do is just accept all conflicts without modifying, and then regenerate after

d2compiler/compile.go
```diff
@@ -735,8 +735,20 @@ func (c *compiler) validateNear(g *d2graph.Graph) {
 					c.errorf(obj.Attributes.NearKey, "constant near keys can only be set on root level shapes")
 					continue
 				}
-				if len(obj.ChildrenArray) > 0 {
-					c.errorf(obj.Attributes.NearKey, "constant near keys cannot be set on shapes with children")
+
+				var edgeConnectToOutsideOfNear *d2graph.Edge
```
alixander, line 739: no need for another variable, just report the error when you run into it

d2compiler/compile.go
```diff
@@ -735,8 +735,20 @@ func (c *compiler) validateNear(g *d2graph.Graph) {
...
+				for _, edge := range g.Edges {
```
alixander, line 740: this loop should be outside the surrounding loop.

d2compiler/compile.go
```diff
@@ -735,8 +735,20 @@ func (c *compiler) validateNear(g *d2graph.Graph) {
...
+					srcNearContainer := edge.Src.OuterNearContainer()
+					dstNearContainer := edge.Dst.OuterNearContainer()
+
+					if srcNearContainer != dstNearContainer {
+						edgeConnectToOutsideOfNear = edge
+						break
```
alixander, line 746: don't break. we want to get all the errors

alixander (changes requested):

d2compiler/compile.go
donglixiaoche: fixed

d2compiler/compile.go
donglixiaoche: fixed, the new logic also need to validae it's `isConstant` or `isKey`

d2compiler/compile.go
```diff
@@ -749,16 +749,32 @@ func (c *compiler) validateNear(g *d2graph.Graph) {
...
 				continue
 			}
 		}
 	}
+
+	for _, edge := range g.Edges {
+		srcNearContainer := edge.Src.OuterNearContainer()
+		dstNearContainer := edge.Dst.OuterNearContainer()
+
+		var isSrcNearConst, isDstNearConst bool
```
alixander, line 764: nice, good consideration

d2graph/d2graph.go
```diff
@@ -968,6 +968,18 @@ func (obj *Object) GetDefaultSize(mtexts []*d2target.MText, ruler *textmeasure.R
 	return &dims, nil
 }
 
+func (obj *Object) OuterNearContainer() *Object {
+	temp := obj
```
alixander, line 972: i don't think u need temp. just change `obj`

d2graph/d2graph.go
```diff
@@ -968,6 +968,18 @@ func (obj *Object) GetDefaultSize(mtexts []*d2target.MText, ruler *textmeasure.R
...
+	for temp != nil {
+		// find outer near container that isn't itself
```
alixander, line 974: why exclude itself? doesn't that mean this code would let you connect two near containers?
https://github.com/terrastruct/d2/pull/1071/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-82cd6b1f1fa955a8d32bdac010cfbd436bc544dd3b445faf012100e8eb3f7a0dR773

alixander: i tested and it's prevented somewhere else, but i still think there's no need to exclude itself

d2lib/d2.go
alixander: it seems like this belongs in `WithoutConstantNears`.

d2lib/d2.go
```diff
@@ -68,14 +68,27 @@ func compile(ctx context.Context, g *d2graph.Graph, opts *CompileOptions) (*d2ta
...
+		for _, tempGraph := range constantNearGraphs {
+			nearObject := tempGraph.Root.ChildrenArray[0]
+			nearObject.Parent = tempGraph.Root
+			if err = coreLayout(ctx, tempGraph); err != nil {
+				return nil, err
+			}
+			nearObject.Parent = g.Root
+			for _, obj := range tempGraph.Objects {
+				obj.Graph = g
+			}
```
alixander, line 83: and then this belongs in `d2near.Layout`

alixander (changes requested):
apologies for how long my reviews have taken @donglixiaoche . this is really good and i'm excited about this. tested locally across all the layout engine and works beautifully

d2layouts/d2near/layout.go
```diff
@@ -106,7 +174,34 @@ func WithoutConstantNears(ctx context.Context, g *d2graph.Graph) (nears []*d2gra
 			}
 		}
 	}
-	return nears
+	return constantNearGraphs
+}
+
+func pluckOutNearObjectAndEdges(g *d2graph.Graph, obj *d2graph.Object) (descendantsObjects []*d2graph.Object, edges []*d2graph.Edge) {
```
alixander, line 180: just `pluckObjAndEdges`. ("out" is implied). "near" is not referenced here.

d2layouts/d2near/layout.go
```diff
@@ -94,8 +150,20 @@ func WithoutConstantNears(ctx context.Context, g *d2graph.Graph) (nears []*d2gra
 		}
 		_, isConst := d2graph.NearConstants[d2graph.Key(obj.Attributes.NearKey)[0]]
 		if isConst {
-			nears = append(nears, obj)
-			g.Objects = append(g.Objects[:i], g.Objects[i+1:]...)
+			descendantObjects, edges := pluckOutNearObjectAndEdges(g, obj)
+
+			tempGraph := d2graph.NewGraph()
+			tempGraph.Root.ChildrenArray = []*d2graph.Object{obj}
+			tempGraph.Root.Children[obj.ID] = obj
```
alixander, line 157: lowercase to match: https://github.com/donglixiaoche/d2/blob/05a9498c992faa191b81cab437b051b0121b6752/d2graph/d2graph.go#L565

d2layouts/d2near/layout.go
alixander: nice, that looks clean to me.

d2layouts/d2near/layout.go
```diff
@@ -134,6 +229,9 @@ func boundingBox(g *d2graph.Graph) (tl, br *geo.Point) {
 				y2 = math.Max(y2, obj.TopLeft.Y+obj.Height)
 			}
 		} else {
+			if obj.OuterNearContainer() != nil {
```
alixander, line 232: what's this for?

d2layouts/d2near/layout.go
```diff
@@ -10,79 +10,135 @@ import (
...
-			obj.LabelPosition = go2.Pointer(string(label.OutsideBottomCenter))
-		} else if obj.Attributes.Icon != nil {
-			obj.LabelPosition = go2.Pointer(string(label.InsideTopCenter))
-		} else {
-			obj.LabelPosition = go2.Pointer(string(label.InsideMiddleCenter))
+	return nil
+}
+
+func attachChildren(g *d2graph.Graph, obj *d2graph.Object) {
+	if obj.ChildrenArray != nil && len(obj.ChildrenArray) != 0 {
```
alixander, line 69: you don't need this line

d2layouts/d2near/layout.go
```diff
@@ -10,79 +10,135 @@ import (
...
 
 	// Imagine the graph has two long texts, one at top center and one at top left.
 	// Top left should go left enough to not collide with center.
 	// So place the center ones first, then the later ones will consider them for bounding box
 	for _, processCenters := range []bool{true, false} {
-		for _, obj := range constantNears {
+		for _, tempGraph := range constantNearGraphs {
+			obj := tempGraph.Root.ChildrenArray[0]
 			if processCenters == strings.Contains(d2graph.Key(obj.Attributes.NearKey)[0], "-center") {
+				preX, preY := obj.TopLeft.X, obj.TopLeft.Y
```
alixander, line 30: small nitpick, but convention in this codebase is `prevX` instead of `preX`

d2layouts/d2near/layout.go
```diff
@@ -10,79 +10,135 @@ import (
...
 				obj.TopLeft = geo.NewPoint(place(obj))
+				dx, dy := obj.TopLeft.X-preX, obj.TopLeft.Y-preY
+
+				subObjects, subEdges := tempGraph.Objects, tempGraph.Edges
```
alixander, line 34: why this?

alixander (changes requested):

donglixiaoche:
@alixander no worries, actually i'm on vacation right now. and i know i could do better about the details. and i'd love to pick another hard one, do you got one for me?

alixander:
@donglixiaoche nice! where are you vacationing?

Does this sound interesting? https://github.com/terrastruct/d2/issues/1107

One of the biggest improvements in D2 0.3 was that font sizes got "subset", meaning reduced to only include the characters needed for them. but we don't split by styles. so the letter "b" may be used for italic, but we'll include it for regular and bold fonts as well. theoretically this should be another significant decrease in file size if we can further optimize by style.

Or if you want to stick to rendering stuff, this one could result in some big layout improvements: https://github.com/terrastruct/d2/issues/1031

Feel free to say "pass" and I'm more than happy to find more!

donglixiaoche:
@alixander haha, i'm now in Tibet watching those big views.

and #1107 sounds interesting, want to do some optimizations about font size 🙋🏻‍♂️

d2layouts/d2near/layout.go
donglixiaoche: once we append a near container inside the main layout, those objects inside a near container(which doesn't have near attribite either) shouldn't affect the way we calc bounding box

d2layouts/d2near/layout.go
donglixiaoche: fixed

d2layouts/d2near/layout.go
donglixiaoche: fixed

d2layouts/d2near/layout.go
donglixiaoche: fixed, use `tempGraph.Objects` and `tempGraph.Edges` directly

d2lib/d2.go
donglixiaoche: fixed

d2lib/d2.go
donglixiaoche: fixed

d2graph/d2graph.go
donglixiaoche: i recall i was trying to fix a ci issue, is i don't exclude itself, `TestCompile/near_bad_connected` this testcase will throw   more that one error, is that normal and acceptable?
image

alixander: multiple errors is normal, but in this case, only one of those errors should show up. but that's too disconnected from this code for me to think the proper fix is here

donglixiaoche: so what about don't exclude itself, and change the expErr to 
`d2/testdata/d2compiler/TestCompile/near_bad_connected.d2:3:12: constant near keys cannot be set on connected shapes
d2/testdata/d2compiler/TestCompile/near_bad_connected.d2:5:5: cannot connect objects from within a container, that has near constant set, to objects outside that container`

donglixiaoche: well, indeed, the second error doesn't make sense for this case, what about add a extra judgement when validate edges?

image

e2etests/testdata/regression/unconnected/elk/board.exp.json
gavin-ts: I think we need it since based on the comment, it is necessary to position labels for objects with constant near positions

donglixiaoche: ohh i see, but for now only constant text type is allowed right? and set near to another shape is only for TALA engine.

gavin-ts: oh it looks like it doesn't need to be text type though, the following is valid:

```d2
a.shape: person
a.near: top-left
```
so in this case it would need a labelPosition set, and since its a person it would go to outside bottom center

gavin-ts: Screen Shot 2023-04-04 at 8 29 51 PM

it also looks like the current d2 version has a bug where icon position isn't set correctly

```d2
a.shape: person
a.near: top-left
a.icon: https://icons.terrastruct.com/essentials/142-target.svg
```

d2graph/d2graph.go
alixander: i think you can just get rid of the first error handle, as it'll always be handled by the second case right?

https://github.com/donglixiaoche/d2/blob/7a395ecf7175d1cd386cfe824c2e1daca20f8820/d2compiler/compile.go#L745

and then remove the `&&` you just added.

d2lib/d2.go
alixander: is it? i still see it here

d2layouts/d2near/layout.go
alixander: am i missing something on my end or did you forget a push perhaps? these aren't fixed for me

Screen Shot 2023-04-05 at 11 05 57 AM

alixander (changes requested):

ci/release/changelogs/next.md
```diff
@@ -1,5 +1,7 @@
 #### Features 🚀
 
+- Container with constant key near attribute now can have descendant objects and connections [#1071](https://github.com/terrastruct/d2/pull/1071)
+
```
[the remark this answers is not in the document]
donglixiaoche: done

d2graph/d2graph.go
donglixiaoche: done

d2lib/d2.go
donglixiaoche: done...

ci/release/changelogs/next.md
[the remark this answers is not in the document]
donglixiaoche: done

d2layouts/d2near/layout.go
alixander: oh i see. in the same way that any object inside any container won't affect bounding box? or something different?

d2layouts/d2near/layout.go
```diff
@@ -104,9 +176,38 @@ func WithoutConstantNears(ctx context.Context, g *d2graph.Graph) (nears []*d2gra
 					break
 				}
 			}
+
+			obj.Parent = tempGraph.Root
```
alixander, line 180: i'd prefer this to be moved up next to line 161, where ur setting tempGraph.Root.Children.

d2layouts/d2near/layout.go
```diff
@@ -10,79 +10,139 @@ import (
...
+					subObject.TopLeft.Y += dy
+				}
+				for _, subEdge := range tempGraph.Edges {
+					for _, point := range subEdge.Route {
+						point.X += dx
+						point.Y += dy
+					}
+				}
+
+				g.Edges = append(g.Edges, tempGraph.Edges...)
```
alixander, line 56: why not add objects back the same way? why do the `attachChildren` way?

alixander (changes requested):

d2layouts/d2near/layout.go
donglixiaoche: line 172 still use `obj.Parent`

alixander: ohh i see.

d2layouts/d2near/layout.go
donglixiaoche: it depends on our requirement, for this case, i think keep bounding box the original one should be correct way. consider this case below:
```
x -> y

a: {
  near: top-center
  c
}
b: {
  near: top-right
  c
}
```

if we don't skip the `c` object inside `a`, after append `a` and `c` inside the main diagram, when it's turn to place `b` object, the new bounding box (blue colore) will be larger than the original one(red color), and so `top-right` will be placed higher than `top-center`
image

keep the bounding box the original one when place all the near objects will make more sense here
image

d2layouts/d2near/layout.go
donglixiaoche: my bad, deleted `attachChildren` should be ok, `tempGraph` contain all the objects, no need to do the recursive way, and also i place `g.Edges = append(g.Edges, tempGraph.Edges...)` to second loop, which is now for attach objects and edges to original graph

alixander (approved):
beautiful! 💯 

is this good to go for merging?

donglixiaoche:
> beautiful! 💯
> 
> is this good to go for merging?

ready to merge if you don't have any other considerations

alixander:
nope, that's what the approval means =). i just started taking a habit of confirming with the author after ran into an instance where i merged too early (https://github.com/terrastruct/d2/pull/1091#issuecomment-1486201686). Thanks @donglixiaoche

PR #1090, opened 2023-03-25
fix scenarios/steps inheritence

alixander:
fixes a bug where scenarios and steps were not inheriting correctly.

we cannot compile scenarios after compiling the whole map.

we have to compile scenarios as we encounter them, so that further edits reference the inherited fields

## Before
Screen Shot 2023-04-05 at 7 18 20 PM

## After
Screen Shot 2023-04-05 at 7 11 11 PM


alixander:
dug into this a little.

i think you cannot compileScenarios after compiling the whole map.

You have to compile scenarios as you encounter them

d2ir/compile.go
```diff
@@ -128,6 +86,22 @@ func (c *compiler) compileField(dst *Map, kp *d2ast.KeyPath, refctx *RefContext)
 				parent: f,
 			}
 		}
+		switch NodeBoardKind(f) {
+		case BoardScenario:
+			c.overlay(dst.Parent().Parent().(*Map), f)
```
nhooyr, line 91: Why is this `dst.Parent().Parent()`? `f` here would be the root of the board and so shouldn't it be `f.Parent().Parent()`?

d2ir/compile.go
```diff
@@ -128,6 +86,22 @@ func (c *compiler) compileField(dst *Map, kp *d2ast.KeyPath, refctx *RefContext)
...
+		case BoardStep:
+			stepsMap := dst.Parent().Map()
+			for i := range stepsMap.Fields {
+				if stepsMap.Fields[i] == f {
+					if i == 0 {
+						c.overlay(dst.Parent().Parent().(*Map), f)
```
nhooyr, line 97: .

nhooyr (changes requested):
nice

d2ir/compile.go
alixander:
```
...
scenarios: {
  1: {
     ...
  }
}
```

`f` is the field `1`, `scenarios` map is the `dst`. 

so `dst.Parent()` gets the field `scenarios`
and then `dst.Parent().Parent()` gets the map that the field `scenarios` is contained in, which is what we want.

i can change to use `f`, but tests only pass if it's `f.Parent().Parent().Parent()`

nhooyr: Right. The problem with using `dst` is that you don't know what it actually is. You could have `scenarios.x: {}` where now `dst` is the base map.

I think the best solution is `ParentField(ParentField(f)).Map()` to grab the base map.

nhooyr: `d2ir.ParentField()` I mean

alixander: oh i didn't know there were that many helpers. updated to use ParentBoard and stuff

d2ir/compile.go
```diff
@@ -88,13 +88,13 @@ func (c *compiler) compileField(dst *Map, kp *d2ast.KeyPath, refctx *RefContext)
 		}
 		switch NodeBoardKind(f) {
 		case BoardScenario:
-			c.overlay(dst.Parent().Parent().(*Map), f)
+			c.overlay(ParentBoard(f).Map(), f)
```
nhooyr, line 91: lmfao i forgot i even had `ParentBoard`

nhooyr (approved):

PR #1115, opened 2023-03-30
fix builds on 32bit

Vaelatern:
If you put the uint32 type at the top instead of later, 32 bit builds

All credit @ahesford

### DIFF
--- a/ci/release/changelogs/next.md
+++ b/ci/release/changelogs/next.md
@@ -6,5 +6,6 @@
 
 #### Bugfixes ⛑️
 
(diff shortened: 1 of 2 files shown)


nhooyr:
@alixander that would work. I opened https://github.com/terrastruct/d2/issues/1121

alixander:
great. @Vaelatern can you please attach a changelog line that this was a bug fixed? thanks again

https://github.com/terrastruct/d2/blob/master/ci/release/changelogs/next.md

Vaelatern:
Done

Issue #1117, opened 2023-03-31
Feature Request: Make `CapsLock` toggable for elements


alexstoick:
In the `Terminal` theme `CapsLock` is set to true making _all_ labels uppercase. 

Can this either be:
(1) configurable at the theme level? (ie: we can still use `Terminal` but w/o everything being Caps)
(2) toggable at the component level in the `style` section:

```
style {
  caps_lock: true/false
}
```

Issue #285, opened 2022-11-30
add different scalar types


gavin-ts:
https://github.com/terrastruct/d2/blob/7f7977eb8b2f7594e72391f2f068715160835536/d2graph/d2graph.go#L45

If it is already valid, I wonder if we shouldn't have it properly set in `d2graph.Scalar` to use the parsed value instead of `Value` as a string everywhere and requiring to cast it every time we need to use it

_Originally posted by @ejulio-ts in https://github.com/terrastruct/d2/pull/250#discussion_r1036327979_

PR #1118, opened 2023-03-31
feat: Add `text-transform` attribute to the `style` block

alexstoick:
This provides an option to override the functionality provided by the
theme, or apply `text-transform` outside of the theme.

The functionality of the `text-transform` is as below:

- `text-transform: none` - will disable **any** transformation (like the uppercasing by `terminal` theme)
- `text-transform: uppercase` (uppercase not upper as per your message) - will force all characters into uppercase.
- `text-transform: lowercase` - will force all characters into lowercase.
- `text-transform: capitalize` - will uppercase the first letter of every word

In addition, this commit introduces:
- helper methods on the `d2graph.Style` struct to determine the type of
  `text-transform` to be applied.
- `ApplyTextTransform` method on the `d2graph.Attributes` which will
  transform the `Label.Value` to the correct text case.

## Screenshots pending.

Fixes: https://github.com/terrastruct/d2/issues/1117


alixander (changes requested):
amazing, thanks so much for opening a PR for this.

1. your intuition is right about the boolean type, but we should make all of them typed in a refactor and test rigorously, vs having a temporary one-off type in the list of styles. (there's an issue for this: https://github.com/terrastruct/d2/issues/285). could you use the same `Scalar` type here please?
2. Tests: https://github.com/terrastruct/d2/blob/a53eb48676772409e0f3f85e3eac130263278f48/e2etests/themes_test.go#L13

- Add a caps lock = false for an object in there
- Regenerate the test with `TESTDATA_ACCEPT=1 ./ci/test.sh ./e2etests -run TestE2E/themes`
- Check the output SVG to see that caps lock=false worked.

And then do that for another test to test capslock=true works.

alexstoick:
@alixander - updated as per your comment:
- removed `Boolean` class. I kept the helper methods - as I think the duplication would be mad, but if you consider it's not in line with the rest of the style I can remove them 👍 
- added tests & updated description with screenshots.

alixander:
@janydoe that's an interesting point. maybe we literally just do https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform

yeah that's more flexible. i've found myself wanting to do capital casing sometimes, and i think many others would to. So then the way to disable the theme's uppercasing is to do `text-transform: none`

Sorry for the moving goal post @alexstoick , would you mind changing to that?

alexstoick:
@alixander I'm happy to implement the `text-transform` spec. Just to be clear on desired behaviour:

- `text-transform: none` - will disable **any** transformation (like the uppercasing by `terminal` theme)
- `text-transform: uppercase` (uppercase not upper as per your message) - will force all characters into uppercase.
- `text-transform: lowercase` - will force all characters into lowercase.
- `text-transform: capitalize` - will uppercase the first letter of every word (defined by some separator) - I'll need to check if there is a helper method for this, or I might come back to you with a proposal.


If the above sounds good - LMK and I'll update the PR. 

Thank you 🙇

alixander:
@alexstoick sounds good to me!

alexstoick:
@alixander - I've implemented the `text-transform` as requested; before I go ahead and add final touches to description (screenshots, etc) - let me know if this is 🆗 for you. I've changed a few of the tests to cover all of the `text-transform` options.

d2graph/d2graph.go
```diff
@@ -131,6 +133,22 @@ type Attributes struct {
...
+
+	if a.Style.UppercaseTextTransform() {
+		a.Label.Value = strings.ToUpper(a.Label.Value)
+	}
+	if a.Style.LowercaseTextTransform() {
+		a.Label.Value = strings.ToLower(a.Label.Value)
+	}
+	if a.Style.CapitalizeTextTransform() {
+		a.Label.Value = cases.Title(language.Und).String(a.Label.Value)
+	}
```
alixander, line 149: inline the method calls please. no need to jump around for a one-liner used in one place

d2graph/d2graph.go
```diff
@@ -340,6 +375,14 @@ func (s *Style) Apply(key, value string) error {
 			return errors.New(`expected "double-border" to be true or false`)
 		}
 		s.DoubleBorder.Value = value
+	case "text-transform":
+		if s.TextTransform == nil {
+			break
+		}
+		if !go2.Contains(textTransforms, strings.ToLower(value)) {
+			return errors.New(`expected "text-transform" to be one of (none, uppercase, lowercase, capitalize)`)
```
alixander, line 383: use strings.join on textTransforms to emit the list here

d2graph/d2graph.go
```diff
@@ -151,25 +169,42 @@ func (r Reference) InEdge() bool {
...
+	Animated      *Scalar `json:"animated,omitempty"`
+	Bold          *Scalar `json:"bold,omitempty"`
+	Italic        *Scalar `json:"italic,omitempty"`
+	Underline     *Scalar `json:"underline,omitempty"`
+	Filled        *Scalar `json:"filled,omitempty"`
+	DoubleBorder  *Scalar `json:"doubleBorder,omitempty"`
+	TextTransform *Scalar `json:"capsLock,omitempty"`
+}
+
+func (s Style) NoTextTransform() bool {
```
alixander, line 194: subtle, but this should be `NoneTextTransform`. it's a special value set, otherwise can be confused with having no value (e.g. s.TextTransform == nil)

d2graph/style_helper.go
```diff
@@ -0,0 +1,3 @@
+package d2graph
+
+var textTransforms = []string{"none", "uppercase", "lowercase", "capitalize"}
```
alixander, line 3: just define these here plz: https://github.com/terrastruct/d2/blob/2e9c05410ccb9d6c457206d15b55386464d2283d/d2graph/d2graph.go#L1621

alixander (changes requested):
generally looks good to me, let's add tests and get this in, ty @alexstoick !

d2graph/d2graph.go
alixander: or i guess this is moot given the comment about inlining these instead

alexstoick: @alixander -  I'm happy to inline _all_ of them **except this one** - as this is being called from a few places. I can rename this to `NoneTextTransform` to avoid having this `if` clause all over the code. 

Will make life much easier once we come back to refactoring these to not all use `d2graph.Scalar`.

You OK with this?

alexstoick:
@alixander - when you say `let's add tests` - I have already updated some of the tests in `e2e/themes`; are there more tests that you would like to see on specific components, or new tests all together?

d2graph/d2graph.go
alixander: oh i see. yeah that's fine

alixander:
oh, i guess those tests do cover it all, nvm!

d2graph/d2graph.go
```diff
@@ -151,25 +169,30 @@ func (r Reference) InEdge() bool {
...
+	Font          *Scalar `json:"font,omitempty"`
+	FontSize      *Scalar `json:"fontSize,omitempty"`
+	FontColor     *Scalar `json:"fontColor,omitempty"`
+	Animated      *Scalar `json:"animated,omitempty"`
+	Bold          *Scalar `json:"bold,omitempty"`
+	Italic        *Scalar `json:"italic,omitempty"`
+	Underline     *Scalar `json:"underline,omitempty"`
+	Filled        *Scalar `json:"filled,omitempty"`
+	DoubleBorder  *Scalar `json:"doubleBorder,omitempty"`
+	TextTransform *Scalar `json:"capsLock,omitempty"`
```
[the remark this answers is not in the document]
alexstoick: eagle eyes - nice one!

alixander (commented):
can you add a changelog here too please?

alixander:
@alexstoick looks like a format error, if you run `./make.sh fmt` it should fix

alexstoick:
@alixander - yep; odd that my formatter didn't auto-fix. Updated now ✅

d2graph/d2graph.go
```diff
@@ -1475,16 +1510,9 @@ func (g *Graph) SetDimensions(mtexts []*d2target.MText, ruler *textmeasure.Ruler
...
-	capsLock := g.Theme != nil && g.Theme.SpecialRules.CapsLock
-
 	for _, obj := range g.Objects {
 		if obj.Attributes.Label.Value != "" {
 			text := obj.Text()
-			if capsLock && !strings.EqualFold(obj.Attributes.Shape.Value, d2target.ShapeCode) {
-				if obj.Attributes.Language != "latex" {
-					text.Text = strings.ToUpper(text.Text)
-				}
-			}
```
alixander, line 1487: how come these are removed?

d2graph/d2graph.go
```diff
@@ -1513,9 +1541,6 @@ func (g *Graph) Texts() []*d2target.MText {
 	for _, edge := range g.Edges {
 		if edge.Attributes.Label.Value != "" {
 			text := edge.Text()
-			if capsLock {
-				text.Text = strings.ToUpper(text.Text)
-			}
```
alixander, line 1518: and this

d2graph/d2graph.go
alexstoick: We are altering the `obj.Attributes.Label.Value` when we call the `ApplyTextTransform` method, which in turn means we no longer need to call `ToUpper` on it here.

alixander: this function is not called in the course of compilation. it's just an exposed API for getting all the texts, so they still have to be applied here.

alixander (changes requested):

d2graph/d2graph.go
alexstoick: Ok. I thought this would be covered in a test; so when I saw all text still looked as expected I thought it was safe to RM.

I will make the required changes to keep the if clause and add the other transformations.

alixander: yeah there probably should be tests for it

d2graph/d2graph.go
```diff
@@ -1512,8 +1548,9 @@ func (g *Graph) Texts() []*d2target.MText {
 	}
 	for _, edge := range g.Edges {
 		if edge.Attributes.Label.Value != "" {
+			edge.Attributes.ApplyTextTransform()
 			text := edge.Text()
-			if capsLock {
+			if capsLock && edge.Attributes.Style.NoneTextTransform() {
```
alixander, line 1553: !NoneTextTransform

alixander (changes requested):

alixander (approved):
tested and works great, thanks @alexstoick

PR #1124, opened 2023-04-02
fix appendix line alignment

### LINKED ISSUE #1104: center align appendix lines
it's slightly off center vertically right now

Screen Shot 2023-03-29 at 2 50 47 PM

to reproduce:

`TESTDATA_ACCEPT=1 ./ci/test.sh ./d2renderers/d2svg/appendix`

segelt:
fixes #1104


alixander (commented):
thanks @segelt ! ohh interesting i didn't realize that the separator wasn't aligned, I was referring to the text. 

Screen Shot 2023-04-02 at 11 41 29 AM

do you want to include that change in here? no worries at all if not, happy to merge this small fix as is.

segelt:
Hi,

I don't think that there was an issue with the separator, its just that right now when the separator is centered exactly from the top of the first circle and the bottom of the rectangle. I can revert those initial changes, it seems that this wasn't an issue.

As for the text itself, I believe adding a small padding would suffice, because inside the `generateNumberedIcon` method the same padding value is used to center the number inside the circle.

segelt:
I have reverted the separator adjustment, in my opinion if its an issue it should have its own issue created.

alixander (changes requested):
sounds good @segelt . Could I trouble you to add a line to the changelog as well? Then it should be good to go, thanks again.

https://github.com/terrastruct/d2/blob/master/ci/release/changelogs/next.md

segelt:
no issues, i have updated the changelog @alixander .

alixander (approved):
🚀

PR #1129, opened 2023-04-03
fix: Truncate Atan2 result to 32-bit precision (Fixes #921)

### LINKED ISSUE #921: fix precision difference for sketch mode running on different architectures
there's something inside here that's causing sketch tests to be different when run on `amd` and `arm` architectures (e.g. M1 macbook and intel macbook). 

To reproduce:

Requires a machine that can run both architectures (I'm on M1 macbook which can run both), or VM environments or such

Run 
`GOARCH=arm64 TESTDATA_ACCEPT=1 ./ci/test.sh ./d2renderers/d2sketch/ -count=1`

Notice nothing changes. (the red asterisk in my terminal means changed files). This is the architecture that our CI runs on.

Run
`GOARCH=amd64 TESTDATA_ACCEPT=1 ./ci/test.sh ./d2renderers/d2sketch/ -count=1`

Notice there are changes!

Screen Shot 2023-04-02 at 3 03 42 PM

I'm not sure if this issue is in rough.js, the JS engine (goja), rounding on architectures, or something in our code reading/writing data from/to rough.js.

ifndev:
This PR is a fix for #921

The divergence between platforms was caused by how Go's math.Atan2() works. It sometimes depends on the result of xatan() ([atan.go line 55](https://go.dev/src/math/atan.go)) which uses floating point arithmetics on constants. Arm and amd work differently when it comes to IEEE floating point numbers, so Go's implementation is only consistent up to a certain precision.

As we don't need 64-bit precision for these operations, we can simply truncate the result of Atan2 in [lib/geo/vector.go](https://github.com/terrastruct/d2/blob/0c9fc134667384e8f3388f3154ecd7230965a0b2/lib/geo/vector.go#L80) -> Radians (line 80). 

We could instead make this change in Arrowheads() ([d2renderers/d2sketch/sketch.go](https://github.com/terrastruct/d2/blob/6d5f2c4f5fcdc4b9aa0086fe5c2a76ff9136d48d/d2renderers/d2sketch/sketch.go#L812) line 812 and 851) if we don't want to change the behaviour of vectors but this is less durable.

/claim #921


gavin-ts (approved):
awesome

we have these calls too which should also be updated: https://github.com/terrastruct/d2/blob/master/lib/shape/shape_oval.go#L38-L62

PR #1140, opened 2023-04-06
Lifeline should inherit the stroke style from the actor

### LINKED ISSUE #1137: Feature Request: Lifeline should inherit Actor border styles
This is relevant only for `sequence diagrams`.

Currently the `lifeline` has a set of hard-coded properties, defined here: https://github.com/terrastruct/d2/blob/master/d2layouts/d2sequence/sequence_diagram.go#L392

The above should be updated so that the `lifeline` style is inherited from the actor's one via the 3 parameters:
- `stroke-dash`
- `stroke-width`
- `stroke` [alias for color]

alexstoick:
Fixes #1137

## Screenshots

image


d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -389,8 +389,9 @@ func (sd *sequenceDiagram) addLifelineEdges() {
 		sd.lifelines = append(sd.lifelines, &d2graph.Edge{
 			Attributes: &d2graph.Attributes{
 				Style: d2graph.Style{
-					StrokeDash:  &d2graph.Scalar{Value: fmt.Sprintf("%d", LIFELINE_STROKE_DASH)},
-					StrokeWidth: &d2graph.Scalar{Value: fmt.Sprintf("%d", LIFELINE_STROKE_WIDTH)},
+					StrokeDash:  actor.Attributes.Style.StrokeDash,  //&d2graph.Scalar{Value: fmt.Sprintf("%d", LIFELINE_STROKE_DASH)},
+					StrokeWidth: actor.Attributes.Style.StrokeWidth, //&d2graph.Scalar{Value: fmt.Sprintf("%d", LIFELINE_STROKE_WIDTH)},
```
gavin-ts, line 393: I think we need to check if each of these was manually set or not. the default values should be used here because nothing was set on the actors:

Screen Shot 2023-04-06 at 2 06 46 PM

alexstoick: @gavin-ts - Yep, I saw the diffs and realised that 👍 Will update!

alexstoick: @gavin-ts - updated now! Diff looks much better now 👍

d2layouts/d2sequence/sequence_diagram.go
```diff
@@ -386,15 +386,24 @@ func (sd *sequenceDiagram) addLifelineEdges() {
 		}
 		actorLifelineEnd := actor.Center()
 		actorLifelineEnd.Y = endY
+		style := d2graph.Style{
+			StrokeDash:  &d2graph.Scalar{Value: fmt.Sprintf("%d", LIFELINE_STROKE_DASH)},
+			StrokeWidth: &d2graph.Scalar{Value: fmt.Sprintf("%d", LIFELINE_STROKE_WIDTH)},
+		}
+		if actor.Attributes.Style.StrokeDash != nil {
+			style.StrokeDash = actor.Attributes.Style.StrokeDash
```
gavin-ts, line 394: I think it might be safer to make a copy of the scalar vs having both point to the same struct, but not sure that it matters here

gavin-ts (approved):
looks good to me

d2layouts/d2sequence/sequence_diagram.go
alexstoick: let me address this quickly!

gavin-ts:
https://github.com/terrastruct/d2/pull/557#issuecomment-1367468730 strikes again

alixander (changes requested):
i'm thinking maybe stroke width should never affect the line. @alexstoick do you think that's something people would actually want? i suspect more often than not, they want to change the actor stroke width without line changes.

alexstoick:
@alixander - Personally, I'm looking for a way to highlight a particular actor in a sequence diagram. Changing the lifeline color (via stroke) - is a feature I'd want.

Another completly different way of tackling this would be to allow the actor to be part of a group - and then it'd be easy to highlight as I could change the group fill.

alixander:
@alexstoick 

> Changing the lifeline color (via stroke)

yeah for sure, but what about stroke width? i say we should have edges inherit a very small set of attributes from its actor.
- stroke
- stroke-dash

alexstoick:
@alixander - sorry; I misread your comment. I think tentatively I will say I'm OK with _not_ inheriting the `stroke-width` - but I can see a scenario where it might be desirable, again for highlighting a particular life line.

alixander:
yeah we can see how well color serves for that purpose, and if there's requests for stroke-width too, we can add it later

alexstoick:
OK - I will remove the `stroke-width` from the PR. Anything else outstanding?

alixander:
nope! looks good to me

alexstoick:
Should be all resolved now :) Thanks for your comments 🙇

alixander:
please do! @alexstoick

alexstoick:
@alixander - `CHANGELOG` updated ✅

alixander (approved):
🚀 🙏

PR #1161, opened 2023-04-09
fix urls in changelogs

janydoe:
You should also change it in the latest release


alixander (approved):
oof, thank you thank you

PR #1139, opened 2023-04-06
GH 821: PowerPoint Export

ejulio-ts:
This fixes #821 

PowerPoint (`.pptx`) files are just a bunch of XML files following the Office Open XML (OOXML) standard and then zipped.
This adds a `pptx` lib that wraps all the logic to create these files around a `Presentation`. There's a base template with the files that don't change and then the slides and some other required metadata are generated on the fly. The XML content was based on the content of an exported presentation. The intent here wasn't to create pptx library.

Important in `pptx/pptx.go` I tested 3 different approaches to generate the XML files.
- long template string (`getRelsSlideXml`)
- string builder with long templates (`getPresentationXmlRels`)
- string builder with a single tag per line (`getAppXml`)

I don't have any strong preference. The first one is quite complicated to update, as finding the template tags can be hard. The last one spans over many lines with long functions. So, open for ideas here and then I can use a single standard after this review. Just made it this way to see if there was some preference to keep.

Tested the resulting `.pptx. file on:
- MS PowerPoint on Mac and Windows
- Open Office on Ubuntu
- KeyNote on Mac
- OneDrive
- Google Drive

ToDo:
- [ ] add links and appendix: this can be a separate PR with just linking, to keep this one small..

Below is the result for
```
mobile -> backend

backend: {
    link: layers.backend
}

mobile: {
    link: layers.mobile
}

layers: {
    backend: {
        api -> mongodb
        worker <-> kafka
        worker -> mongodb

        kafka.shape: queue
        mongodb.shape: cylinder
    }
    mobile: {
        a button
        b button
        c button
        steps: {
            1: {
                user clicks -> a button
            }
            2: {
                a button -> authorized?
                authorized? -> return: no    
            }
            3: {
                authorized? -> save data: yes
            }
        }
    }
}

scenarios: {
    offline: {
        backend.style.opacity: 0.1
        mobile -> local db
        local db.shape: cylinder
        mobile -> backend: try to reconnect every 1s
    }
}
```
[app-flow.pptx](https://github.com/terrastruct/d2/files/11171603/app-flow.pptx)

About the metadata, here's a screenshot

PR #1168, opened 2023-04-09
GH 821: links on pptx

ejulio-ts:
Follow up of https://github.com/terrastruct/d2/pull/1139

This adds links to `.pptx` files.

testing on PowerPoint and KeyNote on Mac

https://user-images.githubusercontent.com/102628724/230991930-f2b7aed3-e0c9-4015-bf8b-7788dc013815.mp4


alixander (changes requested):
is there a way to show mouse cursor for external links?

other than these small things, looks good to go

gavin-ts:
as another issue, I wonder if the appendix should take in a root name so `1: root > mobile` matches the slide title `out / mobile` based on filename `out`?

ejulio-ts:
> as another issue, I wonder if the appendix should take in a root name so 1: root > mobile matches the slide title out / mobile based on filename out?

That's something I was thinking too. Another alternative is to not name the board based on the output file and use `root`

ejulio-ts:
@alixander what do you mean here

> is there a way to show mouse cursor for external links?

Keep the standard cursor instead of pointer for external links?

alixander:
Screen Shot 2023-04-11 at 7 11 19 AM

no i want all links to be cursor pointer. here, the mouse doesn't switch to cursor pointer?

ejulio-ts:
It seems to be a bug in PowerPoint. When I click on a slide link and then move the mouse over another link, it loses the pointer. I need to click again to get it back.

https://user-images.githubusercontent.com/102628724/231221021-0aa359a8-5c97-48a7-8737-cb53a229676e.mp4

lib/png/png.go
```diff
@@ -19,6 +19,9 @@ import (
 	"oss.terrastruct.com/util-go/xmain"
 )
 
+// ConvertSVG scales the image by 2x
+const SCALE = 2.
```
alixander, line 23: use it in `png` too

alixander (approved):

Issue #1170, opened 2023-04-10
elk self edges need to have enough space for labels


alixander:
<img width="670" alt="Screen Shot 2023-04-09 at 5 19 16 PM" src="https://user-images.githubusercontent.com/3120367/230803441-86f351f6-5080-4ad4-aa64-1b0ca2b70650.png">

```
# Generated by ent2d2 (https://github.com/tmc/ent2d2).

# Edges
User <-> User: "spouse" {
  # o2o optional
  source-arrowhead: {
    shape: cf-one-required
  }
  target-arrowhead: {
    shape: cf-one
  }
}
User <-> User: "children/parent" {
  # o2m optional
  source-arrowhead: {
    shape: cf-one-required
  }
  target-arrowhead: {
    shape: cf-many
  }
}
User <-> Pet: "pets/owner" {
  # o2m optional
  source-arrowhead: {
    shape: cf-one-required
  }
  target-arrowhead: {
    shape: cf-many
  }
}
User <-> Card: "card/owner" {
  # o2o optional
  source-arrowhead: {
    shape: cf-one-required
  }
  target-arrowhead: {
    shape: cf-one
  }
}
User <-> Post: "posts/author" {
  # o2m optional
  source-arrowhead: {
    shape: cf-one-required
  }
  target-arrowhead: {
    shape: cf-many
  }
}
User <-> Metadata: "metadata/user" {
  # o2m optional
  source-arrowhead: {
    shape: cf-one-required
  }
  target-arrowhead: {
    shape: cf-many
  }
}
User <-> Info: "info/user" {
  # o2m optional
  source-arrowhead: {
    shape: cf-one-required
  }
  target-arrowhead: {
    shape: cf-many
  }
}

# Tables
Card: {
  shape: sql_table
  id: int {constraint: primary_key}
  owner_id: int {constraint: foreign_key}
  # this is a comment
}
Info: {
  shape: sql_table
  id: int {constraint: primary_key}
  content: json.RawMessage
  # this is a comment
}
Metadata: {
  shape: sql_table
  id: int {constraint: primary_key}
  age: int
  # this is a comment
}
Pet: {
  shape: sql_table
  id: int {constraint: primary_key}
  owner_id: int {constraint: foreign_key}
  # this is a comment
}
Post: {
  shape: sql_table
  id: int {constraint: primary_key}
  text: string
  # this is a comment
  author_id: int {constraint: foreign_key}
  # this is a comment
}
User: {
  shape: sql_table
[...]

PR #1190, opened 2023-04-13
e2etests: Add ent2d2 test case

tmc:
👋 I have a newer project called https://github.com/tmc/ent2d2 and my test render doesn't show up all that well (as filed in #1170).

This is a small PR to add the current output from my project on a test schema as an e2etest.


alixander (approved):
good test case, much appreciated

Issue #793, opened 2023-02-11
render: consider arrowheads for label positions


alixander:
<img width="439" alt="Screen Shot 2023-02-10 at 9 23 14 PM" src="https://user-images.githubusercontent.com/3120367/218241871-fc692f90-7aff-4ee8-8eeb-05c45331df75.png">

donglixiaoche:
@alixander  may i take a look for this one? should be an easy one

alixander:
@donglixiaoche absolutely!

PR #1192, opened 2023-04-13
Consider arrowhead for label position

donglixiaoche:
#793 

consider testcase `arrowhead_labels`, before this pr, the result svg will be
image
because we didn't consider arrowhead for labelPosition, the position will be a little be downward along the rendering path

after this pr, `srcAdj`, `dstAdj`, `srcArrowhead`, `dstArrowhead` will have an effect on labelPostion
image


donglixiaoche:
@alixander the logic of this pr focus on `connection.GetLabelTopLeft`

previously, we got `labelTL` base on `connection.Route`, but the real rendering path of connection may diff a little bit with `connection.Route`, due to the exist of `stroke-width`(of both edge and shape). and for now, `connection.LabelPosition` will always be `INSIDE_MIDDLE_CENTER` (am i right? do i miss something?) , so `stroke-width` won't effect the way we calc labelPosition

in this pr, we need to consider `stroke-width` and `arrowhead` for labelPosition, i think there may be two ways doing this:
1. construct the real rendering route before we calc the labelPosition (give this up, more complicated than i thought)
2. calc the real length we need walk through along `connection.Route`

what do you think?

alixander:
> connection.LabelPosition will always be INSIDE_MIDDLE_CENTER (am i right? do i miss something?)

no, only for dagre and elk. TALA places it anywhere along the connection.

e.g. notice "turn into target" is closer to the destination

https://play.d2lang.com/?script=fJIxjoQwDEV7n-JfYCm2TLHNVnsMlA0QDSSRk4xAA3cfOSESEtJUfPnZ_l841oWcFP7kgxcBUbMNSWH__yYAWPH1g42AnQDt3WDHqPBbROZeXkaZA9JkFlPUoCAnLHruN5-rjA-T9ETAQQdRce6qnXiEnqPhs3wadWWnULMGz-nGB2HaL8HON1adPzTUPNLARn6QYaKa4jqkmqJWueZRSJkdrEse9UjU0HWxOhW1gsD4HOkdAAD__w%3D%3D&layout=tala&sketch=1&theme=302&

donglixiaoche:
ohh, i see. don't know how TALA is integrated. i used to assume all my work is for dagre and elk engine. so i'd better to try construct target route and calc labelPosition base on that right?

alixander:
yeah actually i don't think you'd need to do anything for TALA. will take a look at this PR shortly

d2renderers/d2svg/d2svg.go
```diff
@@ -549,9 +549,18 @@ func drawConnection(writer io.Writer, labelMaskID string, connection d2target.Co
...
+
 	var labelTL *geo.Point
 	if connection.Label != "" {
-		labelTL = connection.GetLabelTopLeft()
+		srcArrowWidth, _ := arrowheadDimensions(connection.SrcArrow, float64(connection.StrokeWidth))
+		dstArrowWidth, _ := arrowheadDimensions(connection.DstArrow, float64(connection.StrokeWidth))
+
+		srcAdjDistance := geo.EuclideanDistance(srcAdj.X, srcAdj.Y, 0, 0) + srcArrowWidth
+		dstAdjDistance := geo.EuclideanDistance(dstAdj.X, dstAdj.Y, 0, 0) + dstArrowWidth
+		labelTL = connection.GetLabelTopLeft(srcAdjDistance, dstAdjDistance)
```
gavin-ts, line 563: instead of updating `GetLabelTopLeft`, and `GetPointOnRoute`, I think it might be better to temporarily move the route start and end points to begin after the arrowheads (e.g. `connection.Route[0].X += srcAdj.X`) and then revert this after getting the label position on the route between arrowheads. This avoids the extra complexity in the `GetPointOnRoute` funcs.

alixander:
also @donglixiaoche just a heads up that @gavin-ts is working on a cousin issue of this, which is adjusting arrowhead labels (different from yours, but related). might be worth taking a look at his to see potential merge conflicts that i think will arise: https://github.com/terrastruct/d2/pull/1207

d2renderers/d2svg/d2svg.go
donglixiaoche: to temporarily construct the route, we can't just add a start and end point to original route, but also we need to remove  the extra points before and after, can this be done by calc the distance along the route? what if the connection is a curve?

donglixiaoche:
> also @donglixiaoche just a heads up that @gavin-ts is working on a cousin issue of this, which is adjusting arrowhead labels (different from yours, but related). might be worth taking a look at his to see potential merge conflicts that i think will arise: #1207

copy, will try to figure out the a proper way to do it

d2renderers/d2svg/d2svg.go
donglixiaoche: what if there are multiple points between original point and new point? if so there would still be issue

gavin-ts: that's true but there shouldn't be multiple points within the length of the arrowhead. It will be more robust to account for that though in case it does come up.

alixander:
@donglixiaoche sorry for the many merge conflicts here. whenever you want to move this forward and resolve those, feel free to ping me to take another look

PR #1200, opened 2023-04-14
GH 1151: export GIF

### LINKED ISSUE #1151: gif export
https://gist.github.com/nitoyon/10108182cc0c12f54878

ejulio-ts:
Fixes #1151 

This allows diagrams to be exported as animated GIFs.

There is a minor refactor to introduce `exportExtension` in order to avoid having many file extensions around the code. It is the first commit and then further commits have minor changes to the initial idea.

The overall export piece needs some refactoring, as the PNG code is duplicated in all formats (pptx, png, pdf, gif). Similarly, the sequence in which we export multiple boards. As it seemed a bigger change, I left it out for now and may revisit it later.

The basic idea is:
- Export all boards as PNGs
- Convert PNGs to Bitmaps (GIF frames) with 255 colors (this is the reason I had to add a new library)
- Make all GIF frames the same size (largest width and height of all boards)
- Center the diagrams in the frame and keep a white background
- Export the animation with an infinite loop

Note that this can "slow" as there are many image manipulations going on and it depends on the image size. I considered using goroutines here, but left out for simplicity at first. The examples below took between 5s-8s to compile.

8s
animated

5s
animated-2


lib/xgif/xgif.go
```diff
@@ -0,0 +1,133 @@
...
+		if err != nil {
+			return nil, err
+		}
+		gifImg, err := gif.Decode(buf)
+		if err != nil {
+			return nil, err
+		}
+		palettedImg, ok := gifImg.(*image.Paletted)
+		if !ok {
+			return nil, fmt.Errorf("decoded git image could not be cast as *image.Paletted")
```
gavin-ts, line 68:
```suggestion
			return nil, fmt.Errorf("decoded gif image could not be cast as *image.Paletted")
```

lib/xgif/xgif.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		// 2. make GIF frames of the same size, keeping images centered and with a white background
```
gavin-ts, line 71: its probably faster to do as a SVG post processing step before converting everything to PNG

ejulio-ts: Good idea. Though, this isn't the part that makes it slow.
Of course, this could save some seconds for large exports, but in the examples I shared they were < 1s.
The costly part is converting the PNG to Bitmap
Quantization is quite expensive and then, later on, mapping from the PNG color to the quantized version is also expensive. Basically, ~WxHxC where C is the number of colors (255)

alixander (changes requested):
amazing that it works, hell yeah.

Can you add a periodic message that it's still running?
https://github.com/ejulio-ts/d2/blob/d1bb3cc575a9f6f1bc5d00067ff8e39a61a2e5de/lib/png/png.go#L93

lib/xgif/xgif_test.go
```diff
@@ -0,0 +1,35 @@
...
+var test_output []byte
+
+func TestPngToGif(t *testing.T) {
+	boards := [][]byte{test_input1, test_input2}
+	interval := 1_000
+	gifBytes, err := AnimatePNGs(boards, interval)
+	assert.NoError(t, err)
+
+	// use this to update the test output
+	if false {
```
alixander, line 27: shouldn't this just skip in ci only?

d2cli/export.go
```diff
@@ -0,0 +1,38 @@
...
+	"path/filepath"
+)
+
+type exportExtension string
+
+const GIF = ".gif"
+const PNG = ".png"
+const PPTX = ".pptx"
+const PDF = ".pdf"
+const SVG = ".svg"
```
alixander, line 13: why don't you make these `exportExtension` type?

d2cli/export.go
```diff
@@ -0,0 +1,38 @@
...
+
+var KNOWN_EXTENSIONS = []string{SVG, PNG, PDF, PPTX, GIF}
```
alixander, line 15: SUPPORTED_EXTENSIONS

alixander, line 15: this reminds me, we should have an error message for if users try to export to `.ppt`: "D2 does not support ppt exports, did you mean "pptx?"

can you make that an issue

d2cli/main.go
```diff
@@ -186,13 +187,13 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
...
 	if outputPath != "-" {
 		outputPath = ms.AbsPath(outputPath)
-		if *animateIntervalFlag > 0 {
-			// Not checking for extension == "svg", because users may want to write SVG data to a non-svg-extension file
-			if filepath.Ext(outputPath) == ".png" || filepath.Ext(outputPath) == ".pdf" || filepath.Ext(outputPath) == ".pptx" {
-				return xmain.UsageErrorf("-animate-interval can only be used when exporting to SVG.\nYou provided: %s", filepath.Ext(outputPath))
-			}
+		if *animateIntervalFlag > 0 && !outputFormat.supportsAnimation() {
+			return xmain.UsageErrorf("-animate-interval can only be used when exporting to SVG or GIF.\nYou provided: %s", filepath.Ext(outputPath))
+		} else if *animateIntervalFlag <= 0 && outputFormat == GIF {
```
alixander, line 195: make another method, `requiresAnimation()`?

alixander (changes requested):

lib/xgif/xgif_test.go
ejulio-ts: my intent here is just to leave the code ready so that, if someone needs to view the generated output, just changes to `true`

d2cli/main.go
```diff
@@ -400,7 +424,7 @@ func compile(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, rende
 		var out []byte
 		if len(boards) > 0 {
 			out = boards[0]
-			if animateInterval > 0 {
+			if animateInterval > 0 && ext != GIF {
```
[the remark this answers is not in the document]
ejulio-ts: Residual. My first attempt was to reuse the same routine for SVG, but it adds a bunch of other things that aren't required..

d2cli/export.go
ejulio-ts: Already solved here, no need for a new issue

d2cli/export.go
```diff
@@ -0,0 +1,46 @@
...
+
+const GIF exportExtension = ".gif"
+const PNG exportExtension = ".png"
+const PPTX exportExtension = ".pptx"
+const PDF exportExtension = ".pdf"
+const SVG exportExtension = ".svg"
+
+var SUPPORTED_EXTENSIONS = []exportExtension{SVG, PNG, PDF, PPTX, GIF}
+
+func getExportExtension(outputPath string) (exportExtension, error) {
```
alixander, line 18: this is the wrong place to add the check, which is apparent from how often you ignore the errors.

just add a `if ex == ppt, log and return` somewhere in main

alixander (changes requested):

d2cli/main.go
```diff
@@ -186,13 +187,16 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
...
 		outputPath = ms.AbsPath(outputPath)
-		if *animateIntervalFlag > 0 {
-			// Not checking for extension == "svg", because users may want to write SVG data to a non-svg-extension file
-			if filepath.Ext(outputPath) == ".png" || filepath.Ext(outputPath) == ".pdf" || filepath.Ext(outputPath) == ".pptx" {
-				return xmain.UsageErrorf("-animate-interval can only be used when exporting to SVG.\nYou provided: %s", filepath.Ext(outputPath))
-			}
+		if *animateIntervalFlag > 0 && !outputFormat.supportsAnimation() {
+			return xmain.UsageErrorf("-animate-interval can only be used when exporting to SVG or GIF.\nYou provided: %s", filepath.Ext(outputPath))
+		} else if *animateIntervalFlag <= 0 && outputFormat.requiresAnimationInterval() {
+			return xmain.UsageErrorf("-animate-interval must be greater than 0 for GIF outputs.\nYou provided: %d", *animateIntervalFlag)
```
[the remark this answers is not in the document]
ejulio-ts: `outputFormat` is `.gif`, not `GIF`.
Also, it breaks the format of the message above

d2cli/main.go
```diff
@@ -186,13 +187,16 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
 			inputPath = filepath.Join(inputPath, "index.d2")
 		}
 	}
+	outputFormat := getExportExtension(outputPath)
+	if outputFormat == PPT {
```
alixander, line 191: ppt is not supported, it's wrong to shoehorn it into that list of supported extensions.

just do
`filepath.Ext == .ppt`

alixander (changes requested):

alixander (approved):

PR #1204, opened 2023-04-14
GH 1199: navbar links

### LINKED ISSUE #1199: navbar for pdf/ppt should be clickable
Screen Shot 2023-04-13 at 8 36 24 PM

ejulio-ts:
Fix #1199 

This makes the title (navbar) of PDFs and PPTXs navigable.

I'm not sure if I should add an underline in the PDF as there's already the horizontal ruler separating the content. Since PPTXs don't have it, the underline seemed ok.

Here are some examples

[animated.pdf](https://github.com/terrastruct/d2/files/11237508/animated.pdf)

[animated.pptx](https://github.com/terrastruct/d2/files/11237510/animated.pptx)


d2cli/main.go
```diff
@@ -33,6 +33,7 @@ import (
 	"oss.terrastruct.com/d2/lib/background"
 	"oss.terrastruct.com/d2/lib/imgbundler"
 	ctxlog "oss.terrastruct.com/d2/lib/log"
+	"oss.terrastruct.com/d2/lib/pdf"
 	pdflib "oss.terrastruct.com/d2/lib/pdf"
```
alixander, line 37: import once, don't rename it

d2cli/main.go
```diff
@@ -749,19 +757,19 @@ func renderPDF(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, opt
 	}
 
 	for _, dl := range diagram.Layers {
-		_, err := renderPDF(ctx, ms, plugin, opts, "", page, ruler, dl, pdf, currBoardPath, pageMap)
+		_, err := renderPDF(ctx, ms, plugin, opts, "", page, ruler, dl, pdf, currBoardPath, LAYERS, pageMap)
```
alixander, line 760: why not just modify `currBoardPath` here instead of passing along another parm

alixander (changes requested):

alixander (commented):
sweet, tested and works

d2cli/main.go
ejulio-ts: It's not so clean.
If I do it here I'd need to extend the slice outside the loop if there are layers/scenarios/steps, and then only change the last element of the slice, otherwise I'd be concatenating sibling boards instead of descendants.
Also, it would require to move the `if diagram.Name == "" {path = root}` outside this function, requiring 2 places to keep in sync.

d2cli/main.go
```diff
@@ -684,19 +684,26 @@ func _render(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, opts
 	return svg, nil
 }
 
-func renderPDF(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, opts d2svg.RenderOpts, outputPath string, page playwright.Page, ruler *textmeasure.Ruler, diagram *d2target.Diagram, pdf *pdflib.GoFPDF, boardPath []string, pageMap map[string]int) (svg []byte, err error) {
+func renderPDF(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, opts d2svg.RenderOpts, outputPath string, page playwright.Page, ruler *textmeasure.Ruler, diagram *d2target.Diagram, pdfDoc *pdf.GoFPDF, boardPath []pdf.BoardTitle, boardType string, pageMap map[string]int) (svg []byte, err error) {
```
alixander, line 687: just `doc`

d2cli/main.go
alixander: can you do a commit for me to see? not convinced

ejulio-ts: see the last commit

d2cli/main.go
```diff
@@ -749,26 +737,46 @@ func renderPDF(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, opt
...
 		if err != nil {
 			return svg, err
 		}
 	}
 
+	var prevBoardID string
+	var lastIndex int
+	if len(diagram.Layers) > 0 || len(diagram.Scenarios) > 0 || len(diagram.Steps) > 0 {
+		lastIndex = len(boardPath)
+		boardPath = append(boardPath, pdf.BoardTitle{})
```
alixander, line 750: why are you doing this? 

in the loops below, why not just append a new board title? 

unless i'm missing something, this code is only confusing because you've made it that way. if you do it right it should be a lot clearer

alixander (changes requested):

d2cli/main.go
ejulio-ts:
```
boardPath = [root]
layers = [layer1, layer2]

# first iter
boardPath = append(boardPath, dl)
# results in boardPath = [root, layer1]
renderPDF()

# second iter
boardPath = append(boardPath, dl)
# results in boardPath = [root, layer1, layer2], wrong path, should be [root, layer2]
renderPDF()
```

The same problem happens for scenarios and steps
After appending the layers, boardPath would be `[root, layers, ...]` and then scenarios would append there instead of having `[root, scenarios]`

alixander: use `:=` then?

ejulio-ts: yeah, I updated it..
Still, the below can happen

```
var boardPath []string

boardPath = append(boardPath, "a")
boardPath = append(boardPath, "b")
boardPath = append(boardPath, "c")

boardPath2 := append(boardPath, "d")
boardPath3 := append(boardPath, "e")

fmt.Printf("%v\n", boardPath)
fmt.Printf("%v\n", boardPath2)
fmt.Printf("%v\n", boardPath3)

// expected
// [a b c]
// [a b c d]
// [a b c e]
// actual
// [a b c]
// [a b c e]
// [a b c e]
```

d2cli/main.go
```diff
@@ -749,26 +737,38 @@ func renderPDF(ctx context.Context, ms *xmain.State, plugin d2plugin.Plugin, opt
...
+		_, err := renderPDF(ctx, ms, plugin, opts, "", page, ruler, dl, doc, path, pageMap)
 		if err != nil {
 			return nil, err
 		}
 	}
 	for _, dl := range diagram.Steps {
-		_, err := renderPDF(ctx, ms, plugin, opts, "", page, ruler, dl, doc, currBoardPath, STEPS, pageMap)
+		path := append(boardPath, pdf.BoardTitle{
+			Name:    dl.Name,
+			BoardID: strings.Join([]string{boardPath[len(boardPath)-1].BoardID, STEPS, dl.Name}, "."),
```
alixander, line 769: this still smells. how about boardPath is a slice of `BoardTitle`, and BoardTitle has
- Name
- Type

and that's it. the boardID can be derived when it's used.

alixander (changes requested):

d2cli/main.go
ejulio-ts: It can, given that we have special rules for `root` since its ID is just `root` and not `type.root`.
Then, I wonder if it makes sense for `pdf.BoardTilte` to know the logic of `BoardIDs` built by `buildBoardIDToIndex`. Assuming `pdf` is a separate package, it doesn't make much to have this heavily coupled assumption about the IDs in it.

alixander: oh i see. okay, don't have to do that then. can u get rid of boardtype in ppt funcs too though

alixander (changes requested):

alixander (approved):

Issue #10, opened 2022-10-19
cli: Watcher client favicon should be red on compile errors


alixander:
like how github has (it's yellow here but it's red when there's an error)

<img width="136" alt="Screen Shot 2023-01-02 at 1 45 55 PM" src="https://user-images.githubusercontent.com/3120367/210280029-8398e0f1-5037-4373-9a3f-e6b1df02b2b8.png">

sinyo-matu:
Hi d2 authors.
I want to contribute to this issue.
And I am already done with my work.
You can check it at [add favicon indicator](https://github.com/sinyo-matu/d2/commit/f9dc6e7595428a2da0e8d4f1e7447f519aadbf61).
Is it Ok to open a PR for this issue?

alixander:
oh wow, please do! thanks @sinyo-matu

sinyo-matu:
@alixander Thanks!
Here it is #1240

Nepo26:
Shouldn't this issue be closed @alixander as it was already merged?

alixander:
good point!

PR #1240, opened 2023-04-22
add favicon error indicator

sinyo-matu:
Fixing #10

- add `favicon.ico` and `favicon-err.ico` to cli watch server resource.
- add code to handle changing favicons depending on Websocket messages.

Here are some examples
- without error
image

- with error
image


d2cli/static/watch.js
```diff
@@ -73,3 +74,18 @@ function init(reconnectDelay) {
...
   };
 }
+
+const changeFavicon = function (iconURL) {
+  const head = document.getElementsByTagName("head")[0];
+  const newLink = document.createElement("link");
+  newLink.rel = "icon";
+  newLink.href = iconURL;
+  removeExistingFavicons();
+  head.appendChild(newLink);
```
alixander, line 84: do you have o remove then append? can you not just get the existing element and set the `href` attribute?

alixander (changes requested):
thanks for this @sinyo-matu !

would you mind signing your commits please? https://github.com/terrastruct/d2/blob/master/docs/CONTRIBUTING.md#logistics

d2cli/static/watch.js
sinyo-matu: @alixander sorry for the delay.
I totally agree with you.
I updated my code.
Please recheck it when you got time.

d2cli/watch.go
```diff
@@ -429,6 +429,7 @@ func (w *watcher) handleRoot(hw http.ResponseWriter, r *http.Request) {
 	<title>%s</title>
 	<script src="./static/watch.js"></script>
 	<link rel="stylesheet" href="./static/watch.css">
+	<link id="favicon-link" rel="icon" href="./static/favicon.ico">
```
alixander, line 432: just "favicon" for the id please

alixander (changes requested):
minor comment then good to go!

d2cli/watch.go
sinyo-matu: Fixed!

alixander (approved):
🚀

PR #1245, opened 2023-04-25
sql_table: support multiple constraints via arrays

satoqz:
## Description

This PR adds a `primary_key_foreign_key` option to the `constraint` field of the `sql_table`, rendered as "PK FK". Useful in the rare yet valid case that a column is both a primary and a foreign key.

Looking for pointers on what/where to add/modify test cases on this.

If this addition is welcomed I'd also go add this to the documentation.

## Example

```d2
test: {
  shape: sql_table
  abc: INT {constraint: primary_key_foreign_key}
  def: INT {constraint: foreign_key}
  ghi: VARCHAR(255)
}
```

test


alixander (changes requested):
thank you @satoqz !

i think the proper way to do this is to have "constraint" keyword accept an array, as well as a string. 

arrays are implemented in the parser but currently unused in the compiler for anything, so it might be a bit tricky to figure out what to do. you're welcome to try it though. you can start by adding something like this in `compiler_test.go`, and hacking on `compiler.go` until it works.

```
x: {
  shape: sql_table
  a: int {constraint: primary_key}
  b: int {constraint: [primary_key, foreign_key]}
}
```

satoqz:
I definitely agree @alixander, especially since this would enable other niche constraint combinations such as foreign key + unique.

I'm up for the challenge and will see what I can do once there's some time available :)

alixander:
i just implemented using arrays for `class` keyword, so there's an example now! https://github.com/terrastruct/d2/pull/1256/files

satoqz:
@alixander I've updated the compiler and related parts of the code and everything seems to work as wanted, see example at the bottom of this comment.

A couple things of notice:

In below code, the `else if` statement would never execute to begin with since `f.Primary() != nil` is already a requirement to reach the switch statement.
https://github.com/terrastruct/d2/blob/6d2613effbf3432b7353a021078f851aea95133b/d2compiler/compile.go#L479-L490
Therefore I've moved the handling of composite fields entirely to the top of the function, which covers class composites and constraint composites. I'm not sure how I feel about the resulting structure, it should probably be cleaned up at some point.

This also results in class arrays now being picked up by the end-to-end tests correctly.

One thing left to do is to update the test data, I've reviewed the resulting diffs and they look good to me but you'll probably want to take a look yourself as updating them via `TA=1` results in ~400 changed files which I didn't want to include in this PR just yet :smile:

Example:

```d2
test: {
  shape: sql_table
  a: INT
  b: INT {constraint: foreign_key}
  c: INT {constraint: [foreign_key; primary_key]}
  d: VARCHAR(255) {constraint: [foreign_key; unique]}
}
```

test

d2compiler/compile.go
```diff
@@ -307,9 +307,28 @@ func (c *compiler) compileLabel(attrs *d2graph.Attributes, f d2ir.Node) {
 
 func (c *compiler) compileReserved(attrs *d2graph.Attributes, f *d2ir.Field) {
 	if f.Primary() == nil {
-		if f.Composite != nil && !strings.EqualFold(f.Name, "class") {
+		compositeAccepted := strings.EqualFold(f.Name, "class") || strings.EqualFold(f.Name, "constraint")
+		if f.Composite != nil && !compositeAccepted {
```
alixander, line 311: composite might still be nil below.

the switch statement should go inside `if f.Composite != nil`, and then the default case is the error

d2compiler/compile_test.go
```diff
@@ -2139,13 +2139,17 @@ ok: {
...
 		},
 		{
-			name: "sql-panic",
-			text: `test {
-    shape: sql_table
-    test_id: varchar(64) {constraint: [primary_key, foreign_key]}
-}
-`,
-			expErr: `d2/testdata/d2compiler/TestCompile/sql-panic.d2:3:27: reserved field constraint does not accept composite`,
+			name: "sql-constraints",
```
alixander, line 2142: nice test!

d2graph/d2graph.go
```diff
@@ -1036,9 +1036,9 @@ func (obj *Object) GetDefaultSize(mtexts []*d2target.MText, ruler *textmeasure.R
 			}
 			maxTypeWidth = go2.Max(maxTypeWidth, typeDims.Width)
 
-			if c.Constraint != "" {
-				// covers UNQ constraint with padding
-				constraintWidth = 60
+			// 60 covers UNQ constraint with padding, 50 for further constraints covers UNQ + space
```
alixander, line 1039: can you add some e2e tests (just put it in stable_test) to see how this in action?

alixander (changes requested):
awesome! just a few comments and tests to see what it looks like, thanks!

alixander:
i wonder if we should add some kind of visual separator to multiple constraints. maybe like a `|`? @katwangy wdyt

satoqz:
The formula to calculate the width is obviously not perfect, but should look good for "valid" use cases up to three constraints. (i.e. assuming users don't set a field to "UNQ UNQ UNQ UNQ" or similar for some reason :smile:)

d2target/sqltable.go
```diff
@@ -41,14 +43,22 @@ func (c SQLColumn) Texts(fontSize int) []*MText {
...
+		case "foreign_key":
+			abbr = "FK"
+		case "unique":
+			abbr = "UNQ"
+		}
+
+		abbrs = append(abbrs, abbr)
 	}
+
+	return strings.Join(abbrs, " ")
```
alixander, line 63: what if we join with comma? `, `. i think that could help distinguish multiple better

alixander (changes requested):
@satoqz sorry for the looong delay here. this looks good to me. can you update this PR and run tests with `TA=1` to accept all test changes?

alixander:
oh and include a line in the `changelog.md` please

satoqz:
@alixander All done :)

Preview of what it looks like with commas here: https://github.com/terrastruct/d2/blob/0ee9ce42419d586cb250ce28a23c83abe2b1cb0f/e2etests/testdata/stable/sql_table_constraints_width/dagre/sketch.exp.svg

alixander (approved):
looks great! ty again @satoqz

Issue #1107, opened 2023-03-29
more font optimization: subset by style


alixander:
currently we subset fonts by collecting all characters and encoding, in every style, those characters.

we can further optimize by subsetting each character WITH its style, so that the font encoding for bold only includes the bolded characters, the font encoding for italic only includes the italics, etc. 

related initial optimization: https://github.com/terrastruct/d2/pull/1089

alixander:
@donglixiaoche can you leave a comment here so i can assign you? (github won't let me assign otherwise)

donglixiaoche:
@alixander coooopy, wanna give this issue a shot

donglixiaoche:
@alixander when will a text be `semibold`? there's no such a flag inside `Text` struct indicate it's `semibold`, only `bold` flag

donglixiaoche:
> 

ohh i see, when there is markdown inside diagram, right?

donglixiaoche:
https://github.com/terrastruct/d2/blob/feae388f19b72bb9c90de7452472bfd154bcd86c/d2renderers/d2svg/d2svg.go#L1377-L1392

and why do we need regular font for `md` class? below these lines, we also add semibold font for `md` class

donglixiaoche:
below is the overall idea of this pr:
well in the previous optimization, we collect all the characters, called corpus, inside a diagram, and then subset our font by cutting the unnecssary fonts. what remains after this optimization is, say,if we only have a bold `a` character, we'd still include both regular or other fonts for this `a` character.
so, what we need to do is when collecting characters, we also need their style info, then when we subset our fonts, we can get rid of all unnecessary fonts.

PR #1248, opened 2023-04-26
Subset fonts by fontstyle

donglixiaoche:
#1107


donglixiaoche:
below is the overall idea of this pr:
well in the previous optimization, we collect all the characters, called corpus, inside a diagram, and then subset our font by cutting the unnecssary fonts. what remains after this optimization is, say,if we only have a bold a character, we'd still include both regular or other fonts for this a character.
so, what we need to do is when collecting characters, we also need their style info, then when we subset our fonts, we can get rid of all unnecessary fonts.

donglixiaoche:
@alixander i asked you some questions in the related issue but got no responses... anyway, no worries

i draft this pr, can you help to take a glimpse, so i can know if i'm on the right direction

alixander (changes requested):
this looks like the right approach!

the e2e test deltas have some errors though

(btw to get these, i just run
`git checkout master e2etests/testdata`
and then
`./ci/e2ereport.sh --delta`
)

Screen Shot 2023-04-26 at 4 58 29 PM

Screen Shot 2023-04-26 at 4 58 48 PM

donglixiaoche:
@alixander for this script:
```
costumes: {
  shape: sql_table
  id: int {constraint: primary_key}
}
```

i notice that for the generated diagram, text `consumes` is bold. the text style is added here
https://github.com/terrastruct/d2/blob/b08431b73a91c6a645767df6eafba092e9b70aaf/d2graph/d2graph.go#L571

but inside the output svg, class `text-bold` can't be found, is that expected ? is it a bug?

alixander:
@donglixiaoche ah, yeah i think that is a bug. it just didn't affect anything because Class has its own renderer that ignored the value set on d2target: https://github.com/terrastruct/d2/blob/91beb8d12dddda6dabb09a6fc5cb6d25bac029d8/d2renderers/d2svg/class.go#L14

donglixiaoche:
@alixander but basically we add `bold` font according to `text-bold` class, and after this optimization, if no `text-bold` is found, bold font won't be included. so the bug you mentioned above is caused by this bug

alixander:
yes, i think it was just an innocuous bug until now, but this method will require that bugfix

PR #1356, opened 2023-05-31
fonts: fix subset cmap table length

berniexie:
The CMAP length calculation was incorrect. I recalculate it based on how the original fpdf [calculates it](https://github.com/Setasign/FPDF/blob/f4104a04c9a3f95c4c26a0a0531abebcc980987a/makefont/ttfparser.php#L476) instead. Thanks to @mikeday for pointing us in the right direction.

Export before fix
broken

Rendered on Prince
Screenshot 2023-05-31 at 4 32 02 PM

Export after fix
fixed

Rendered on Prince
Screenshot 2023-05-31 at 4 32 58 PM


lib/font/subsetFont.go
```diff
@@ -0,0 +1,899 @@
...
+		cmap = append(cmap, 0)
+
+	}
+	cmap = append(cmap, 0)
+	for _, start := range cidArrayKeys {
+		cmap = append(cmap, cidArray[start]...)
+	}
+	cmap = append(cmap, 0)
+
+	// Calculating cmap length based off of fpdf https://github.com/Setasign/FPDF/blob/f4104a04c9a3f95c4c26a0a0531abebcc980987a/makefont/ttfparser.php#L476
```
berniexie, line 371: actual changed code is around here

alixander (approved):

PR #1364, opened 2023-06-02
compile: no empty block strings

berniexie:
Throw a compiler error for empty block strings.


d2compiler/compile.go
```diff
@@ -270,6 +270,9 @@ func (c *compiler) compileLabel(attrs *d2graph.Attributes, f d2ir.Node) {
 		// TODO: Delete instead.
 		attrs.Label.Value = scalar.ScalarString()
 	case *d2ast.BlockString:
+		if scalar.ScalarString() == "" {
```
alixander, line 273: use https://pkg.go.dev/strings#TrimSpace too. weird things happen in our app when it's just spaces

alixander (changes requested):

alixander (approved):

PR #1403, opened 2023-06-13
Fix: struct literal uses unkeyed fields

yunginnanet:
### oss.terrastruct.com/d2/d2cli

Fixes: `d2cli/watch.go:406:21: oss.terrastruct.com/util-go/xhttp.HandlerFuncAdapter struct literal uses unkeyed fields`


alixander (approved):
thanks!

PR #1420, opened 2023-06-18
fix nested board AST

alixander:
fixes https://github.com/terrastruct/d2/pull/1414

was still using the root AST instead of the board's


testdata/d2compiler/TestCompile/link-board-key-nested.exp.json
```diff
@@ -308,158 +308,15 @@
                         "range": ",0:0:0-0:0:0",
                         "value": [
                           {
-                            "string": "x"
+                            "string": "yo"
```
alixander, line 311: this test confirms it. 

the test DSL:
```
x: {
  y.link: layers.x
}
layers: {
  x: {
    yo
  }
}
```

this diff shows that only `yo` remains with no value, and `x` from the previous layer doesn't appear in the AST anymore

nhooyr (commented):
✅

Issue #1423, opened 2023-06-20
specify icon and label position


alixander:
reuse `near` keywords, nested.

```
x: hello {
  icon: https://whatever
  icon.near: top-left
  label.near: top-center
}
```

PR #1437, opened 2023-06-22
label and icon position attributes

alixander:
see tests. only compiling, not used

for https://github.com/terrastruct/d2/issues/1423

### DIFF
--- a/d2compiler/compile.go
+++ b/d2compiler/compile.go
@@ -382,6 +382,47 @@ func (c *compiler) compileLabel(attrs *d2graph.Attributes, f d2ir.Node) {
 	attrs.Label.MapKey = f.LastPrimaryKey()
 }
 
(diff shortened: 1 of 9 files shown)


d2compiler/compile.go
```diff
@@ -382,6 +382,47 @@ func (c *compiler) compileLabel(attrs *d2graph.Attributes, f d2ir.Node) {
 	attrs.Label.MapKey = f.LastPrimaryKey()
 }
 
+func (c *compiler) compilePosition(attrs *d2graph.Attributes, f *d2ir.Field) {
+	name := f.Name
+	if f.Map() != nil {
+		for _, f := range f.Map().Fields {
```
gavin-ts, line 388: should we rename outer `f` to `field`?

gavin-ts (approved):

Issue #735, opened 2023-01-28
d2compiler: Null to unset any attribute/value/object/edge


nhooyr:

alixander:
another use case:
(ideally this is not opacity 0, it's null)

```
this
that

this.class: toggle
that.class: toggle

layers: {
  for-clients: {
    classes: {
      toggle: { style.opacity: 0 }
    }
  }
}
```

PR #1446, opened 2023-06-26
Null keyword

alixander:
only for objects and edges. no attribute nulling yet

partial of #735


d2compiler/compile.go
```diff
@@ -836,6 +834,19 @@ func (c *compiler) compileArrowheads(edge *d2graph.Edge, f *d2ir.Field) {
 	}
 }
 
+func (c *compiler) nullify(g *d2graph.Graph) {
```
nhooyr, line 837: I'd prefer to null as it comes in rather then at the end. As, after a null, the object is gone. So if you try to create it again, it shouldn't assume all properties from before the null, only what's added after the null.

nhooyr (changes requested):

d2compiler/compile.go
alixander: i first tried to null in the IR, then at compiler as it reads fields, and finally arrived at doing it at the end.

the implicit deletion is hard to do otherwise:

implicit edge
```d2
a -> b
b: null
```

you can only know if the edge should be deleted at the end, if the final ref to one of the endpoints was null.

implicit children
```d2
a.b.c
a.b: null
```

`a.b.c` only exists if `a.b` wasn't nulled somewhere.

> after a null, the object is gone. So if you try to create it again, it shouldn't assume all properties from before the null, only what's added after the null.

oh interesting. yeah i can see that. i can also see the flip side: ordering correctly is annoying. like it's not obvious D2 would evaluate an inner scope's underscore null setting first. will think on it a bit but may just keep this for now.

alixander: actually, i guess the case can be made that for the implicit edge example, when `b` reinitializes, it's a completely new `b` that doesn't inherit the previous `b`'s connections. likewise with the children

alixander: yeah ur right, that's better.

d2ir/d2ir.go
```diff
@@ -751,7 +765,33 @@ func (m *Map) DeleteField(ida ...string) *Field {
 			continue
 		}
 		if len(rest) == 0 {
+			for _, fr := range f.References {
+				for i, e := range m.Edges {
+					for _, er := range e.References {
+						if er.Context == fr.Context {
+							m.DeleteEdge(e.ID)
+							i--
```
nhooyr, line 773: this doesn't work in range loops. `i` will be set on the next iteration as if `i--` never happened.

d2ir/d2ir.go
```diff
@@ -751,7 +765,33 @@ func (m *Map) DeleteField(ida ...string) *Field {
...
+				}
+			}
 			m.Fields = append(m.Fields[:i], m.Fields[i+1:]...)
+
+			// If a field was deleted from a keyword-holder keyword and that holder is empty,
+			// then that holder becomes meaningless and should be deleted too
+			parent := ParentField(f)
+			for keyword := range d2graph.ReservedKeywordHolders {
+				if parent != nil && parent.Name == keyword && len(parent.Map().Fields) == 0 {
+					styleParentMap := ParentMap(parent)
```
nhooyr, line 786: is it necessarily a `style`?

nhooyr (commented):
looks good except for comments
🔴

d2ir/d2ir.go
alixander: oh right. it wasn't necessary anyway, for the same reason

PR #1473, opened 2023-07-11
Vars

### LINKED ISSUE #105: syntax: variables
E.g. have $vars.light_primary hold the hex code which you reference as background on a bunch of objects

alixander:
closes #105


d2compiler/compile.go
```diff
@@ -277,6 +277,16 @@ func (c *compiler) compileField(obj *d2graph.Object, f *d2ir.Field) {
...
 		}
 		return
+	} else if f.Name == "vars" {
+		if f.Map() != nil {
+			if len(f.Map().Edges) > 0 {
+				c.errorf(f.Map().Edges[0].LastRef().AST(), "vars cannot contain an edge")
+			}
+		}
+		if d2ir.NodeBoardKind(d2ir.ParentMap(f)) == "" {
+			c.errorf(f.LastRef().AST(), "vars must be defined at the root of a board")
```
nhooyr, line 287: Why is this a requirement? I wanted vars to be able to be defined on a per map basis rather than per board.

nhooyr (changes requested):
This isn't implemented how I wanted it.

You_ can't compile vars before everything else. The var's map is a map just like any other D2 map and itself has to be compiled first. Vars should be resolved at the end and be defined on a per map basis, not a per board. Just like how variable scopes work.

If you define a variable in a for loop, it's only available in the for loop. You don't have to define it at the start of your function in order to make it available in your for loop.

d2ir/compile.go
```diff
@@ -92,7 +93,114 @@ func (c *compiler) compileClasses(m *Map) {
...
+func (c *compiler) compileVars(m *Map, ast *d2ast.Map) {
+	for _, n := range ast.Nodes {
+		if n.MapKey != nil && n.MapKey.Key != nil && len(n.MapKey.Key.Path) == 1 && strings.EqualFold(n.MapKey.Key.Path[0].Unbox().ScalarString(), "vars") {
+			c.compileKey(&RefContext{
+				Key:      n.MapKey,
+				Scope:    ast,
+				ScopeMap: m,
+				ScopeAST: ast,
+			})
+			break
```
nhooyr, line 183: You could have more than one vars map in a map, so you can't break here.

nhooyr: Either way it's best you do resolution of variables all at the end so you don't need any of this code.

nhooyr (changes requested):

d2ir/compile.go
```diff
@@ -92,7 +94,130 @@ func (c *compiler) compileClasses(m *Map) {
...
+	return resolved
+}
+
+func (c *compiler) overlayVars(base, overlay *Map) {
+	vars := overlay.GetField("vars")
+	if vars == nil {
+		return
+	}
+
+	lVars := base.GetField("vars")
```
nhooyr, line 211:
```suggestion
	baseVars := base.GetField("vars")
```

not sure what you mean by lVars.

d2ir/compile.go
```diff
@@ -92,7 +94,130 @@ func (c *compiler) compileClasses(m *Map) {
...
+
+	if lVars == nil {
+		lVars = vars.Copy(base).(*Field)
+		base.Fields = append(base.Fields, lVars)
+	} else {
+		overlayed := vars.Copy(base).(*Field)
```
nhooyr, line 217: duplicated code as above, also don't need a new variable here.

d2ir/compile.go
```diff
@@ -92,7 +94,130 @@ func (c *compiler) compileClasses(m *Map) {
...
+		for i, box := range s.Value {
+			if box.Substitution != nil {
+				for _, vars := range varsStack {
+					resolvedField = c.resolveSubstitution(vars, box.Substitution)
+					if resolvedField != nil {
+						break
+					}
+				}
+				if resolvedField != nil {
+					if resolvedField.Composite != nil {
```
nhooyr, line 138: why can't the map variable be both a map and have a primary? That's how the syntax works.
```suggestion
					if resolvedField.Primary == nil {
```

nhooyr: Actually one thing missing here is that composite variables should work. So for unquoted strings that have no other text except for a variable subst, both the primary and composite should be substituted.

d2ir/compile.go
```diff
@@ -92,7 +94,130 @@ func (c *compiler) compileClasses(m *Map) {
...
+						return
+					}
+					// If lone and unquoted, replace with value of sub
+					if len(s.Value) == 1 {
+						scalar.Value = resolvedField.Primary().Value
+					} else {
+						s.Value[i].String = go2.Pointer(resolvedField.Primary().String())
+						subbed = true
+					}
+				} else {
```
nhooyr, line 149: Flip to remove indentation.

d2ir/compile.go
```diff
@@ -92,7 +94,130 @@ func (c *compiler) compileClasses(m *Map) {
...
+					}
+				}
+				if resolvedField != nil {
+					if resolvedField.Composite != nil {
+						c.errorf(node, `cannot reference map variable "%s"`, strings.Join(box.Substitution.IDA(), "."))
+						return
+					}
+					s.Value[i].String = go2.Pointer(resolvedField.Primary().String())
+					subbed = true
+				} else {
```
nhooyr, line 175: flip for less indentation.

d2ir/compile.go
```diff
@@ -92,7 +94,130 @@ func (c *compiler) compileClasses(m *Map) {
...
+					}
+				} else {
+					c.errorf(node, `could not resolve variable "%s"`, strings.Join(box.Substitution.IDA(), "."))
+					return
+				}
+			}
+		}
+		if subbed {
+			s.Coalesce()
+			scalar.Value = d2ast.RawString(s.ScalarString(), false)
```
nhooyr, line 157: Why do you do this here but not below?

d2ir/compile.go
```diff
@@ -92,7 +94,130 @@ func (c *compiler) compileClasses(m *Map) {
...
+			s.Coalesce()
+		}
+	}
+}
+
+func (c *compiler) resolveSubstitution(vars *Map, substitution *d2ast.Substitution) *Field {
+	var resolved *Field
+	for _, p := range substitution.Path {
+		if vars == nil {
+			resolved = nil
```
nhooyr, line 191: just return nil

d2ir/compile.go
```diff
@@ -92,7 +94,130 @@ func (c *compiler) compileClasses(m *Map) {
...
+			break
+		}
+		r := vars.GetField(p.Unbox().ScalarString())
+		if r == nil {
+			resolved = nil
```
nhooyr, line 196: just return nil

d2ir/compile.go
```diff
@@ -92,7 +94,130 @@ func (c *compiler) compileClasses(m *Map) {
...
+			break
+		}
+		vars = r.Map()
+		resolved = r
+	}
+	return resolved
+}
```
nhooyr, line 203: This function can actually be simplified to:

```go
func (c *compiler) resolveSubstitution(vars *Map, substitution *d2ast.Substitution) *Field {
	if vars == nil {
		return nil
	}
	for i, p := range substitution.Path {
		f := vars.GetField(p.Unbox().ScalarString())
		if f == nil {
			return nil
		}

		if i == len(substitution.Path)-1 {
			return f
		}
		vars = f.Map()
	}
	return nil
}
```

nhooyr (changes requested):
Add a test for
```
vars: {
  my-var: ${my-var}
}
```
too

d2ir/compile.go
alixander: oh oops, copied from elsewhere

d2ir/compile.go
alixander: oh interesting. like

```d2
vars: {
  colors: my colors {
    primary: red
  }
}
a: ${colors}
b: ${colors.primary}
```
?

alixander: re composite variables: i don't know. i can't think of a use case. feels like a `classes` use case

nhooyr: It's about completeness here rather than a specific scenario. It should work as it's consistent with the language. Otherwise it's confusing.

But to give a concrete example, maybe you want to create a whole scenario and duplicate it a few times and so you throw it into a variable. That doesn't seem unreasonable to me. And then you variadic substitute it in as `{ ...${vars.scenario}; overrides }`.

So that's another thing that needs to be added. Variadic substitutions. They're implemented in the parser already.

d2ir/compile.go
alixander: my b, this was left in by mistake

d2ir/compile.go
```diff
@@ -92,7 +94,133 @@ func (c *compiler) compileClasses(m *Map) {
...
+				if resolvedField == nil {
+					c.errorf(node, `could not resolve variable "%s"`, strings.Join(box.Substitution.IDA(), "."))
+					return
+				}
+				if resolvedField.Composite != nil {
+					c.errorf(node, `cannot reference map variable "%s"`, strings.Join(box.Substitution.IDA(), "."))
+					return
+				}
+				// If lone and unquoted, replace with value of sub
+				if len(s.Value) == 1 {
```
nhooyr, line 152: Why bother with this?

nhooyr (changes requested):

d2ir/compile.go
alixander: it's for this test: https://github.com/terrastruct/d2/pull/1473/files#diff-ff6179ec34b122cf73328d5524adaf22e6049d777051407b3f9dac1da530cea5R3309

(if link breaks, search for "quoted-var")

the variable is a quoted string, but the substitution is unquoted string. the contents won't match the type

nhooyr: Right but that doesn't matter at the IR stage. I'm fairly certain you could remove this if clause and that test would still pass.

d2ir/compile.go
```diff
@@ -143,14 +143,33 @@ func (c *compiler) resolveSubstitutions(varsStack []*Map, node Node) {
 					c.errorf(node.LastRef().AST(), `could not resolve variable "%s"`, strings.Join(box.Substitution.IDA(), "."))
 					return
 				}
+				if box.Substitution.Spread {
+					if resolvedField.Composite == nil {
+						c.errorf(box.Substitution, "cannot spread non-map into map")
```
nhooyr, line 148: cannot spread non composite.

d2ir/compile.go
```diff
@@ -143,14 +143,33 @@ func (c *compiler) resolveSubstitutions(varsStack []*Map, node Node) {
...
+						continue
+					}
+					// TODO arrays
+					if resolvedField.Map() != nil {
+						OverlayMap(ParentMap(node), resolvedField.Map())
+					}
+					// Remove the placeholder field
+					f := node.(*Field)
```
nhooyr, line 156: Add a DeleteField method if one doesn't already exist.

d2ir/compile.go
```diff
@@ -143,14 +143,33 @@ func (c *compiler) resolveSubstitutions(varsStack []*Map, node Node) {
 					c.errorf(node.LastRef().AST(), `could not resolve variable "%s"`, strings.Join(box.Substitution.IDA(), "."))
 					return
 				}
+				if box.Substitution.Spread {
+					if resolvedField.Composite == nil {
+						c.errorf(box.Substitution, "cannot spread non-map into map")
+						continue
+					}
+					// TODO arrays
```
nhooyr, line 151: Do handle arrays in this PR, shouldn't be much code.

d2ir/compile.go
```diff
@@ -255,6 +274,21 @@ func (c *compiler) compileMap(dst *Map, ast, scopeAST *d2ast.Map) {
 				ScopeMap: dst,
 				ScopeAST: scopeAST,
 			})
+		case n.Substitution != nil:
+			if !n.Substitution.Spread {
+				c.errorf(n.Import, "invalid non-spread substitution in map")
```
nhooyr, line 279: I could be wrong but I believe the parser will already error on this? If not, do add a test.

d2ir/compile.go
```diff
@@ -255,6 +274,21 @@ func (c *compiler) compileMap(dst *Map, ast, scopeAST *d2ast.Map) {
...
+				continue
+			}
+			// placeholder field to be resolved at the end
```
nhooyr, line 282: Seems reasonable to me. Should be the same for arrays except a placeholder `*Scalar`.

nhooyr (changes requested):

d2ir/compile.go
alixander: there is one but it takes an `ida`, which is much more common and useful. this is just a one-off

d2ir/compile.go
alixander: it doesn't, but ur right that it doesn't reach here. added a comment that it should

d2compiler/compile.go
```diff
@@ -281,6 +281,12 @@ func (c *compiler) compileField(obj *d2graph.Object, f *d2ir.Field) {
 		if f.Map() != nil {
 			if len(f.Map().Edges) > 0 {
 				c.errorf(f.Map().Edges[0].LastRef().AST(), "vars cannot contain an edge")
+			} else {
+				for _, f := range f.Map().Fields {
+					if f.Primary() == nil && f.Composite == nil {
```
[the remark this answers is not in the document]
alixander: hmm i think we should error on setting a var to null

d2ir/compile.go
alixander: the test does not if i remove this.

alixander: i suspect if you set unquotedstring to a string with quotes, something interpreting the string just gets rid of it, maybe in formatter. i can look deeper if you're not satisfied

d2ir/compile.go
```diff
@@ -210,18 +210,15 @@ func (c *compiler) overlayVars(base, overlay *Map) {
 	if vars == nil {
 		return
 	}
+	vars = vars.Copy(base).(*Field)
```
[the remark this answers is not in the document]
alixander: wouldn't this test have caught that? https://github.com/terrastruct/d2/pull/1473/commits/301bc8ddfbd789549af58e16add68a80d9ce84cb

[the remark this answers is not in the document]
nhooyr: Thing is you're passing them into the function flipped too lol

d2ir/compile.go
nhooyr: I'll look into it. I don't see how the formatter can be involved.

d2ir/compile.go
nhooyr: Why add a comment? Why doesn't it reach here?

d2compiler/compile.go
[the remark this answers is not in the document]
nhooyr: Why? I could see someone wanting to null composite variables to make them not spread into anything.

d2ir/compile.go
alixander: did you read the comment? it doesn't error but it's quite harmless.

d2compiler/compile.go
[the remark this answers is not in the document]
alixander: should substituting to a nulled variable just give "can't resolve" error?

d2ir/compile.go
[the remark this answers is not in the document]
nhooyr: Looks like `overlayVars` isn't required. The code being incorrect like I described isn't being noticed because even if you comment it out or flip the base/overlay arguments to it, the tests still pass though the testData changes slightly.

I think the varStack is being kept across boards and so you don't need to overlay vars for the parent board's vars to take into effect if the child board doesn't have its own.

d2ir/compile.go
nhooyr: Here's the right code:

```diff
diff --git a/d2ir/compile.go b/d2ir/compile.go
index f4f0e7d4c..6171b4cc2 100644
--- a/d2ir/compile.go
+++ b/d2ir/compile.go
@@ -195,13 +195,8 @@ func (c *compiler) resolveSubstitutions(varsStack []*Map, node Node) {
 						n.Primary_ = nil
 					}
 				} else {
-					if len(s.Value) == 1 {
-						// If lone and unquoted, replace with value of sub
-						node.Primary().Value = resolvedField.Primary().Value
-					} else {
-						s.Value[i].String = go2.Pointer(resolvedField.Primary().String())
-						subbed = true
-					}
+					s.Value[i].String = go2.Pointer(resolvedField.Primary().Value.ScalarString())
+					subbed = true
 				}
 				if resolvedField.Composite != nil {
 					switch n := node.(type) {
@@ -237,7 +232,7 @@ func (c *compiler) resolveSubstitutions(varsStack []*Map, node Node) {
 					c.errorf(node.LastRef().AST(), `cannot substitute map variable "%s" in quotes`, strings.Join(box.Substitution.IDA(), "."))
 					return
 				}
-				s.Value[i].String = go2.Pointer(resolvedField.Primary().String())
+				s.Value[i].String = go2.Pointer(resolvedField.Primary().Value.ScalarString())
 				subbed = true
 			}
 		}
```

Three tests broke but it looks like they were testing for incorrect behaviour.

d2ir/compile.go
nhooyr: Oh I see. I would still say it's unnecessary. If that behavior ever changed, it would be on the parser to error, not the IR as it's invalid syntax.

nhooyr: But I'll defer.

d2compiler/compile.go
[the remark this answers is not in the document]
nhooyr: Argh yea maybe it's best to err on the side of erroring for now instead of silently ignoring.

d2ir/compile.go
```diff
@@ -106,6 +106,13 @@ func (c *compiler) compileSubstitutions(m *Map, varsStack []*Map) {
...
 			c.resolveSubstitutions(varsStack, f)
 		}
+		if arr, ok := f.Composite.(*Array); ok {
+			for _, val := range arr.Values {
+				if scalar, ok := val.(*Scalar); ok {
+					c.resolveSubstitutions(varsStack, scalar)
+				}
+			}
+		}
 		if f.Map() != nil {
```
nhooyr, line 116: Should be else with above.

d2ir/compile.go
```diff
@@ -145,20 +152,34 @@ func (c *compiler) resolveSubstitutions(varsStack []*Map, node Node) {
 				}
 				if box.Substitution.Spread {
 					if resolvedField.Composite == nil {
-						c.errorf(box.Substitution, "cannot spread non-map into map")
+						c.errorf(box.Substitution, "cannot spread non-composite into composite")
```
nhooyr, line 155: "cannot spread non-composite."

d2parser/parse.go
```diff
@@ -1560,7 +1560,8 @@ func (p *parser) parseArrayNode(r rune) d2ast.ArrayNodeBox {
 
 	p.replay(r)
 	vbox := p.parseValue()
-	if vbox.UnquotedString != nil && vbox.UnquotedString.ScalarString() == "" {
+	if vbox.UnquotedString != nil && vbox.UnquotedString.ScalarString() == "" &&
+		!(len(vbox.UnquotedString.Value) > 0 && vbox.UnquotedString.Value[0].Substitution != nil) {
```
nhooyr, line 1564: Was this change intended? I'm not sure what it's trying to do.

nhooyr: Also that first condition that was added seems redundant as the earlier existing condition checks that the ScalarString is empty.

d2ir/compile.go
```diff
@@ -167,6 +188,12 @@ func (c *compiler) resolveSubstitutions(varsStack []*Map, node Node) {
 						c.errorf(node.LastRef().AST(), `cannot substitute map variable "%s" as part of a string`, strings.Join(box.Substitution.IDA(), "."))
```
nhooyr, line 188: needs to be updated.

d2ir/compile.go
```diff
@@ -167,6 +188,12 @@ func (c *compiler) resolveSubstitutions(varsStack []*Map, node Node) {
...
 						return
 					}
+					switch n := node.(type) {
```
nhooyr, line 191: What's the reason for this?

d2ir/compile.go
nhooyr: Also should check for composite here as both could be nil.

d2ir/compile.go
```diff
@@ -145,20 +152,34 @@ func (c *compiler) resolveSubstitutions(varsStack []*Map, node Node) {
...
+								arr.Values = append(append(arr.Values[:i], resolvedArr.Values...), arr.Values[i+1:]...)
+								break
+							}
+						}
+					case *Field:
+						if resolvedField.Map() != nil {
+							OverlayMap(ParentMap(n), resolvedField.Map())
+						}
+						// Remove the placeholder field
+						m := n.parent.(*Map)
```
nhooyr, line 177: why not use ParentMap?

nhooyr (changes requested):

d2ir/compile.go
nhooyr: Like you can't spread a non composite period, not only into a composite.

d2ir/compile.go
nhooyr: I guess it is guaranteed to be a map so nvm.

d2ir/compile.go
nhooyr: I'm not sure if we should do this. I wouldn't expect it at least.

nhooyr: Only an explicit `null` should do this I would say.

d2ir/compile.go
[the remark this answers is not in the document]
nhooyr: It doesn't have any effect because the new vars field is appended to the end of the board whereas it should be prepended if we wanted it to take effect. But we don't since things work anyway.

d2ir/compile.go
alixander: ah right

d2ir/compile.go
alixander: both could be nil is handled in the compiler (cleaner I felt). there's a test for it. `bad-var`

d2ir/compile.go
alixander: we check if primary is nil in places. https://github.com/alixander/d2/blob/ffdb5121f49e16e4f82f63cb034452881a77a91d/d2compiler/compile.go#L437

(if link is broken, line 1 of `compileReserved` function)

i think it makes sense. the primary was the substitution, which got resolved. and the resolution could be just a map with no primary.

d2ir/compile.go
alixander: oh i see. calling `String` is what invoked the formatter.

d2ir/compile.go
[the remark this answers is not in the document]
alixander: ohhhh right.

d2ir/compile.go
nhooyr: I don't see how that's cleaner nor why it's in `d2compiler`. The IR compiler should be complaining about that. Variables are not the responsibility of `d2compiler`.

d2ir/compile.go
nhooyr: I don't understand what you wrote there. In regards to that check, why bother with that check? Shouldn't there be an error instead if you have a primary on `class`?

If I had a shape with a primary and a composite. Then I assigned it a composite variable, I wouldn't expect that the primary was overridden to nil even though the variable I assigned it had no primary.

nhooyr (changes requested):
left two comments

d2ir/compile.go
alixander: nvm it was wrong behavior anyway. no composite and no primary should be allowed (e.g. maybe people want to spread), but it's a mistake when the resolved variable itself has no composite and no primary, which is right here.

d2compiler/compile.go
```diff
@@ -281,12 +281,6 @@ func (c *compiler) compileField(obj *d2graph.Object, f *d2ir.Field) {
 		if f.Map() != nil {
 			if len(f.Map().Edges) > 0 {
```
nhooyr, line 282: This should be in `d2ir` too.

nhooyr (changes requested):

d2ir/compile.go
alixander: > Shouldn't there be an error instead if you have a primary on class?

no, class can be either primary or composite

> why bother with that check?

this check is everywhere in the compiler. `Primary() != nil` or `Primary() == nil`. i think that should continue to be a way of easily checking if the field has a primary or not.

> If I had a shape with a primary and a composite. Then I assigned it a composite variable, I wouldn't expect that the primary was overridden to nil even though the variable I assigned it had no primary.

I would. 
```d2
vars: { x: { a: b } }
a: 1
a: ${x}
```

it's definitely unexpected to me that it retains the 1 here

nhooyr: Oh I see, I misunderstood the check in that code.

> this check is everywhere in the compiler. Primary() != nil or Primary() == nil. i think that should continue to be a way of easily checking if the field has a primary or not.

Why are you bringing this up? When did I assert that Primary != nil is not an adequate way to check if a field has a primary or not?

> it's definitely unexpected to me that it retains the 1 here

I'll defer.

alixander: > When did I assert that Primary != nil is not an adequate way to check if a field has a primary or not?

you didn't. but if you substitute a composite with no primary and keep around the primary, then `Primary != nil` is no longer an adequate way to check. those places that check don't make the assumption that the primary could be a substitution that actually has no primary

d2compiler/compile.go
```diff
@@ -329,7 +336,7 @@ func (c *compiler) compileField(obj *d2graph.Object, f *d2ir.Field) {
 			Scope:           fr.Context.Scope,
 			ScopeAST:        fr.Context.ScopeAST,
 		}
-		if fr.Context.ScopeMap != nil {
+		if fr.Context.ScopeMap != nil && !d2ir.IsVar(fr.Context.ScopeMap) {
```
nhooyr, line 339: Instead, strip the IR of `vars` as you add them to the varStack. That way the compiler won't see vars.

nhooyr (changes requested):

d2compiler/compile.go
```diff
@@ -277,6 +277,13 @@ func (c *compiler) compileField(obj *d2graph.Object, f *d2ir.Field) {
 			}
 		}
 		return
+	} else if f.Name == "vars" {
+		// if f.Map() != nil {
```
nhooyr, line 281: Remove this block.

nhooyr (changes requested):

d2compiler/compile.go
alixander: when i do spread substitution, i'm overlaying the var map. that retains all the parent scope, field references, etc. it seems annoying to clean it.

alixander: i also don't think i should be modifying the IR like that. like for language tooling, going to def of a var requires it be present in the IR

d2ir/compile.go
nhooyr: I'm super confused. Are you trying to use `Primary != nil` to check that there is no composite?

d2compiler/compile.go
nhooyr: Fair enough.

nhooyr (approved):

PR #1479, opened 2023-07-13
d2ir: Globs

nhooyr:


d2ir/d2ir.go
```diff
@@ -825,65 +896,235 @@ func (m *Map) GetEdges(eid *EdgeID) []*Edge {
...
+	}
+
+	for _, src := range srcFA {
+		for _, dst := range dstFA {
+			if src == dst && (len(srcFA) > 1 || len(dstFA) > 1) {
+				// Globs do not make self edges.
+				continue
+			}
+
+			if srcKP.HasDoubleGlob() || dstKP.HasDoubleGlob() {
```
alixander, line 1087: this is kind of unexpected right?

```d2
vars: {
  d2-config: {
    layout-engine: elk
  }
}

a

**.b -> c
```

Screen Shot 2023-07-30 at 7 58 45 AM

d2ir/pattern.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	var fa []*Field
+	m._doubleGlob(&fa)
+	return fa, true
+}
+
+func (m *Map) _doubleGlob(fa *[]*Field) {
+	for _, f := range m.Fields {
+		if _, ok := d2graph.ReservedKeywords[f.Name]; ok {
+			if _, ok := d2graph.BoardKeywords[f.Name]; !ok {
```
alixander, line 21: it shouldn't go into `layers` right? that's completely new board

d2ir/d2ir.go
```diff
@@ -663,29 +679,73 @@ func (m *Map) EnsureField(kp *d2ast.KeyPath, refctx *RefContext) (*Field, error)
...
+	var fa []*Field
+	err := m.ensureField(i, kp, refctx, create, &fa)
+	return fa, err
 }
 
-func (m *Map) ensureField(i int, kp *d2ast.KeyPath, refctx *RefContext) (*Field, error) {
+func (m *Map) ensureField(i int, kp *d2ast.KeyPath, refctx *RefContext, create bool, fa *[]*Field) error {
+	us, ok := kp.Path[i].Unbox().(*d2ast.UnquotedString)
+	if ok && us.Pattern != nil {
+		fa2, ok := m.doubleGlob(us.Pattern)
```
alixander, line 691: i wonder if in the parser we should restrict double globs to just be on their own. here i was testing what double glob behavior would be with prepending or appending something, but i realize it actually just treated this as two separate globs, which is meaningless. 

```d2
fast: {
  a
  far
}

task: {
  a
}

f**.style.fill: red
```

Screen Shot 2023-07-30 at 11 24 23 AM

d2ir/d2ir.go
```diff
@@ -663,29 +679,73 @@ func (m *Map) EnsureField(kp *d2ast.KeyPath, refctx *RefContext) (*Field, error)
...
+					}
+					err := f.Map().ensureField(i+1, kp, refctx, create, fa)
+					if err != nil {
+						return err
+					}
+				}
+			}
+			return nil
+		}
+		for _, f := range m.Fields {
```
alixander, line 710: you can put this in the else of the above right

d2ir/d2ir.go
```diff
@@ -1159,6 +1400,26 @@ func IDA(n Node) (ida []string) {
 	}
 }
 
+// RelIDA returns the absolute path to n relative to p.
```
alixander, line 1403: just "the path to n relative to p",

d2ir/d2ir.go
```diff
@@ -825,65 +896,235 @@ func (m *Map) GetEdges(eid *EdgeID) []*Edge {
...
+
+	for _, src := range srcFA {
+		for _, dst := range dstFA {
+			if src == dst && (len(srcFA) > 1 || len(dstFA) > 1) {
+				// Globs do not make self edges.
+				continue
+			}
+
+			if srcKP.HasDoubleGlob() || dstKP.HasDoubleGlob() {
+				// If either has a double glob we only select leafs, those without children.
```
alixander, line 1088: it shouldn't be either, it should only apply to the side that has the double glob.

Otherwise you can't do this:

```d2
fast: {
  a
  far
}

task: {
  a
}

task.** -> fast
```

Screen Shot 2023-07-30 at 11 33 02 AM

d2ir/d2ir.go
```diff
@@ -825,65 +896,235 @@ func (m *Map) GetEdges(eid *EdgeID) []*Edge {
...
+		return err
+	}
+	dstFA, err := m.EnsureField(dstKP, nil, true)
+	if err != nil {
+		return err
+	}
+
+	for _, src := range srcFA {
+		for _, dst := range dstFA {
+			if src == dst && (len(srcFA) > 1 || len(dstFA) > 1) {
```
alixander, line 1082: i don't understand what the righthand side of `&&` is checking for

alixander (changes requested):
didn't inspect all the boilerplate code, but i understand the implementation

d2ir/pattern.go
nhooyr: It should, you might be setting all shapes a `page` everywhere. Or some kind of default styling.

d2ir/d2ir.go
nhooyr: Not sure what you mean. d2ir can use the parsed pattern from the parser to detect the kind of double glob you've given an example of above, it just doesn't mean anything beyond a single glob right now.

d2ir/d2ir.go
nhooyr: I don't see how. This is the code for single globs, the above is the code for double globs. They have different behaviour.

d2ir/d2ir.go
nhooyr: Updated, should be more clear now.

d2ir/d2ir.go
nhooyr: fixed

d2ir/d2ir.go
nhooyr: fixed.

alixander (approved):

PR #1484, opened 2023-07-18
d2ir: Fix variables to passthrough primary type

nhooyr:
If the variable is a number it should be parsed as a number and not coerced into a string. See test change.

### DIFF
--- a/testdata/d2compiler/TestCompile2/vars/basic/shape-label.exp.json
+++ b/testdata/d2compiler/TestCompile2/vars/basic/shape-label.exp.json
@@ -92,7 +92,23 @@
                 "range": "d2/testdata/d2compiler/TestCompile2/vars/basic/shape-label.d2,4:4:29-4:5:30",
                 "value": [
                   {
-                    "string": "im a var"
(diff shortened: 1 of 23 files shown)


alixander (approved):

testdata/d2compiler/TestCompile2/vars/basic/shape-label.exp.json
```diff
@@ -92,7 +92,23 @@
                 "range": "d2/testdata/d2compiler/TestCompile2/vars/basic/shape-label.d2,4:4:29-4:5:30",
                 "value": [
                   {
-                    "string": "im a var"
+                    "substitution": {
```
alixander, line 95: does this change make sense to u?

nhooyr: Yes, the AST was being modified in place before but now it's not in these cases.

alixander: ohh ok

PR #1503, opened 2023-07-29
layers support in watch mode

### LINKED ISSUE #1176: svg composition for watch mode
the internal links all work normally, but in watch mode does not

alixander:
Let users develop multi-board diagrams in watch mode.

Example of clicking links to navigate in watch mode:

https://github.com/terrastruct/d2/assets/3120367/5dfb4e99-919b-4b6a-8ff1-7ffdd4d43738

The back button doesn't work to update the board, at least on mac. I guess it doesn't trigger a page refresh for some reason, but this can be deferred.

closes #1176

### DIFF
--- a/d2cli/watch.go
+++ b/d2cli/watch.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 	"time"
 
@@ -49,6 +50,7 @@ type watcherOpts struct {
 	port            string
 	inputPath       string
 	outputPath      string
+	boardPath       string
 	pwd             string
 	bundle          bool
 	forceAppendix   bool
@@ -362,7 +364,7 @@ func (w *watcher) compileLoop(ctx context.Context) error {
 			w.pw = newPW
 		}
 
-		svg, _, err := compile(ctx, w.ms, w.plugins, w.layout, w.renderOpts, w.fontFamily, w.animateInterval, w.inputPath, w.outputPath, w.bundle, w.forceAppendix, w.pw.Page)
+		svg, _, err := compile(ctx, w.ms, w.plugins, w.layout, w.renderOpts, w.fontFamily, w.animateInterval, w.inputPath, w.outputPath, w.boardPath, w.bundle, w.forceAppendix, w.pw.Page)
 		errs := ""
 		if err != nil {
 			if len(svg) > 0 {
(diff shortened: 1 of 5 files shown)


d2cli/watch.go
```diff
@@ -430,15 +432,23 @@ func (w *watcher) handleRoot(hw http.ResponseWriter, r *http.Request) {
...
+	<link id="favicon" rel="icon" href="/static/favicon.ico">
 </head>
 <body data-d2-dev-mode=%t>
 	<div id="d2-err" style="display: none"></div>
 	<div id="d2-svg-container"></div>
 </body>
 </html>`, filepath.Base(w.outputPath), w.devMode)
+
+	// if path is "/x.svg", we just want "x"
+	split := strings.Split(strings.TrimPrefix(r.URL.Path, "/"), ".")
```
nhooyr, line 446: Use `strings.IndexLastByte` here instead to clearly indicate what the code is trying to do here.

d2cli/watch.go
```diff
@@ -430,15 +432,23 @@ func (w *watcher) handleRoot(hw http.ResponseWriter, r *http.Request) {
...
+	boardPath := strings.Join(split[:len(split)-1], ".")
+	if boardPath != w.boardPath {
```
nhooyr, line 448: What's the purpose of this check? Shouldn't we always render the requested board path?

d2target/d2target.go
```diff
@@ -72,6 +73,66 @@ type Diagram struct {
...
+
+	head := path[0]
+
+	if head == "index" {
+		return d
+	}
+
+	switch head {
+	case "layers", "scenarios", "steps":
+		if len(path) < 2 {
```
nhooyr, line 92: Just repeat this for each switch case below.

d2target/d2target.go
```diff
@@ -72,6 +73,66 @@ type Diagram struct {
...
+			return nil
+		}
+	}
+
+	switch head {
+	case "layers":
+		for _, b := range d.Layers {
+			if b.Name == path[1] {
+				return b.GetBoard(strings.Join(path[2:], string(os.PathSeparator)))
```
nhooyr, line 101: Add and reimplement in terms of an unexported `getBoard` that takes the path as an array to avoid this.

d2target/d2target.go
```diff
@@ -72,6 +73,66 @@ type Diagram struct {
...
+		}
+	case "steps":
+		for _, b := range d.Steps {
+			if b.Name == path[1] {
+				return b.GetBoard(strings.Join(path[2:], string(os.PathSeparator)))
+			}
+		}
+	}
+
+	for _, b := range d.Layers {
```
nhooyr, line 118: Should this be allowed? Why not require the board type prefix?

nhooyr (changes requested):

d2cli/watch.go
alixander: prevent duplication of compile requests on the initial load

d2target/d2target.go
alixander: when the root diagram has just `layers`, e.g.

```d2
x

layers: {
  y: { ... }
}
```

then the file dir looks like

```
- y/
  - ...
- index.svg
```

---------

But if the root diagram has multiple board types, like

```d2
x

layers: { ... }
scenarios: { ... }
```

then the directory will look like

```
- layers/
- scenarios/
- index.svg
```

nhooyr: Right!

d2cli/watch.go
nhooyr: Not clear to me how this accomplishes that. Can you elaborate? Also, fairly certain this is racey, I don't think you can modify `w.boardPath` like that without putting it under a mutex as the compilation occurs in the background.

nhooyr (changes requested):
.

d2cli/watch.go
alixander: > Can you elaborate?

I just figure on initial load it's already compiling (behavior before this PR). So I don't want to add another recompile on top of that

> fairly certain this is racey

it calls `requestCompile()`, which dodges races with channels.

alixander: will address in another PR if needed

nhooyr: That doesn't dodge the race. You can't modify a variable that's read in another goroutine without synchronization. The channel send merely guarantees that the compileLoop goroutine will run a compile at some point in the future, not that it isn't already running one.

alixander: i did `go install -race .`, then reran the manual test in the PR description, and it didn't say there was a race.

nhooyr: The race detector never produces false positives, but it does produce false negatives. It can't prove your program doesn't have races, only that it does.

alixander: will look into this again later https://github.com/terrastruct/d2/issues/1508

PR #1509, opened 2023-07-29
d2ir: Single Level Ampersand Filters

nhooyr:
Depends on #1479 being merged.

I didn't implement recursive filters in this PR. i.e you can't do something like:

```d2
*: {
  &style: {
    fill: red
    multiple: true
  }
  label: red and multiple
}
```

You have to do the following instead:
```d2
*: {
  style: {
    &fill: red
    &multiple: true
  }
  label: red and multiple
}
```


nhooyr:
In fact I don't think we want to have composite filters anyhow. I think single field filters are easier to read and much easier to implement. For example for composite filters we'd have to error if the filter itself had ampersand filters because that wouldn't make sense.

d2ir/compile.go
```diff
@@ -396,20 +410,87 @@ func (c *compiler) compileKey(refctx *RefContext) {
...
+	for _, f := range fa {
+		c._compileField(f, refctx)
+	}
+}
+
+func (c *compiler) ampersandFilter(refctx *RefContext) bool {
+	if !refctx.Key.Ampersand {
+		return true
+	}
+	if len(refctx.Key.Edges) > 0 {
```
alixander, line 432: is ampersand not functional for edges?

```d2
fast: {
  a
  far
}

task: {
  a
}

rats

* -> rats: {
  &label: task
}
```

Screen Shot 2023-07-30 at 11 42 05 AM

d2ir/compile.go
```diff
@@ -396,20 +410,87 @@ func (c *compiler) compileKey(refctx *RefContext) {
...
+	return true
+}
+
+func (c *compiler) _ampersandFilter(f *Field, refctx *RefContext) bool {
+	f2 := ParentMap(f).Map().GetField(refctx.Key.Key.IDA()...)
+	if f2 == nil {
+		return false
+	}
+	if refctx.Key.Value.ScalarBox().Unbox() == nil {
+		c.errorf(refctx.Key, "ampersand filters cannot be composites")
```
alixander, line 459: i think we should call them "glob filters" user-facing.

alixander (changes requested):

d2ir/compile.go
nhooyr: Ah I forgot will fix.

nhooyr: Oh no so it does work on edge maps. But you can't use ampersands to filter maps with edges. E.g. you can't say apply this to a map with an edge from y -> z.

nhooyr: Nvm there was a bug. fixed. See test.

alixander: i don't understand. the test doesn't test this case. after latest pull, i now get `glob filters cannot be used outside globs` with the same code.

alixander (approved):

d2ir/compile.go
nhooyr: I'm not sure what you're trying to do in your test there.

PR #1522, opened 2023-08-01
ci: Fix daily part 2

nhooyr:

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -6,8 +6,4 @@ if [ "$*" = "" ]; then
   set ./...
 fi
 
-if [ "${CI:-}" ]; then
-  export FORCE_COLOR=1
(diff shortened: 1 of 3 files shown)


ci/test.sh
```diff
@@ -6,8 +6,4 @@ if [ "$*" = "" ]; then
   set ./...
 fi
 
-if [ "${CI:-}" ]; then
```
alixander, line 9: why remove this entirely? then other CI will fail

nhooyr: It won't, it always go through `make.sh`.

nhooyr: Oh but I guess I have to remove DAILY then.

alixander (approved):

PR #1523, opened 2023-08-02
fix(cli): do not exit after 1st formatted file

maxbrunet:
Hello! 👋

Thank you so much for `d2`, loving it so far!

Found a small bug in `d2 fmt`: If ran on multiple non-formatted files at once, it would format the 1st and exit, leaving the others non-formatted.

### DIFF
--- a/ci/release/changelogs/next.md
+++ b/ci/release/changelogs/next.md
@@ -5,3 +5,5 @@
 #### Improvements 🧹
 
 #### Bugfixes ⛑️
+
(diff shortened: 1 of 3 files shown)


maxbrunet:
> can you add a test in [`master`/e2etests-cli/main_test.go](https://github.com/terrastruct/d2/blob/master/e2etests-cli/main_test.go?rgh-link-date=2023-08-02T18%3A46%3A57Z) ?
> i've just added a basic fmt test that you can follow as example: #1527

Awesome, exactly what I was looking for!

alixander (approved):
🙏 !

PR #1525, opened 2023-08-02
d2parser: Support reading utf16 files

nhooyr:


d2parser/parse.go
```diff
@@ -27,24 +35,43 @@ type ParseOptions struct {
...
 	}
 
 	p := &parser{
-		path:   path,
-		reader: r,
+		path: path,
 
-		utf16: opts.UTF16,
-		err:   opts.ParseError,
+		utf16Pos: opts.UTF16Pos,
```
alixander, line 52: so `opts.UTF16Pos` is never actually used then? if detection is 100%, then we can get rid of CompilerOpts.UTF16Pos and ParserOpts.UTF16Pos and detect every time.

but if it's not absolute, we should only run detection and change if UTF16Pos is passed in as false

alixander (changes requested):

d2parser/parse.go
alixander: can you get the changelog line while you're at it

alixander (commented):
should we write the output in utf-16 too?

nhooyr:
For now let's stick to UTF-8 for output.

Will perhaps be annoying with `d2 fmt` but as far as I know, Windows/modern text editors fully support UTF-8 if detected. A user's text editor will just switch encodings.

nhooyr:
If it ends up being an issue, see `utf16_gen.go` for how to write Go strings as UTF-16.

d2parser/parse.go
alixander: ohh lol

alixander (approved):

PR #1528, opened 2023-08-03
feat: add d2-fmt pre-commit hook

maxbrunet:
Add a [pre-commit](https://pre-commit.com/) hook for `d2 fmt`, usage:

```yaml
# .pre-commit-config.yaml
repos:
  - repo: https://github.com/terrastruct/d2
    rev: v0.7.0
    hooks:
      - id: d2-fmt
```

`pre-commit` will take care of cloning the repo, downloading `go` if needed, and building `d2`: [pre-commit.com/#golang](https://pre-commit.com/#golang)

Can be tested with:

```shell
pre-commit try-repo --all-files --ref=feat/fmt-pre-commit-hook https://github.com/maxbrunet/d2
```

_Note:_ It would be nice to have a `d2` hook to generate the diagrams, but `d2Version=` is included in the headers and requires the version `ldflag` to be set at build time which is not supported by `pre-commit`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: d2-fmt
+  name: d2 fmt
+  description: Format d2 files
+  entry: d2 fmt
+  language: golang
+  files: \.d2$


alixander (changes requested):
hmm was initially reluctant to have this file take up a row in the top-level dir, but doing a search on GitHub for this framework's adoption yields some pretty widespread usage so I'm for it. Thanks for doing this!

Can you remove it from changelog though? Doesn't feel like a new d2 feature.

alixander (approved):

PR #1534, opened 2023-08-04
d2-vscode LanguageServerChanges

BarryNolte:
These changes are for the language server in d2-vscode.

When the D2_LSP_MODE environment variable is set, the d2 cli will read the d2 file, produce the ast (and possible errors), convert it to JSON, print it out to stdout, then terminate.  This was done this way to keep the changes to the d2 cli code to a minimum.

PR for d2-vscode to come after this is accepted

@gavin-ts  @alixander


alixander:
AFAIK, LSPs should be daemon processes, otherwise it'd have to reindex each call or load an index from disk, both of which seem too slow for autocomplete. I haven't looked too much into LSPs, is this just for testing or the beginning of a true LSP effort? @BarryNolte

BarryNolte:
I should have made this clearer.  This is the only change needed in the d2 cli code.  I wasn't going to duplicate the language parsing in typescript, I'm too old for that, so I'm leveraging the de facto parser to get the AST for the language server.  I have a working and 99% complete language server (yes, it runs as a deamon) in the d2-vscode/LanguageServer branch, I'm taking the weekend to squash the last couple bugs, and I wanted this in first.

d2compiler/compile.go
```diff
@@ -27,18 +29,46 @@ type CompileOptions struct {
...
+
+	lod.Ast = ast
 	if err != nil {
+		lod.Err = err
 		return nil, nil, err
 	}
 
+	if lspMode {
+		return nil, nil, nil
+	}
```
gavin-ts, line 70:
```suggestion
func Compile(p string, r io.Reader, opts *CompileOptions) (*d2graph.Graph, *d2target.Config, error) {
	if opts == nil {
		opts = &CompileOptions{}
	}

	ast, err := d2parser.Parse(p, r, &d2parser.ParseOptions{
		UTF16Pos: opts.UTF16Pos,
	})

	if os.Getenv("D2_LSP_MODE") == "1" {
		jsonOutput, _ := json.Marshal(LspOutputData{Ast: ast, Err: err})
		fmt.Print(string(jsonOutput))
		os.Exit(42)
		return nil, nil, err
	}

	if err != nil {
		return nil, nil, err
	}
```

I think it is cleaner to do this without the global + defer

gavin-ts (changes requested):

d2compiler/compile.go
BarryNolte: Yes, better.  I didn't pull off my goal of least amount of code churn possible.

d2compiler/compile.go
```diff
@@ -35,6 +43,14 @@ func Compile(p string, r io.Reader, opts *CompileOptions) (*d2graph.Graph, *d2ta
 	ast, err := d2parser.Parse(p, r, &d2parser.ParseOptions{
 		UTF16Pos: opts.UTF16Pos,
 	})
+
+	if os.Getenv("D2_LSP_MODE") == "1" {
+		jsonOutput, _ := json.Marshal(LspOutputData{Ast: ast, Err: err})
+		fmt.Print(string(jsonOutput))
+		os.Exit(42)
+		return nil, nil, err
+	}
```
gavin-ts, line 52: This looks ok but I wonder if there's a better place for it? What if we create a `d2lib.Parse` alongside the existing `d2lib.Compile` that just does the existing parsing above, and then in `d2cli.compile` we check this env flag there, and call `d2lib.Parse` instead of `d2lib.Compile` writing this json and exiting from d2cli?

https://github.com/terrastruct/d2/blob/master/d2cli/main.go#L395

gavin-ts: or it could even be a `d2cli.parse` outside of `d2cli.compile`, but I think it seems better to have the os.Exit / output in d2cli instead of d2compiler

gavin-ts (approved):
I can update where this code is in a follow up

BarryNolte:
Sounds good.  Thanks!

PR #1552, opened 2023-08-16
d2ir: Implement lazy globs and triple glob

nhooyr:
This finishes up the globs implementation!

See tests for what I mean by lazy globs and what the triple glob does.


alixander (changes requested):
`failed to compile: /Users/alexanderwang/dev/d2-docs/static/bespoke-d2/triple-glob.d2:1:11: fill must be style.fill`

```d2
***.style.fill: yellow
**.shape: circle
*.style.multiple: true

x: {
  y
}

layers: {
  next: {
    a
  }
}
```

alixander (changes requested):
```d2
# Create a shape `a`
a

* -> y

b
c
```

Screen Shot 2023-08-15 at 10 33 13 PM

alixander (changes requested):
```d2
**.child

a
b
c
```

Screen Shot 2023-08-16 at 1 01 52 AM

nhooyr:
Fixed.

alixander (changes requested):
i think globs need to be overridable. currently `a` is yellow with this script, but it should be red.

```d2
**.style.fill: yellow
**.style.fill: red

a
```

nhooyr:
Certainly was the intent, not sure how that's happening.

nhooyr:
fixed.

alixander (changes requested):
did u hardcode the word layers in that last fix? i get this err again if i just change to scenarios

```d2
**.style.fill: red

scenarios: {
  b: {
    a -> b
  }
}
```

Screen Shot 2023-08-16 at 6 34 32 PM

alixander (changes requested):
```d2
(*** -> ***)[*].label: hi

a -> b

layers: {
  hi: {
    (*** -> ***)[*].label: bye

    scenarios: {
      b: {
        # This label is "hi", but it should be "bye"
        a -> b
      }
    }
  }
}
```

nhooyr:
> did u hardcode the word layers in that last fix? i get this err again if i just change to scenarios
nope using d2graph.BoardKeywords so I'm very confused by that too.

alixander (commented):
what do you think about this one? 

```d2
(*** -> ***)[*].label: hi

# This is "hey" right now but should be "hi"?
a -> b

(*** -> ***)[*].label: hey
```

interpretation 1: i've defined the first glob and expect all connections to lazily apply it until it's overridden, so the current behavior is broken.

interpretation 2: globs are global. i've defined two globs of the same pattern, so the last one should win, no matter what order of statements come in between/around them, so the current behavior is correct.

---

i'm leaning to interpretation 1. i.e. Globs should apply backwards __unless__ an identical pattern already has lazily applied to it.

nhooyr:
I think number 2 is the right behavior as it's implemented now. Last one to apply always wins as that's consistent with the rest of the language. Otherwise you can't override/unset globs which isn't good. Especially if you're importing a d2 file you don't control.

Why are you leaning towards 1? Under what kind of scenario do you see this coming up where 1 is preferable?

nhooyr:
>  lazily apply it until it's overridden, so the current behavior is broken.

But you did override it with the second one?

nhooyr:
> i'm leaning to interpretation 1. i.e. Globs should apply backwards unless an identical pattern already has lazily applied to it.

In this case the final glob isn't being applied backwards. It's being applied normally.

alixander (changes requested):
ok, i can see that. moving on, this one's a bug for sure:

```d2
(*** -> ***)[*].label: hi

a -> b: {
  label: bye
}
```

Screen Shot 2023-08-17 at 9 44 39 AM

alixander (changes requested):
looks like with filters, it's not overriding

```d2
x

*.style.opacity: 0.1
*: {
  &label: x
  style.opacity: 1
}
```

Screen Shot 2023-08-17 at 11 41 47 AM

nhooyr:
First one fixed, working on this next one.

nhooyr:
Also btw https://github.com/terrastruct/d2/pull/1552#pullrequestreview-1581636406 was because of `***` being changed to `**`, not because of layers -> scenarios.

And I think this next one isn't related to globs but that `label` isn't handled specially by filters to match the ID of a field without a label.

alixander:
@nhooyr 

nah i just tried this and it's the same issue

```d2
*.style.opacity: 0.1
*: {
  &label: x
  style.opacity: 1
}

x: x
```

nhooyr:
Same issue there, `label` needs special handling. Pushed fix for `*Field` but I think I need to add the same exception for `*Edge` too. So hold on.

nhooyr:
Fixed.

alixander (changes requested):
looks like the label filtering isn't working here for edges

```d2
(* -> *)[*].style.opacity: 0.1

(* -> *)[*]: {
  &label: hi
  style.opacity: 1
}

x -> y: hi
x -> y
```

Screen Shot 2023-08-18 at 8 16 01 AM

nhooyr:
Fixed.

alixander (changes requested):
edge case: can you exclude double and triple globs from creating anything to class and table shapes? (styles should still apply, just can never create children on them)

```d2
***: {
  c: d
}

table: {
  shape: sql_table
  a: b
}

clas: {
  shape: class
  a: b
}
```

Screen Shot 2023-08-18 at 11 17 41 AM

alixander (changes requested):
edge case:

if i remove either of the globs, it becomes blue. i'm not sure why

```d2
*: {
  style.fill: red
}
**: {
  style.fill: red
}

table: {
  style.fill: blue
  shape: sql_table
  a: b
}
```
Screen Shot 2023-08-18 at 11 17 41 AM

alixander (changes requested):
[removed]

alixander (changes requested):
```d2
(a -> *)[*].style.stroke: red
(* -> *)[*].style.stroke: red

b -> c
```

Screen Shot 2023-08-18 at 11 28 10 AM

nhooyr:
Turns out there was a bug in `assertQuery`. If nothing matched the query then it would pass instead of failing lol. Like nothing as in the targeted IDA didn't exist... I think that explains both bugs but we'll see, fixing rn.

nhooyr:
> edge case: can you exclude double and triple globs from creating anything to class and table shapes? (styles should still apply, just can never create children on them)

I can and did prevent them from creating fields under the fields of classes/globs to prevent errors but I can't prevent them from creating fields under classes/globs easily.

nhooyr:
Nvm, even what I did makes no sense. Cause then `style` won't work. Yea I think only real solution for now is to disable the error and just not render those children.

nhooyr:
Nvm that I figured out a way.

nhooyr:
All fixed, ready for review again.

d2ir/pattern_test.go
```diff
@@ -314,21 +339,355 @@ task.** -> fast
...
+				assertQuery(t, m, 6, 2, nil, "")
+				assertQuery(t, m, 0, 0, 0.1, "(x -> y)[0].style.opacity")
+				assertQuery(t, m, 0, 0, 0.1, "(x -> y)[1].style.opacity")
+			},
+		},
+		{
+			name: "alixander-review/7",
+			run: func(t testing.TB) {
+				m, err := compile(t, `
+*: {
```
alixander, line 522: this wasn't the test btw. it was "if i comment either of the globs out, it becomes blue". so it shoudl test that having each of these globs removed, the table fill is still red

nhooyr: Yea I didn't fully understand your comment. There was a bug here but the bug was that it should always be blue because lazy globs do not override normal keys. Only globs after normal keys override them. So if you move one of the globs below table, then it'll work as you suggest.

alixander (changes requested):
getting this weird behavior:

```d2
# Nulling glob doesn't work
*a.icon: https://icons.terrastruct.com/essentials%2F073-add.svg
a.icon: null

# Regular icon nulling works
b.icon: https://icons.terrastruct.com/essentials%2F073-add.svg
b.icon: null

# Shape nulling works
*.shape: circle
a.shape: null
b.shape: null
```

Screen Shot 2023-08-20 at 9 55 26 AM

the thing is, if I just prepend a line, `a` as line 0 (declare `a` before globs), the nulling works.

alixander (changes requested):
perhaps related. there's something about declaring an object first and then setting attributes later that makes the glob not match. I suspect that every time an object's attributes change, globs need to re-evaluate, instead of just on creation

```d2
*: {
  &label: a
  style.fill: yellow
}

a
# if i remove this line, the glob applies as expected
b
b.label: a
```

Screen Shot 2023-08-20 at 10 32 21 AM

alixander (changes requested):
this should null the object instead of just nulling label

```d2
parent: {
  a -> b1
  a -> b2
  a -> b3

  b1 -> c1
  b1 -> c2

  c1: {
    c1-child.class: hidden
  }

  c2: {
    C2-child.class: hidden
  }
  c2.class: hidden
  b2.class: hidden
}

classes: {
  hidden: {
    style: {
      fill: red
    }
  }
}

# Error
**: null {
  &class: hidden
}
```

Screen Shot 2023-08-30 at 12 00 29 AM

alixander (changes requested):
is it easy to get globs working for imports?

## rules.d2
```d2
*.style.fill: red
```

## index.d2
```d2
...@rules
x
```

nhooyr:
Right now no unfortunately. Would require changing how imports work to merge ASTs together instead of IRs.

alixander:
i think that can be another use case of `***`. they cross board boundaries and file boundaries

alixander:
can we encode globs into the IR? i think imports should continue just dealing with IRs, globs just need to be in there

nhooyr:
Perhaps, I'm not sure of a good way to do it right now.

alixander (commented):
this doesn't work, but if I remove the `x`, it works. Looks like something straightforward missed?

Screen Shot 2023-09-04 at 11 32 53 AM

Screen Shot 2023-09-04 at 11 32 59 AM

alixander (approved):

PR #1583, opened 2023-09-12
Fix UsageError typo for `--animate-interval`

dcousens:
This pull request fixes a typo when displaying a `xmain.UsageErrorf`, the flag displayed says you need to specify `-animate-interval`, when it should be `--animate-interval`.

### DIFF
--- a/d2cli/main.go
+++ b/d2cli/main.go
@@ -216,9 +216,9 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
 	if outputPath != "-" {
 		outputPath = ms.AbsPath(outputPath)
 		if *animateIntervalFlag > 0 && !outputFormat.supportsAnimation() {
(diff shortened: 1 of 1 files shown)


alixander:
interesting, different CLI tools handle this differently:

Screen Shot 2023-09-12 at 10 31 41 AM
Screen Shot 2023-09-12 at 10 34 22 AM

I guess because `animate-interval` doesn't have a shorthand form, it should be `--` always.

dcousens:
I think the GNU and POSIX convention is to use `-` for short form, and `--` for long form, in any event, `-animate-interval` doesn't actually work with `d2`.

```bash
err: bad usage: failed to parse flags: unknown shorthand flag: 'a' in -animate-interval
```

alixander:
@dcousens right, this is a welcome change then, thank you. Would you mind updating the failing test that's blocking CI?

PR #1621, opened 2023-09-26
add MkDocs Plugin to community plugins

landmaj:
This PR adds [MkDocs Plugin](https://github.com/landmaj/mkdocs-d2-plugin) to the list of community plugins.

At work we are in the process of switching from PlantUML to D2 and I discovered there was no plugin for [MkDocs](https://www.mkdocs.org/), which we use to generate documentation for our clients. So I wrote one. You can see a result of using the plugin [here](https://landmaj.github.io/mkdocs-d2-plugin/).


gavin-ts (approved):

gavin-ts:
oh sorry could you sign the commit? https://github.com/terrastruct/d2/pull/557#issuecomment-1367468730

landmaj:
Sorry, I didn't notice the requirement before. At least I learned something new today, I've never done it before. 🙂 Should be signed now.

gavin-ts:
thanks! We've run into this a few times where we don't notice the commit signature missing until after a review approval, so I just made this into its own clear check here: https://github.com/terrastruct/ci/pull/28 and it should make its way into the d2 repo soon.

Issue #1655, opened 2023-10-11
Null in sql_table constraint


landmaj:
When `null` is used in a list of sql_table constraints, it causes an invisible constraint in the rendered diagram.

```d2
table: {
  shape: sql_table
  invisible: {constraint: [foreign_key; null; next]}
  ok: {constraint: null}
}
```

It seems like `null` is never rendered, which is fine, you can double quote it to get the expected result. However, when used in a list of constraints, it results in an "invisible" constraint.

![d2](https://github.com/terrastruct/d2/assets/13437951/833941dd-a3d9-4c93-a658-7a4d5daf8ec0)

[The documentation](https://d2lang.com/tour/sql-tables/) is also missing any mention of this behaviour (requiring double quotes around `null`).

PR #1660, opened 2023-10-12
fix invisible null constraints in sql_table

landmaj:
Thix PR fixes issue #1655 

## Diagram
```d2
table: {
    shape: sql_table
    a: int {constraint: null}
    b: int {constraint: [TRUE; False; null; NULL]}
}
```

## Before
before

## After
after

## Summary

The behavior of null, when used in arrays, is now consistent with how other types work. When used as the only constraint (without array), only types which implement `d2ast.String` are allowed. Null is also allowed but does nothing, as expected. In arrays all types are allowed but booleans are converted to lower case. Null was previously converted to an empty string, this PR changes it to lower case string "null".

This could also be fixed by changing return value of `ScalarString()` method for `Null` type. I tried it and all tests pass,  however there is this commit: https://github.com/terrastruct/d2/commit/9d0c73cef2e51da0565e09cce2b25b187cad0029

Another solution would be to disallow everything but `d2ast.String` in constraint arrays, to keep it consistent with how non-array constraints work, however this could break existing diagrams and prevent them from compiling. Let me know if you would prefer this solution.

### DIFF
--- a/d2compiler/compile.go
+++ b/d2compiler/compile.go
@@ -443,7 +443,12 @@ func (c *compiler) compileReserved(attrs *d2graph.Attributes, f *d2ir.Field) {
 				if arr, ok := f.Composite.(*d2ir.Array); ok {
 					for _, constraint := range arr.Values {
 						if scalar, ok := constraint.(*d2ir.Scalar); ok {
-							attrs.Constraint = append(attrs.Constraint, scalar.Value.ScalarString())
+							switch scalar.Value.(type) {
+							case *d2ast.Null:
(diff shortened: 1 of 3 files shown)


landmaj:
BTW your new pipeline rejects SSH signatures on commits. Is this intended and I need to figure out GPG?

gavin-ts:
> BTW your new pipeline rejects SSH signatures on commits. Is this intended and I need to figure out GPG?

this is not intended looking into it

alixander (approved):
perfect PR. thank you @landmaj .

PR #1672, opened 2023-10-18
lib/color: avoid unnecessary byte/string conversion

Juneezee:
We should use `(*regexp.Regexp).MatchString` instead of `(*regexp.Regexp).Match([]byte(...))` to avoid unnecessary `[]byte` conversions and reduce allocations. A one-line change for free performance gain.

Example benchmark:

```go
func BenchmarkMatch(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := themeColorRegex.Match([]byte("N1")); !match {
			b.Fail()
		}
	}
}

func BenchmarkMatchString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := themeColorRegex.MatchString("N1"); !match {
			b.Fail()
		}
	}
}
```

Result:

```
goos: linux
goarch: amd64
pkg: oss.terrastruct.com/d2/lib/color
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkMatch-16          	 9894165	       114.3 ns/op	       2 B/op	       1 allocs/op
BenchmarkMatchString-16    	13439838	        83.61 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	oss.terrastruct.com/d2/lib/color	3.306s
```

### DIFF
--- a/lib/color/color.go
+++ b/lib/color/color.go
@@ -14,7 +14,7 @@ import (
 var themeColorRegex = regexp.MustCompile(`^(N[1-7]|B[1-6]|AA[245]|AB[45])$`)
 
 func IsThemeColor(colorString string) bool {
-	return themeColorRegex.Match([]byte(colorString))
+	return themeColorRegex.MatchString(colorString)
 }
 
 func Darken(colorString string) (string, error) {


ci/release/changelogs/next.md
```diff
@@ -4,6 +4,7 @@
 
 - Grid cells can now contain nested edges [#1629](https://github.com/terrastruct/d2/pull/1629)
 - Edges can now go across constant nears, sequence diagrams, and grids including nested ones. [#1631](https://github.com/terrastruct/d2/pull/1631)
+- Avoid unnecessary byte/string conversion [#1672](https://github.com/terrastruct/d2/pull/1672)
```
alixander, line 7: Changelogs should be relevant for users, this one's just a minor internal optimization.

alixander (changes requested):
Thank you! Especially appreciate the benchmark.

ci/release/changelogs/next.md
Juneezee: Changelog removed

alixander (approved):

PR #1691, opened 2023-10-31
compiler: Fix errors in globs

### LINKED ISSUE #1676: glob compiler crash
Adding `label.near` to the `**` glob crashes.

```d2
**: {
  style.border-radius: 16
}

**: {
  &label: end
  label: ⬤
  shape: circle
  height: 10
  width: 10
}

**: {
  &label: start
  label: ⬤
  label.near: center
  style.fill: black
  shape: circle
  height: 10
  width: 10
}

**: {
  &shape: diamond
  label: ""
  height: 30
  width: 30
}

start -> Check PIN

Check PIN: {
  direction: right

  start -> Enter PIN
  choice.shape: diamond

  Enter PIN -> choice: "/check PIN"
  choice -> Enter PIN: "[pin invalid]"
  choice -> end: "[pin OK]"
}

Check PIN -> Search Network: "[pin OK]"
Search Network -> Ready: network found

Check PIN -> Off: power off
Search Network -> Off: power off
Ready -> Off: power off
```

alixander:
fixes #1676

1. This picks up refs set by globs so that the error gets the right ref.
2. Which also adds a dedupe to the compiler list of errors since keys set by globs, if errored, will report multiple errors emanating from the same range.

### DIFF
--- a/d2ir/d2ir.go
+++ b/d2ir/d2ir.go
@@ -319,7 +319,7 @@ func (f *Field) Copy(newParent Node) Node {
 
 func (f *Field) LastPrimaryRef() Reference {
 	for i := len(f.References) - 1; i >= 0; i-- {
-		if f.References[i].Primary() && !f.References[i].DueToLazyGlob() {
+		if f.References[i].Primary() {
 			return f.References[i]
 		}
 	}
--- a/ci/release/changelogs/next.md
+++ b/ci/release/changelogs/next.md
@@ -14,3 +14,4 @@
 - Fixes use of `null` in `sql_table` constraints (ty @landmaj) [#1660](https://github.com/terrastruct/d2/pull/1660)
 - Fixes elk growing shapes with width/height set [#1679](https://github.com/terrastruct/d2/pull/1679)
 - Adds a compiler error when accidentally using an arrowhead on a shape [#1686](https://github.com/terrastruct/d2/pull/1686)
+- Correctly reports errors from invalid values set by globs. [#1691](https://github.com/terrastruct/d2/pull/1691)
--- a/d2compiler/compile.go
+++ b/d2compiler/compile.go
@@ -177,7 +177,14 @@ type compiler struct {
 }
 
 func (c *compiler) errorf(n d2ast.Node, f string, v ...interface{}) {
-	c.err.Errors = append(c.err.Errors, d2parser.Errorf(n, f, v...).(d2ast.Error))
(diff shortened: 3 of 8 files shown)


d2ir/d2ir.go
```diff
@@ -318,11 +318,19 @@ func (f *Field) Copy(newParent Node) Node {
 }
 
 func (f *Field) LastPrimaryRef() Reference {
+	var lastGlobSet Reference
```
gavin-ts, line 321: I think `(e *Edge) LastPrimaryRef()` might also need this if there are multiple globs affecting an edge?

gavin-ts: hmm adding 

```d2
(** -> **)[*]: {
	label.near: center
}
```

to the test correctly shows the new error, maybe this isn't needed?

gavin-ts (approved):

d2ir/d2ir.go
gavin-ts: oh wait

```d2
**: {
  label: hi
  label.near: center
}
(** -> **)[*]: {
	label.near: top-center
}
(** -> **)[*]: {
	label.near: invalid
}

x: {
  a -> b
}
```

has the line number of the first valid near instead of the 2nd
```
        d2/testdata/d2compiler/TestCompile2/globs/double-glob-err-val.d2:4:3: invalid "near" field
        d2/testdata/d2compiler/TestCompile2/globs/double-glob-err-val.d2:7:2: invalid "near" field
```

line 7 should be 10 right?

gavin-ts (changes requested):

d2ir/d2ir.go
alixander: gooood catch

gavin-ts (approved):

d2ir/d2ir.go
```diff
@@ -318,11 +318,21 @@ func (f *Field) Copy(newParent Node) Node {
...
 	for i := len(f.References) - 1; i >= 0; i-- {
-		if f.References[i].Primary() && !f.References[i].DueToLazyGlob() {
-			return f.References[i]
+		if f.References[i].Primary() {
+			if f.References[i].DueToLazyGlob() {
+				if lastGlobSet == nil {
+					lastGlobSet = f.References[i]
+				}
```
gavin-ts, line 327: can this become `return f.References[i]`? when does it need to keep iterating?

alixander: i thought it was that case, but i realized the last ref always wins, so you're right, it can become that

PR #1693, opened 2023-10-31
cli: Fix invalid link paths

### LINKED ISSUE #1665: linked layers watch mode bug
I've not confirmed it yet.

```d2
layers: {
    cream: {
        c -> b
    }
}

a -> b
b.link: cream
```

Screen Shot 2023-10-15 at 12 58 49 PM

alixander:
fixes #1665


d2target/d2target.go
```diff
@@ -95,29 +95,24 @@ func (d *Diagram) getBoard(boardPath []string) *Diagram {
 		return d
 	}
 
+	if len(boardPath) < 2 {
+		return nil
```
gavin-ts, line 99: what about this other form? https://github.com/terrastruct/d2/blob/master/d2target/d2target.go#L76-L77
I think this won't find any boards since it is recursive and will hit this nil near the end

alixander: yeah urgh, I can't in good conscience merge this without tests. will take a bit longer, shouldn't have put it up for review so prematurely

e2etests-cli/main_test.go
```diff
@@ -544,6 +550,169 @@ i used to read
...
+					err := tms.Signal(ctx, os.Interrupt)
+					assert.Success(t, err)
+				}()
+
+				// Wait for watch server to spin up and listen
+				urlRE := regexp.MustCompile(`127.0.0.1:([0-9]+)`)
+				watchURL := waitLogs(ctx, stderr, urlRE)
+
+				if watchURL == "" {
+					t.Error(errors.New(string(stderr.Bytes())))
```
gavin-ts, line 581:
```suggestion
					t.Error(errors.New(stderr.String()))
```

e2etests-cli/main_test.go
```diff
@@ -544,6 +550,169 @@ i used to read
...
+				}
+				stderr.Reset()
```
gavin-ts, line 583: hmm it looks like there could be a race between the broadcasting update and this reset+the rest of the test. e.g. adding a time.Sleep(time.Second) before this and logging has

```
    main_test.go:640: success: listening on http://127.0.0.1:50170
        info: compiling index.d2...
        success: successfully compiled index.d2 to index/cream.svg in 52.002ms
        success: successfully compiled index.d2 to index/index.svg in 51.4785ms
        info: broadcasting update to 0 clients
```

I think if you writeFile with an empty file, then connect the client, then writeFile the test d2 after the server is up, it should correctly show the `broadcasting update to 1 client` log avoiding the race on the initial compile.

gavin-ts (commented):
cool tests with the websockets

e2etests-cli/main_test.go
alixander: You're saying that the race results in those extra logs right which won't get reset?

It won't matter, we look for the right thing later (broadcasting to 1 client).

gavin-ts: in this race those logs are present before the reset

alixander: When we make the page request a couple lines down, it will recompile and rebroadcast there.

gavin-ts: oh interesting I didn't expect clicking the link to cause a recompile when it hasn't changed, is that a bug?

gavin-ts (approved):

e2etests-cli/main_test.go
alixander: I never questioned it. I guess the alternative is to read the already-written files.

The only instance I can think of where that goes awry is if the user compiles, moves the files, then refreshes their browser or navigates with their browser. 

Worth considering: https://github.com/terrastruct/d2/issues/1708

PR #1695, opened 2023-11-01
compiler: vars fixes

### LINKED ISSUE #1641: panic using var on object declared before vars
```d2
x1 -> x2

vars: {
  v: {
    style.fill: green
  }
}

# ok
x3: {...${v}}

# panic 
# x1: { ...${v} }

```

alixander:
1. Fix missing spread substitution var
2. Change vars scoping

## For 2:

The compiler compiles the IR first, then resolves substitutions.

The vars stack is built-up incrementally, so a var is only visible if it's been declared before you access it.

The initial IR compilation coalesces the fields, so that x1, though accessing a var after the var's declaration, is seen as being defined before the var's declaration. 

I don't actually think in a declarative language you should be able to change vars mid-way. Changed so that all vars within the same scope are coalesced first.

closes #1641

### DIFF
--- a/d2ir/compile.go
+++ b/d2ir/compile.go
@@ -465,6 +465,7 @@ func (c *compiler) compileMap(dst *Map, ast, scopeAST *d2ast.Map) {
 					Context_: &RefContext{
 						Scope:    ast,
 						ScopeMap: dst,
+						ScopeAST: scopeAST,
 					},
 				}},
 			}
--- a/d2ir/compile.go
+++ b/d2ir/compile.go
@@ -125,6 +125,8 @@ func (c *compiler) compileSubstitutions(m *Map, varsStack []*Map) {
 		if f.Name == "vars" && f.Map() != nil {
 			varsStack = append([]*Map{f.Map()}, varsStack...)
 		}
+	}
+	for _, f := range m.Fields {
 		if f.Primary() != nil {
 			c.resolveSubstitutions(varsStack, f)
 		}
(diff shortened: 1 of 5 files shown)


d2ir/compile.go
```diff
@@ -459,6 +461,12 @@ func (c *compiler) compileMap(dst *Map, ast, scopeAST *d2ast.Map) {
 						Value: []d2ast.InterpolationBox{{Substitution: n.Substitution}},
 					},
 				},
+				References: []*FieldReference{{
+					Context_: &RefContext{
+						Scope:    ast,
+						ScopeMap: dst,
+					},
```
gavin-ts, line 468: does this need ScopeAST similar to compileKey above?

d2ir/compile.go
```diff
@@ -459,6 +461,12 @@ func (c *compiler) compileMap(dst *Map, ast, scopeAST *d2ast.Map) {
 						Value: []d2ast.InterpolationBox{{Substitution: n.Substitution}},
 					},
 				},
+				References: []*FieldReference{{
```
gavin-ts, line 464: out of curiosity why is this needed?

gavin-ts (approved):
tests look good I don't really know this code that well but it looks good

d2ir/compile.go
alixander: Hm it doesn't right now but better to put it for consistency.

d2ir/compile.go
alixander: It was just inconsistent that a Field existed without a ref. It was like an object without a pointer, so other parts of the code that looked for the pointers failed.

PR #1700, opened 2023-11-03
style: allow current color as valid css color

hboomsma:
In SVG images `curentColor` is allowed as a way to explicitly inherit the current color. When setting `currentColor` as fill color on a shape together with a styling class, the class will now be used instead of the theme, which was not possible otherwise. This way you can put theme classes on shapes making them respond to dark mode. This is most useful in combination with custom themes.


alixander (approved):
great, thanks @hboomsma !

PR #1713, opened 2023-11-08
d2ir: fix glob edge with common glob

### LINKED ISSUE #1712: globs: connection creating node when it shouldn't
This shouldn't have created new nodes

```d2
*.*a -> *.*b

container_1: {
  a
}

container_2: {
  b
}
```

Screen Shot 2023-11-08 at 9 08 43 AM

alixander:
closes #1712

before it was creating nodes unnecessarily

Screen Shot 2023-11-08 at 2 59 33 PM


dunkelziffer:
Thanks for the quick fix. I'm already eager to play around with this :D

dunkelziffer:
@alixander I just built this from source. Your fix works for me and seems to cover all my current use cases. However, it seems to not fix globs in general. For example `**.*a -> **.*b` still generates new nodes. Ill make a new issue with some discussions.

Issue #1644, opened 2023-10-05
feat: `label` at board level


alixander:
Especially useful for root board.

Users should be able to rename this.

![Screen Shot 2023-10-05 at 11 43 11 AM](https://github.com/terrastruct/d2/assets/3120367/2bbf6439-bca1-4892-bd92-827175ab8670)

bo-ku-ra:
and #1439

alixander:
make sure to remove nav is root label is empty

https://github.com/terrastruct/d2/issues/1439#issuecomment-1751459357

PR #1725, opened 2023-11-12
cli: add --target flag

### LINKED ISSUE #1707: cli: `--board` flag
This should allow exporting a single board from a multi-board diagram.

```sh
d2 --board root.x index.d2
```

index.d2:

```d2
a -> b
layers: {
  x: {
    c -> d
  }
}
```

### LINKED ISSUE #1727: fix board lookup for layer names with `/`
```d2
"x.a" -> "y / b"
```

object ids can contain `/` so board lookup needs to account for this, passing the path as `[]string` would be better.

If `/` is a valid character in layer names, wouldn't it also break this existing code?

```go
// boardPath comes in the form of "x/layers/z/scenarios/a"
// or "layers/z/scenarios/a"
// or "x/z/a"
func (d *Diagram) GetBoard(boardPath string) *Diagram {
	path := strings.Split(boardPath, string(os.PathSeparator))
	if len(path) == 0 || len(boardPath) == 0 {
		return d
	}
```

If so, it seems like a bug.

_Originally posted by @landmaj in https://github.com/terrastruct/d2/pull/1725#discussion_r1391620205_

landmaj:
closes #1707
closes #1727 

This PR adds `--target` flag which allows exporting a single board from a multi-board diagram.

```d2
a -> b
layers: {
  x: {
    c -> d
  }
}
```

Layer "b" can be exported with one of the following commands:
```bash
d2 --target x index.d2
d2 --target layers.x index.d2
```

EDIT: comment updated to reflect changes after review.


d2cli/main.go
```diff
@@ -114,6 +114,7 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
 	if err != nil {
 		return err
 	}
+	boardFlag := ms.Opts.String("D2_BOARD", "board", "", "", "render a single board from the diagram. E.g. root.scenarios.name to render scenario 'name'. If not provided, the entire diagram is rendered.")
```
alixander, line 117: Could you update the man page with this new flag too?

https://github.com/terrastruct/d2/blob/master/ci/release/template/man/d2.1

d2cli/main.go
```diff
@@ -260,6 +261,11 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
 			centerFlag = nil
 		}
 	}
+	if ms.Env.Getenv("D2_BOARD") == "" {
```
alixander, line 264: Does this need an environment variable counterpart? It feels highly specific to each diagram, not something I'd see anyone wanting to preserve in environment.

d2cli/main.go
```diff
@@ -329,10 +338,22 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
 		return w.run()
 	}
 
+	var singleBoard bool
+	switch {
+	case boardFlag == nil:
+	case *boardFlag == "":
+		boardFlag = nil
+	default:
+		*boardFlag = strings.Replace(*boardFlag, ".", "/", -1)
```
alixander, line 347: Should do some validation on the input first, even though flag instructions are there. I think just checking that `root` is prefix and `/` doesn't exist should be fine

d2cli/main.go
```diff
@@ -398,6 +419,19 @@ func compile(ctx context.Context, ms *xmain.State, plugins []d2plugin.Plugin, la
 	}
 	cancel()
 
+	if boardPath != nil {
+		diagram = diagram.GetBoard(*boardPath)
+		if diagram == nil {
+			return nil, false, fmt.Errorf(`board "%s" not found`, *boardPath)
+		}
+		if singleBoard {
+			diagram.Name = ""
```
alixander, line 428: Why erase the name too?

alixander (changes requested):
I think in the future D2 should allow exporting just a container from a diagram, which can reuse this flag if we call it `--render-target` or something. 

cc @gavin-ts any alt name suggestions?

d2cli/main.go
landmaj: If name would not be erased erased, the board would be saved to a directory named after the source file. And writing to stdout would be broken:

```bash
❯ go run main.go --render-target root.x index.d2 -
success: successfully compiled index.d2 to -/x in 71.135ms
```

With name erased everything works as expected.

d2cli/main.go
landmaj: Sure, I also added this feature to changelog.

gavin-ts:
> I think in the future D2 should allow exporting just a container from a diagram, which can reuse this flag if we call it `--render-target` or something.
> 
> cc @gavin-ts any alt name suggestions?

hmm maybe the target is implicit, I think I would prefer using `d2 --render root.x index.d2` over `d2 --render-target root.x index.d2`

```
--render [target]
Render a single board from multi-board diagram. E.g. root.scenarios.x to render only scenario x. If not provided, the entire diagram is rendered
```

alixander:
I don't think it's implicit. With just a `--render` flag, I'd think maybe that controls render format, or resolution. `--render-target [path]`

landmaj:
It could also be just `--target`. This is how [Docker handles building specific stages](https://docs.docker.com/build/building/multi-stage/#stop-at-a-specific-build-stage).

alixander:
`--target` sounds good to me.

landmaj:
PR updated to use `--target` flag.

d2cli/main.go
gavin-ts: why do we need to always prefix the argument with root, just to trim it out? It feels more like boilerplate than useful info and you have to type it each use.

- `--target root` is the same as not providing the target == `-target ''` so it doesn't seem like a useful option
- `--target root.x` can just be `--target x` (with future object targeting, if there is an object also named `x`, then `--target layers.x` would be necessary (in future object has priority over layer))

Noting that these are object paths and not file paths, but `x -> "y / b"` is valid so a layer name could have a `/`.
I think for validation it should just see if it can find the object.

gavin-ts: oh `--target root` is when you don't want any scenarios

landmaj: There is a difference:

* `--target ''` would render entire diagram, with all layers, scenarios and steps
* `--target root` would render just root and no children

To allow writing to stdout (which is my main motivation for this feature), this distinction is important.

If `/` is a valid character in layer names, wouldn't it also break this existing code?

```go
// boardPath comes in the form of "x/layers/z/scenarios/a"
// or "layers/z/scenarios/a"
// or "x/z/a"
func (d *Diagram) GetBoard(boardPath string) *Diagram {
	path := strings.Split(boardPath, string(os.PathSeparator))
	if len(path) == 0 || len(boardPath) == 0 {
		return d
	}
```

If so, it seems like a bug.

gavin-ts: `d2 index.d2` would render entire diagram with all layers scenarios steps
`d2 --target '' index.d2` would render just root and no children

that code does seem like it has a bug for layer names with `/`

landmaj: Default value is an empty string (because `nil` is not possible), so `d2 --target `` index.d2` is the same as `d2 index.d2`. To render just root and no children, `--target root` is required.

```bash
❯ go run main.go --target root index.d2
success: successfully compiled index.d2 to index.svg in 69.461959ms

❯ go run main.go --target '' index.d2
success: successfully compiled index.d2 to index/x.svg in 70.542209ms
success: successfully compiled index.d2 to index/index.svg in 69.001958ms

❯ go run main.go index.d2   
success: successfully compiled index.d2 to index/x.svg in 71.247833ms
success: successfully compiled index.d2 to index/index.svg in 69.866416ms
```

landmaj: So I would propose:
 - only check if `--target` starts with `root`
 - all characters are allowed

The bug is in an already existing code and I don't know enough about D2 internals to fix it, so I think it's best to open a new issue and fix it there.

BTW is changing all dots to `/` safe?

gavin-ts: hmm maybe `*` should be the default value since it will target everything by default. Then it is possible to detect that the target flag was provided with `d2 --target ' index.d2`.

gavin-ts: I think these will need to be properly parsed, just doing string manipulations isn't exactly safe, but we can make this change when fixing #1727 .
Screenshot 2023-11-13 at 12 12 26 PM

landmaj: `*` would add a lot of confusion, since this, being a default value, would be visible in `--help` output. It's not obvious that you need to override the default value with nothing (empty string) to render root with no children. And also if `*` renders everything, I would expect `x.*` to render `x` and all of its children, which is not the case.

landmaj: IMO `--target root` and `--target root.x` is easier to understand and use, especially since you can already see it in some places, for example PDF output.

```d2
a -> b

layers: {
  x / y: {
    c -> d
  }
}
```

Screenshot 2023-11-13 at 21 29 10

Screenshot 2023-11-13 at 21 29 02

gavin-ts: > `*` would add a lot of confusion, since this, being a default value, would be visible in `--help` output. It's not obvious that you need to override the default value with nothing (empty string) to render root with no children. And also if `*` renders everything, I would expect `x.*` to render `x` and all of its children, which is not the case.

I think `x.*` to render just a specific scenario and its steps sounds like a good feature to eventually have. We can update the description to explain that `'' or root` can be used to target only the root.

gavin-ts: note that the `root` pdf label we be able to be renamed #1644

landmaj: `*` it shall be. I don't want to introduce half-baked feature, so I will add support for `x.*`. 

> `''` or `root` can be used to target only the root

I don't think this is a good idea since `root` is a valid layer name and there could be a conflict. With good documentation `''` should be enough.

I will ping you when changes are ready.

landmaj: @gavin-ts 

I updated the PR. It even works with this diagram now:

```d2
layers: {
  a: {
    layers: {
      "x / y . z": {
        mad
      }
    }
  }
}
```

```bash
❯ go run main.go --target a."\"x / y . z\"" index.d2
success: successfully compiled index.d2 to index.svg in 27.507ms
```

And asterisk allows rendering specific board with all of its children:

```d2
x
layers: {
  a: {
    layers: {
      b: {b}
      c: {c}
    }
  }
}
```

```bash
❯ go run main.go --target=a.* index.d2
success: successfully compiled index.d2 to index/a/b.svg in 60.910375ms
success: successfully compiled index.d2 to index/a/c.svg in 59.784417ms
```

d2cli/main.go
landmaj: It's outdated now, I removed this piece of code.

gavin-ts (approved):
nice! looks good to me

PR #1774, opened 2023-12-12
allow icon for sql_table, class, code/md

### LINKED ISSUE #1558: applying `icon` to sql_table makes it render differently
Screen Shot 2023-08-20 at 9 51 10 AM

even though the icon doesn't show up, adjustments are made for it as if it will

alixander:
### Before

Screen Shot 2023-12-10 at 10 05 07 AM

### After

Screen Shot 2023-12-11 at 3 45 18 PM

Screen Shot 2023-12-11 at 4 12 14 PM

closes #1558


d2layouts/d2elklayout/layout.go
```diff
@@ -374,6 +374,7 @@ func Layout(ctx context.Context, g *d2graph.Graph, opts *ConfigurableOpts) (err
 		if hasTop || hasBottom {
 			padding := parsePadding(elkNodes[obj].LayoutOptions.Padding)
 			if hasTop {
+				// TODO I think this fails to account for a potential inner label of container
```
alixander, line 377: @gavin-ts there is currently a collision between the icon and label in ELK. I believe it's unrelated to my change and that this code has a bug as described. Can you confirm if so or if I'm just missing something.

Screen Shot 2023-12-11 at 4 16 05 PM

gavin-ts: yeah it looks like elk hasn't been updated for this. previously all icon positions would be placed in an inside position by default so this would only be possible with custom set positions. 

not sure why but elk only seems to be adjusting horizontally for icons but not vertically

https://github.com/terrastruct/d2/blob/56a549d7c197f33e4c4fb2f88af6db5dab1d7869/d2layouts/d2elklayout/layout.go#L1125-L1131

also noticed elk layout is still doing some old height manipulations but we already look for intersections with outside labels so these are no longer needed

https://github.com/terrastruct/d2/blob/56a549d7c197f33e4c4fb2f88af6db5dab1d7869/d2layouts/d2elklayout/layout.go#L1184-L1187

https://github.com/terrastruct/d2/blob/56a549d7c197f33e4c4fb2f88af6db5dab1d7869/d2layouts/d2elklayout/layout.go#L1112-L1115



https://github.com/terrastruct/d2/blob/56a549d7c197f33e4c4fb2f88af6db5dab1d7869/d2graph/layout.go#L439-L445

gavin-ts: in the elk `icon_positions` test it appears this is unaccounted for prior to your PR, it also may not be accounted for in dagre if the label were inside

Screenshot 2023-12-11 at 5 38 05 PM

alixander: okay, i'll get in followup then

alixander: @gavin-ts 
are you sure about the "these are no longer needed"? Many tests change when I remove those:

https://github.com/terrastruct/d2/pull/1775/commits/8ce19c551693a40b884b34a9ea94d95a6945e0c7

gavin-ts: in that commit, you also removed the code that made the other 2 unnecessary

alixander: i reverted, there's still changes: https://github.com/terrastruct/d2/pull/1775/commits/705225ee7948d1db9c88f9feb850a1e89371afc6

gavin-ts: https://github.com/terrastruct/d2/commit/8ce19c551693a40b884b34a9ea94d95a6945e0c7#r134821504

looks like we do want the || icon part for this 
Screenshot 2023-12-12 at 10 46 19 AM

gavin-ts: I think we may actually want this adjustment for all outside labels/icons unless we adjust the spacing in another way
https://github.com/terrastruct/d2/commit/8ce19c551693a40b884b34a9ea94d95a6945e0c7#r134822027

to avoid this
Screenshot 2023-12-12 at 10 49 31 AM

alixander: is this captured in https://github.com/terrastruct/d2/pull/1776 or can you make another issue?

gavin-ts: yes this is captured in #1776, all margins are now included in the width/height passed to elk, so elk ends up routing the edge distance to shape + margin, instead of routing to shape then cutting at margin.

PR #1775, opened 2023-12-12
Elk label collision

alixander:


alixander:
@gavin-ts can you take over this? you know more about this

PR #1782, opened 2023-12-18
cli: allow for incomplete AST's to be sent to the vscode extention

BarryNolte:
Instead of returning after a parse that, "fails", the resulting AST is still useful to the vscode extention for autocomplete while a d2 document is built up.


d2cli/main.go
```diff
@@ -457,11 +457,10 @@ func compile(ctx context.Context, ms *xmain.State, plugins []d2plugin.Plugin, fs
 	}
 
 	if os.Getenv("D2_LSP_MODE") == "1" {
-		// only the parse result is needed if running d2 for lsp
+		// only the parse result is needed if running d2 for lsp,
+		// if this, "fails", the AST is still valid and can be sent
+		// to vscode extention
 		ast, err := d2lib.Parse(ctx, string(input), opts)
```
alixander, line 463: this code should attempt to cast the `err` to `d2parser.ParseError` (right now that's the only thing that can be returned by d2lib.Parse, but may not always remain that way), and only if it casts should it be ignored

alixander (changes requested):

d2cli/main.go
BarryNolte: I don't think I gain anything by doing this.  If the type of the error changes, then I'll have to make changes in the language server, and keeping it this way, if the change is made it becomes clear the type changed and new error handling code in the language server has to change.  The fact there is an error is fine and shouldn't be ignored, the ast is valid and the errors are valid.

alixander: I haven't looked at how the client (LSP) works, but wouldn't you want unexpected errors going through stderr instead of stdout? You can never rule out that a program won't panic, and panics go to stderr, so I'd imagine any client would be like, "if unexpected exit status, read stderr and show user diagnostic msg"

BarryNolte: That would be correct.  On the client/server side, I have code to detect if d2 panics, but I'm not reading stderr in that case.  I'll add that.  Otherwise, this pr should be good as-is.

BarryNolte: Added code to d2-vscode to detect the panic case.

alixander (approved):

PR #1801, opened 2024-01-04
D2oracle set imports 2

alixander:
see tests


alixander:
https://github.com/terrastruct/d2/pull/1802

PR #1803, opened 2024-01-05
dynamic dark theme based on CSS class selector

develar:
* introduce `--dark-theme-class`: the CSS class to enable dark mode. When left unset prefers-color-scheme media query is used;
 * add test for combined light and dark themes (original test uses light theme in dark theme test, not clear why);

 When D2-produced SVG file is embedded into some documentation/site framework where dark mode is controller by CSS class, using `prefers-color-scheme` media query leads to inability to switch dark theme if needed. [Writerside issue](https://youtrack.jetbrains.com/issue/WRS-3828/D2-diagram-does-not-adapt-to-dark-theme-automatically). Same for mkdocs and so on. [IntelliJ D2 plugin](https://plugins.jetbrains.com/plugin/20630-d2) also will use a new flag.

 Media Query Level 5, where it is possible to use some class as a condition, is not supported by any modern browser.

https://github.com/terrastruct/d2/assets/350686/ddcb6fcb-b5ee-4637-8145-f230bcbadc51


d2cli/main.go
```diff
@@ -87,6 +87,7 @@ func Run(ctx context.Context, ms *xmain.State) (err error) {
 	if err != nil {
 		return err
 	}
+	darkThemeClassFlag := ms.Opts.String("D2_DARK_THEME_CLASS", "dark-theme-class", "", "", "the CSS class to enable dark mode. When left unset prefers-color-scheme media query is used. See https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme.")
```
alixander, line 90: Can you include this in the man page as well? https://github.com/terrastruct/d2/blob/master/ci/release/template/man/d2.1

alixander (changes requested):
Thanks for this PR. Let me understand the context a bit more.

1. Certain websites add a CSS class to all images when dark mode is turned on.
2. This flag makes it so that this CSS class flows down into D2's CSS to control dark mode.

The downside is obviously that the image is no longer portable -- it's generated to be on a certain website.

It seems that apps/sites that do this type of dark mode configuration know that this is a limitation on most images and just have you specify two images, e.g.:
1. https://docusaurus.io/docs/markdown-features/assets#github-style-themed-images
2. https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/

I guess rendering an SVG for a particular website and thus having it not be portable for others is fine. I can't think of anything better.

----------------------

I think this flag should be either class or ID. If it starts with `.something`, it's a class. If it's `#something`, it's ID.

develar:
> 1. Certain websites add a CSS class to all images when dark mode is turned on.

The "dark" class is technically applied to the `html` or other top-level elements. See, for example, the [Tailwind Documentation template](https://syntax.tailwindui.com/) — it applies class `dark` to `html`.  (`<html lang="en" class="h-full antialiased __variable_e66fe9 __variable_b436a8 dark" style="color-scheme: dark;">`)

> The downside is obviously that the image is no longer portable -- it's generated to be on a certain website.
> just have you specify two images

Writerside also supports this, enabling us to generate or provide two different static images, but I wouldn't say it is a downside.

When rendering SVGs into bitmap format, the user loses the ability to copy text from the diagrams. To me, this is a valuable feature. Additionally, we should also consider scaling — Retina screens, for example, require 2x resolution images to provide crisp, detailed visuals.

alixander:
Cool and lastly, can you sign your PR please @develar ? We have it turned on as an organization-wide setting.

develar:
> Cool and lastly, can you sign your PR please 

Sign Git commits? Done.

d2renderers/d2svg/d2svg.go
```diff
@@ -2018,96 +2021,108 @@ func Render(diagram *d2target.Diagram, opts *RenderOpts) ([]byte, error) {
...
+	classPrefix := "." + diagramHash
+	out, err := singleThemeRulesets(classPrefix, *themeID, overrides)
 	if err != nil {
 		return "", err
 	}
 
 	if darkThemeID != nil {
-		darkOut, err := singleThemeRulesets(diagramHash, *darkThemeID, darkOverrides)
+		hasThemeClass := len(darkThemeClass) != 0
+		if hasThemeClass {
```
alixander, line 2037: just `if darkThemeClass != ""`

d2renderers/d2svg/d2svg.go
```diff
@@ -2018,96 +2021,108 @@ func Render(diagram *d2target.Diagram, opts *RenderOpts) ([]byte, error) {
...
 			property, property, theme.Colors.AB4,
-			diagramHash,
+			classPrefix,
 			property, property, theme.Colors.AB5,
 		)
 	}
 
 	// Appendix
-	out += fmt.Sprintf(".appendix text.text{fill:%s}", theme.Colors.Neutrals.N1)
+	out += fmt.Sprintf("%s .appendix text.text{fill:%s}", classPrefix, theme.Colors.Neutrals.N1)
```
alixander, line 2121: i don't see why this changes, or any of the lines below which didn't include the diagramHash

d2renderers/d2svg/dark_theme/dark_theme_test.go
alixander, line 1: the changes to this test don't test the new flag though?

alixander (changes requested):
Forgive the delay here. This feature looks valuable, if you're still interested in pursuing it.

develar:
I’ve started working on this PR again. I reviewed the comment above, and it does look like a better approach. It’s not exactly what I’d like to see in the long run, since ideally we should be able to use external CSS to reduce SVG size—but that’s a separate topic, as it involves different scenarios with their own trade-offs.

PR #1857, opened 2024-03-06
fix edge glob bug

alixander:
also fixes some d2oracle functions when globs are present

also upgrades go version 1.20 -> 1.22


d2ir/d2ir.go
```diff
@@ -1094,7 +1094,7 @@ func (m *Map) getEdges(eid *EdgeID, refctx *RefContext, gctx *globContext, ea *[
 					}
 					gctx.appliedEdges[ks] = struct{}{}
 				}
-				*ea = append(*ea, ea2...)
+				*ea = append(*ea, e)
```
alixander, line 1097: this bug was causing globs to be applied multiple times

PR #1872, opened 2024-03-18
fix style.fill for markdown

### LINKED ISSUE #1805: Markdown style regression on v.0.2.1?
### In short
I think v0.2.1 broke something on markdown styling, fill and stroke on markdown text no longer work as expected. Doesn't work on latest version either.
I reviewed changelog for v0.2.1, and don't think that this is an intended change.

I'm new to D2, so could be just a misundestanding on my part, but could you please check?

### Details
PR #460 introduced style.fill and style.stroke for Markdown text in v.0.1.2.
E2E test added in the same PR: https://github.com/terrastruct/d2/commit/d28aed6184d163a45922073d603b39c6af08c137
https://github.com/terrastruct/d2/blob/d28aed6184d163a45922073d603b39c6af08c137/e2etests/stable_test.go#L1491-L1515

Using the above D2 code from the test, I am unable to get expected results on v0.6.3..v0.2.1.
Using v0.2.0 I do get the expected result.

### Demonstration
Using the same test input (above) and ELK layout, results by version:
**v0.1.2 (feature introduced)**
test_v0 1 2

**v0.2.0 (feature last observed working)**
test_v0 2 0

**v0.2.1 (feature stops working)**
test_v0 2 1

**v0.6.3 (latest version)**
test_v0 6 3

GavinNishizawa:
## Summary

Fixes a regression where `style.fill` would not apply for text shapes with markdown.

## Details
- fixes #1805 
- updates old test to use `style.font-color` instead of `style.stroke` for coloring text
- markdown background color will now be set with `style.fill`


d2renderers/d2svg/d2svg.go
```diff
@@ -1370,7 +1370,9 @@ func drawShape(writer, appendixWriter io.Writer, diagramHash string, targetShape
 			if targetShape.FontSize != textmeasure.MarkdownFontSize {
 				styles = append(styles, fmt.Sprintf("font-size:%vpx", targetShape.FontSize))
 			}
-
+			if targetShape.Fill != "" && targetShape.Fill != "transparent" {
```
alixander, line 1373: i mean i guess if they specified transparent, we should let them?

GavinNishizawa: yes it will already be a transparent div by default when rendered, and it is just that "transparent" is the default value for text including markdown (here https://github.com/terrastruct/d2/blob/master/d2exporter/export.go#L92). This avoids unnecessary style attributes in the rendered result in all the svgs.

e.g. in https://github.com/GavinNishizawa/d2/blob/c76bc9f368be9763cca191e92b486511689d5063/e2etests/testdata/stable/markdown_stroke_fill/elk/sketch.exp.svg#L839

this adds a background color
```xhtml
<div xmlns="http://www.w3.org/1999/xhtml" class="md" style="background-color:#CEEDEE;color:red"><p>they did it in style</p>
</div>
```
vs
this already has a transparent background
```xhtml
<div xmlns="http://www.w3.org/1999/xhtml" class="md" style="color:darkorange"><h1>a header</h1>
<p>a line of text and an</p>
<pre><code>{
	indented: "block",
	of: "json",
}
</code></pre>
<p>walk into a bar.</p>
</div>
```

ci/release/changelogs/next.md
```diff
@@ -12,3 +12,4 @@
 - Fixes styles in connections not overriding styles set by globs [#1857](https://github.com/terrastruct/d2/pull/1857)
 - Fixes `null` being set on a nested shape not working in certain cases when connections also pointed to that shape [#1830](https://github.com/terrastruct/d2/pull/1830)
 - Fixes edge case of bad import syntax crashing using d2 as a library [#1829](https://github.com/terrastruct/d2/pull/1829)
+- Fixes `style.fill` not applying to markdown [#1872](https://github.com/terrastruct/d2/pull/1872)
```
alixander, line 15: it's `style.font-color` right?

GavinNishizawa: `style.font-color` was just updating the test case (we used `style.stroke` for font-color when the test was written). The functional change I added was to restore the part which applies `style.fill` to set the background color.

alixander (approved):

PR #1874, opened 2024-03-20
d2oracle: make get writeable refs public

alixander:


e2etests-cli/testdata/TestCLI_E2E/empty-base.exp.svg
alixander, line 1: i don't understand how this changed

PR #1900, opened 2024-04-10
allow minimum dimensions to be less than label dimensions

alixander:
It's especially unreasonable that you can't size a person shape to be smaller than the label.

## Before

Screenshot 2024-04-10 at 8 46 36 AM

## After

Screenshot 2024-04-10 at 8 50 32 AM


alixander:
https://github.com/terrastruct/d2/pull/1901

Issue #1898, opened 2024-04-10
about handling '&' in markdown


bo-ku-ra:
API error
https://play.d2lang.com/?l=&script=BMCxDgIhDADQna9obnC8xhsJ4LcQWsHEUgWSLnz8PWFRD1vIAYQMbfA7Hm2t3_SIZnZW1frls6jg5DxKe_0jXY_OZp9OavF5JLoC5uS2uwMAAP__&

url is difficult to handle.
https://play.d2lang.com/?l=&script=BMAxDsMgDADAnVdYGTrGasY0pG-xYhcq1ZgCkhcen1NR22EqB4CDIDf5xCWPUfuO6O5rMks_WS9T7ELtyu9_5O1BWl9F3L-FzeNzOXk7kM4wwx0AAP__&

alixander:
Notes:

`<a href="asdf&asdf">b</a>` doesn't work because it's raw html which is rendered as-is without escapes (in order for the html to work), but the ampersand gets interpreted as an XML entity in SVG (like `&lsqou;`).

https://github.com/yuin/goldmark is so extensible i'm sure there's a way to add a parsing extension that replaces `&` with `&amp;` or make urls safe when rendering unsafe html. Not high priority to do right now

PR #1934, opened 2024-05-03
fix: add url escaping to markdown processing #1898

jeengland:
Fix for issue #1898

I saw in the issue thread it was suggested to create a goldmark extension, which I'm still open to doing but I wanted to suggest a simpler regex based solution in this PR. Works on the md from the issue as well as a few other test cases:
```
memo: |md
  <a href="https://www.google.com/search?q=d2&newwindow=1">d2</a>
|
```

### DIFF
--- a/lib/textmeasure/markdown.go
+++ b/lib/textmeasure/markdown.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"fmt"
 	"math"
-	"net/url"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


lib/textmeasure/markdown.go
```diff
@@ -78,9 +80,31 @@ func HeaderToFontSize(baseFontSize int, header string) int {
 	return 0
 }
 
+func preProcessMarkdown(m string) string {
+	// This is a simplified pattern; a more robust one would be complex
+	urlRegex := regexp.MustCompile(`\b((?:https?|ftp)://[^\s"']+&\S*)\b`)
+
+	output := urlRegex.ReplaceAllStringFunc(m, func(urlStr string) string {
+		// Attempt to parse
```
alixander, line 88: unnecessary comment

lib/textmeasure/markdown.go
```diff
@@ -78,9 +80,31 @@ func HeaderToFontSize(baseFontSize int, header string) int {
 	return 0
 }
 
+func preProcessMarkdown(m string) string {
+	// This is a simplified pattern; a more robust one would be complex
+	urlRegex := regexp.MustCompile(`\b((?:https?|ftp)://[^\s"']+&\S*)\b`)
```
alixander, line 85: I think we want to add `href=` and `src=` as a condition in this regex. Leave the links that are just text (and not hyperlinked) alone.

alixander (changes requested):
That should work for now.

https://github.com/terrastruct/d2/blob/master/d2compiler/compile_test.go

lib/textmeasure/markdown.go
jeengland: I didn't think of that, great catch!

jeengland:
> Reference

Looking at adding a test now, should I be adding a test within the `TestCompile` block? Also when I do do that, it prompts me to rerun the test with $TA=1 and it edits a bunch of files. Is that correct or am I doing something incorrectly?

alixander:
@jeengland Yeah that's right

`TA=1 ./ci/test.sh ./d2compiler` to rerun compiler tests and accept the results. You don't need to verify the AST for this test because you're just checking it compiles at all.

alixander:
Hi @jeengland are you still interested in pursuing this PR? If not I can push to this branch, it's just the test remaining

jeengland:
Hi @alixander sorry about the delay, I had an uptick in obligations this last few weeks. I'll make this a priority to look at tomorrow!

Issue #654, opened 2023-01-12
animate keyword for bidirectional connections


alixander:
right now the bidirectional connections (arrowheads on both ends) just animate from source to target. It might be cool to add an invisible element in the middle of the connection and then animate in opposite directions.

![sketch exp (10)](https://user-images.githubusercontent.com/3120367/212159821-aa7ce4fc-286b-4fee-b3f6-ffdb0a188ea6.svg)

<img width="664" alt="Screen Shot 2023-01-12 at 2 25 56 PM" src="https://user-images.githubusercontent.com/3120367/212194151-065bf465-010f-48e1-9496-bb5cd8b89f0e.png">

bo-ku-ra:
scrolling;
<-> : From the center to each end ■<<<<<>>>>>■ (direction of each arrow)
-- : From each end to the center 　■>>>>><<<<<■ (handshake image)

if it's one way, i have a problem maybe.
i want to make it right or down direction scrolling by design, not as semantics.
when editing, top and bottom, right and left are switched frequently.

alixander:
i don't understand what you mean by scrolling, or handshake, or top/bottom

bo-ku-ra:
https://github.com/terrastruct/d2/issues/1936#issuecomment-2103588340

PR #1939, opened 2024-05-14
fix: makes bidirectional animations correct

### LINKED ISSUE #1936: bidirectional connection animation
Right now `animated: true` does nothing to connections where it's double-ended
- `--`
- `<->`

The animation can bounce back and forth in these cases.

-------

Context:

Screenshot 2024-05-08 at 9 49 14 AM

23danielsuh:
fixes #1936, #654

https://github.com/terrastruct/d2/assets/63919918/86b56fc9-4f34-4e64-8770-0f6bba9c6fd0

Makes <-> and -- have correct animation. I edit the path string instead of the points due to the rounded corners becoming sharp when using S curves (if you have a path from b -> a and b -> c, it is not trivial to round the corner b). Bezier curve segmentation is not perfect due to the parameterization of a bezier curve being inaccurate, but is good enough.


e2etests/testdata/txtar.txt
```diff
@@ -214,6 +214,13 @@ ok: {
   dog1 -> dog3
 }
 
+-- bidirectional_connection_animation --
+a <-> b: {style.animated: true}
+a <-> c: {style.animated: true}
+a <-> d: {style.animated: true}
+a <-> e: {style.animated: true}
+f <-> g: {style.animated: true}
```
alixander, line 222: can we add a self loop as a test too? `x <-> x`

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,191 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
 	)
 }
 
+// Given control points p1, p2, p3, p4, calculate the segment of this bezier curve from t0 -> t1 where {0 <= t0 < t1 <= 1}
+func bezierCurveSegment(p1, p2, p3, p4 *geo.Point, t0, t1 float64) (geo.Point, geo.Point, geo.Point, geo.Point) {
```
alixander, line 499: Let's put these in `lib/svg`, either in `path.go` or a new `bezier.go`

alixander, line 499: Can you leave the reference to the algorithm in the comment?

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,191 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+			increment = 7
+		case "S":
+			x, _ = strconv.ParseFloat(pathData[i+3], 64)
+			y, _ = strconv.ParseFloat(pathData[i+4], 64)
+
+			pathLength += geo.EuclideanDistance(prevPosition.X, prevPosition.Y, x, y)
+
+			increment = 5
+		default:
+			panic(fmt.Sprintf("unknown svg path command \"%s\"", pathData[i]))
```
alixander, line 581: should never panic. return error and handle please

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,191 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+		}
+
+		prevPosition = geo.Point{X: x, Y: y}
+		i += increment
```
alixander, line 585: why's this not just in the for loop?

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,191 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+		*path += fmt.Sprintf("L %s %s ", pathData[offsetIdx+1], pathData[offsetIdx+2])
+		increment = 3
+	case "C":
+		*path += fmt.Sprintf("C %s %s %s %s %s %s ", pathData[offsetIdx+1], pathData[offsetIdx+2], pathData[offsetIdx+3], pathData[offsetIdx+4], pathData[offsetIdx+5], pathData[offsetIdx+6])
+		increment = 7
+	case "S":
+		*path += fmt.Sprintf("S %s %s %s %s ", pathData[offsetIdx+1], pathData[offsetIdx+2], pathData[offsetIdx+3], pathData[offsetIdx+4])
+		increment = 5
+	default:
+		panic(fmt.Sprintf("unknown svg path command \"%s\"", pathData[offsetIdx]))
```
alixander, line 540: same comment about panic

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,191 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+	q4 := geo.Point{
+		X: (u1*u1*u1)*p1.X + (3*t1*u1*u1)*p2.X + (3*t1*t1*u1)*p3.X + t1*t1*t1*p4.X,
+		Y: (u1*u1*u1)*p1.Y + (3*t1*u1*u1)*p2.Y + (3*t1*t1*u1)*p3.Y + t1*t1*t1*p4.Y,
+	}
+
+	return q1, q2, q3, q4
+}
+
+// Adds a certain line/curve to an existing SVG path string. offsetIdx and pathData provides the points needed
+func addToPath(path *string, pathType *string, offsetIdx int, pathData []string) int {
```
alixander, line 523: Make this function a pure function by just producing a string, and then the caller can append as needed.

It also forces a better name

alixander (changes requested):
the tests outputs look great! Can you please add to `next.md` as well?

Also I'm not sure if you want to implement this for the sketch renderer (`d2renderers/d2sketch`) as well. I don't know what it entails, you're welcome to take a look or just open another issue after this.

d2renderers/d2svg/d2svg.go
23danielsuh: Sorry, I'm not really sure what you mean by this. What should be in the for loop?

alixander: Above your for loop is `for i := 0; i < len(pathData); {`

Why not `for i := 0; i < len(pathData); i += increment {` which is a standard 3-part for loop

23danielsuh: Got it, makes sense.

23danielsuh:
I'll create an issue for the sketch renderer soon, I'll take a look at how much work it is and decide if I want to tackle it. Thanks for all the comments!

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,194 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+		if pastHalf {
+			curPath, err := getSVGPathString(&path2, &pathData[i], i, pathData)
+			path2 += curPath
+
+			if err != nil {
+				return "", "", err
+			}
+
+		} else if sumPathLens+curPathLen < pathLen*percentage {
+			curPath, err := getSVGPathString(&path2, &pathData[i], i, pathData)
```
alixander, line 630: path1?

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,194 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+
+			pastHalf = true
+		}
+
+		incr, err := getPathStringIncrement(&pathData[i])
+		increment = incr
+
+		if err != nil {
+			return "", "", err
+		}
```
alixander, line 678: check err before using value

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,194 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+
+		prevPosition = geo.Point{X: x, Y: y}
+
+		incr, err := getPathStringIncrement(&pathData[i])
+		increment = incr
+
+		if err != nil {
+			return 0, err
+		}
```
alixander, line 575: same comment about err check

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,194 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
 	)
 }
 
+// Gets a certain line/curve's SVG path string. offsetIdx and pathData provides the points needed
+func getSVGPathString(path *string, pathType *string, offsetIdx int, pathData []string) (string, error) {
+	var returnPath string
+
+	switch *pathType {
+	case "M":
+		returnPath = fmt.Sprintf("M %s %s ", pathData[offsetIdx+1], pathData[offsetIdx+2])
```
alixander, line 504: can just return in each case

alixander (changes requested):

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,185 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+		return fmt.Sprintf("C %s %s %s %s %s %s ", pathData[offsetIdx+1], pathData[offsetIdx+2], pathData[offsetIdx+3], pathData[offsetIdx+4], pathData[offsetIdx+5], pathData[offsetIdx+6]), nil
+	case "S":
+		return fmt.Sprintf("S %s %s %s %s ", pathData[offsetIdx+1], pathData[offsetIdx+2], pathData[offsetIdx+3], pathData[offsetIdx+4]), nil
+	default:
+		return "", fmt.Errorf("unknown svg path command \"%s\"", pathData[offsetIdx])
+	}
+}
+
+// Gets how much to increment by on an SVG string to get to the next path command
+func getPathStringIncrement(pathType *string) (int, error) {
```
alixander, line 515: why are you taking the pointer?

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,185 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
 	)
 }
 
+// Gets a certain line/curve's SVG path string. offsetIdx and pathData provides the points needed
+func getSVGPathString(pathType *string, offsetIdx int, pathData []string) (string, error) {
```
alixander, line 499: no pointer

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,185 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+	}
+}
+
+// Gets how much to increment by on an SVG string to get to the next path command
+func getPathStringIncrement(pathType *string) (int, error) {
+	var increment int
+
+	switch *pathType {
+	case "M":
+		increment = 3
```
alixander, line 520: can just return

d2renderers/d2svg/d2svg.go
```diff
@@ -494,6 +495,185 @@ func makeLabelMask(labelTL *geo.Point, width, height int, opacity float64) strin
...
+		}
+
+		curPath, err := getSVGPathString(&pathData[i], i, pathData)
+		if err != nil {
+			return "", "", err
+		}
+
+		if pastHalf {
+			path2 += curPath
+		} else if sumPathLens+curPathLen < pathLen*percentage {
```
alixander, line 625: is it possible to not be either these? i thought it's either pastHalf or not pastHalf?

alixander (changes requested):

d2renderers/d2svg/d2svg.go
23danielsuh: Sorry, I'm used to languages that copy the entire string so I thought it was more efficient.

Issue #1828, opened 2024-02-04
`link.icon`


alixander:
can set to `none` to not show any. but it'd be cool to have the github link have a github icon

https://github.com/terrastruct/d2/issues/1186

bo-ku-ra:
>can set to none to not show any. 

please show me this d2 source.

bo-ku-ra:
#589 
#576

Issue #1653, opened 2023-10-08
Container Icon in one of the corner


ppalka-lingaro:
Is there any way to display icon in corner with lower gap than it's by default

![image](https://github.com/terrastruct/d2/assets/134058357/a7af7d32-3a48-4bf8-bc3c-7e0b5452a362)

alixander:
did you want it on the corner? like this?

<img width="316" alt="Screen Shot 2023-10-08 at 12 31 43 PM" src="https://github.com/terrastruct/d2/assets/3120367/da845cb6-f0be-4e5d-8d9f-7772314f1609">

Palkers76:
Yes top left , This time, next time Maybe top right
Or Maybe outside of the box.

PS. This is my other account.

alixander:
Outside the box, there are keywords for that.

Maybe we should add the following keywords:

`border-x` for x in every position keyword value.

Low priority, but I'll keep this open as a post-1.0 feature

Palkers76:
How about image in corner?

ppalka-lingaro:
on similar note
is there a way to force engine to align all containers on same level ?
I have tried to use     width: 200, height: 700 but only a little better

![image](https://github.com/terrastruct/d2/assets/134058357/4a8a0510-47e8-49f3-8744-0d23c2dc4b3f)

[...]

PR #1961, opened 2024-06-01
feature: link icon on shapes can now be custom

danielsuh05:
Fixes Issue #1828

I didn't format the option like the issue said (`link.icon`) because I was a bit confused by `compositeReservedKeyword` and some of the other things in `compile.go`. If you want it to be like the issue said instead of how I did it, I can look back again.

Screenshot 2024-06-01 at 16 59 35


alixander (changes requested):
1. it should still be circular, like the plain link icon. please use a circular svg as the test example too (github has one)
2. yes, it needs to be `link.icon`. it allows for further expansion down the road if needed, instead of flattening out into a bunch of new keywords. related: https://github.com/terrastruct/d2/issues/1653#issuecomment-2143312490

danielsuh05:
So you want something like this?
```
x: hello {
    link: https://github.com {
        icon: https://github.com/_____
    }
}
```

PR #2065, opened 2024-08-23
Add new dark theme: Dark berry blue

ppenguin:
The number of built-in dark themes is lacking compared to the light themes.
This PR aims to be a first step to alleviate this, in submitting a dark variant of "Mixed Berry Blue".

As a side note: it would be logical to make the IDs of themes a bit more symmetrical: since the ID category seems to be `200 <= DarkThemeID < 300`, it is a logical expectation that a dark variant of an existing light theme is `LightID + 200`?

It would be a reasonable goal to work toward "dark theme coverage" in this sense too?

### DIFF
--- a/d2themes/d2themescatalog/dark_berry_blue.go
+++ b/d2themes/d2themescatalog/dark_berry_blue.go
@@ -1,0 +1,25 @@
+package d2themescatalog
+
+import "oss.terrastruct.com/d2/d2themes"
+
+var DarkBerryBlue = d2themes.Theme{
+	ID:   205,
+	Name: "Dark berry blue",
+	Colors: d2themes.ColorPalette{
+		Neutrals: d2themes.DarkNeutral,
+
+		B1: "#E5F3FF",
+		B2: "#BCDDFB",
+		B3: "#77AFE3",
+		B4: "#2353DD",
+		B5: "#0F36A7",
+		B6: "#102566",
+
+		AA2: "#DACEFB",
+		AA4: "#A162D3",
+		AA5: "#5639A5",
+
+		AB4: "#FFDEF1",
+		AB5: "#EA99C6",
+	},
+}
--- a/d2themes/d2themescatalog/catalog.go
+++ b/d2themes/d2themescatalog/catalog.go
@@ -30,6 +30,7 @@ var LightCatalog = []d2themes.Theme{
 var DarkCatalog = []d2themes.Theme{
 	DarkMauve,
 	DarkFlagshipTerrastruct,
+    DarkBerryBlue,
 }
 
 func Find(id int64) d2themes.Theme {


alixander:
@ppenguin Thanks for this PR! Can you screenshot what it looks like?

ppenguin:
> @ppenguin Thanks for this PR! Can you screenshot what it looks like?

Here's a screenshot:
themex-205

> And no the goal isn't to have a dark theme equivalent to every light theme. There's very [few themes]
> ...
> so they're kind of just two different themes. D2 should categorize the two separately -- mirrors of light/dark themes are fine as long as they make sense and look good, but no need to have mirrors for every theme.

Makes sense, no need to go overboard, it's just that the choice of "tones" felt kind of incomplete, from gut feeling  I think a few more would be nice, this one would go into the category "dark pastel blue/purple".

[This branch](https://github.com/ppenguin/d2/tree/ppdev/testdata/examples) has a handy build script to (re)generate an example for all available themes.

d2themes/d2themescatalog/dark_berry_blue.go
```diff
@@ -0,0 +1,25 @@
+package d2themescatalog
+
+import "oss.terrastruct.com/d2/d2themes"
+
+var DarkBerryBlue = d2themes.Theme{
+	ID:   205,
```
alixander, line 6: This makes sense for mirroring themes but I think it may be a confusing precedent now that we agree the themes shouldn't be mirrored. Can you change this to just the next sequential dark theme ID? (202)

alixander (changes requested):
Looks good! Can you please add a line to changelog (next.md) of the new theme too (under improvements)?

alixander:
@ppenguin Nice build script! I think that'd be useful as a dev tool in the d2themes package if you want to make a separate PR to help include that. The hardcoded theme value in the d2 file used should be removed though (https://github.com/ppenguin/d2/blob/ppdev/testdata/examples/themex.d2#L6).

No obligation to of course, will keep this in mind and get to it myself down the road if not.

d2themes/d2themescatalog/dark_berry_blue.go
ppenguin: IMHO "agreeing not to mirror **all** themes" is quite different from "mirroring **some** themes", and that the DX/UX of using this numbering scheme for themes that **are** mirrored is arguably better (the fact that category number groups are used does already mean that the numbers are not contiguous, so sequential numbering is a moot point). Instead I'd suggest to change the ID of the "flagship" theme to `203` and the world will be in order again :wink: 

(Note that it might even have advantages down the road for people who might want to programatically set their own "corporate identity" theme automatically to a user's preference).

alixander: Actually I was wrong in my original comment about why we shouldn't mirror IDs. There shouldn't be any mirroring at all. D2's theme IDs are blocks of continuous ranges, where each range is a category of themes.

Starting from 0 they are cool.
Starting from 100 they are warm.
Starting from 200 they are dark.

Since it's not binary, mirroring doesn't even make sense. It's very possible 300s start up as something else entirely, like "3d", or "minimal".

In other words, it's not broken up like `x = light` and `20x = dark` where you can make a case for mirroring.

----

Sequential numbering is not moot just because it's discontinuous between groups.

ppenguin: I don't see why we couldn't have both, it's a perceived UX bonus at no extra cost (since "faux contiguous range" has no added value at all). But let's agree to disagree, not really worth haggeling over...

alixander (changes requested):
Please forgive the delay, this looks good, would you mind doing a rebase and updating the next.md? Thank you for this.

PR #2079, opened 2024-08-29
Add build script to make an (svg) demo of all available themes

ppenguin:
As [requested](https://github.com/terrastruct/d2/pull/2065#issuecomment-2315450410)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,3 +21,7 @@ test: fmt
 .PHONY: race
 race: fmt
 	prefix "$@" ./ci/test.sh --race ./...
+
+.PHONY: themesdemo
+themesdemo:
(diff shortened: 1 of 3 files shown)


ppenguin:
@alixander I revived this while checking #2065 and integrated it better (so it's basically just one extra rule to the `Makefile` now)
Maybe we can integrate it like this...

BTW: in my env `prefix` is not found, so the `Makefile` doesn't work for testing. What is `prefix` and what does it do?

alixander (changes requested):
Also apologies for the delay on this. 

Instead of outputting in `testdata/examples`, let's put it in a new `d2themes/samples` directory.

PR #2130, opened 2024-09-30
ci daily fix

### LINKED ISSUE #2087: fix ci daily
`./ci/test.sh --race ./...` fails due to race conditions in cli tests. They're almost certainly innocuous, but we should get that fixed up.

https://github.com/terrastruct/d2/actions/runs/10755823241/job/29828005682

daveads:
Fix for https://github.com/terrastruct/d2/issues/2087

/fix #2087 
/calm #2087 

Fix race condition in HTTP server handling

- Implement safeServer to wrap http.Server
- Modify SafeServe function to use safeServer
- Update goServe method with new SafeServe implementation
- Add synchronization to prevent concurrent resource access

Resolves race condition in xhttp.Serve(), fixing intermittent

New implementation provides better server lifecycle control and thread safety.


daveads:
Test branch... result

https://github.com/daveads/d2/actions/runs/11112115344/job/30873521873

daveads:
@alixander check it out..

alixander (changes requested):
Thank you for the PR.

I would like to continue to use https://github.com/terrastruct/util-go/blob/master/xhttp/serve.go . If the issue is part of that, can you submit a PR to that instead? It's just reusable library for Terrastruct's projects.

daveads:
> Thank you for the PR.
> 
> I would like to continue to use https://github.com/terrastruct/util-go/blob/master/xhttp/serve.go . If the issue is part of that, can you submit a PR to that instead? It's just reusable library for Terrastruct's projects.

hmm let me take a quick look at it.

PR #2139, opened 2024-10-05
cli: Force `d2 fmt` operate when the input file is stdin

### LINKED ISSUE #1577: fmt from stdout should always output the formatted diagram
I want to use d2 to format on save in helix but as it outputs nothing if the doc is already formatted it can completely delete the diagram if I am not careful.

akirak:
This PR fixes #1577.

The change is obvious. Previously, `d2 fmt` CLI command was no-op when the content is unchanged. However, this check should be skipped if the input file is stdin–because editor integrations would use it for on-the-fly reformatting.

### DIFF
--- a/d2cli/fmt.go
+++ b/d2cli/fmt.go
@@ -42,7 +42,7 @@ func fmtCmd(ctx context.Context, ms *xmain.State) (err error) {
 		}
 
 		output := []byte(d2format.Format(m))
-		if !bytes.Equal(output, input) {
+		if !bytes.Equal(output, input) || inputPath == "-" {
 			if err := ms.WritePath(inputPath, output); err != nil {
 				return err
 			}


alixander (commented):
Thank you for the PR but I don't think stdin should be treated any differently.

https://github.com/terrastruct/d2/issues/1577#issuecomment-2395105203

Following what gofmt would do, `d2 fmt ...` should output to stdout and `d2 fmt -w ...` should overwrite the files.

akirak:
> I don't think stdin should be treated any differently.

> We should just conform to what https://pkg.go.dev/cmd/gofmt does.

Is this implemented even now? From the document on `gofmt`, you see the following description:

> Without an explicit path, it processes the standard input. 

but `d2 fmt` requires at least one file argument, as seen in the below example:

> 01:40 $ d2 fmt < sample.d2
err: failed to fmt: bad usage: fmt must be passed at least one file to be formatted

The definition of conformance looks vague here.

akirak:
Okay, I think `d2 fmt` should first support zero arguments. Closing this PR anyway.

PR #2162, opened 2024-10-16
doc: Add Python pandoc filter as reference

fliiiix:
I needed a pandoc filter and didn't want to use typescript.


alixander:
Hey @fliiiix it looks like just the commit signing, if you wouldn't mind.

fliiiix:
Should be signed now :)

PR #2274, opened 2025-01-11
d2renderers/d2svg: add IDAttrVal to RenderOpts struct

### LINKED ISSUE #2273: Writing multiple SVGs to an HTML file results in invalid HTML
The `id` attribute value is hardcoded (`d2-svg`), and you can't have multiple HTML elements with the same `id` attribute value. We should be able to include the `id` attribute value in the `d2svg.RenderOpts` struct. If we don't specify the `id` attribute value, fallback to `d2-svg` or `d2-svg-trulyUniqueHashGoesHere` as described below. That should be backward compatible.

Or get rid of it completely.

jmooring:
Closes #2273

### DIFF
--- a/d2renderers/d2svg/d2svg.go
+++ b/d2renderers/d2svg/d2svg.go
@@ -10,10 +10,9 @@ import (
 	"hash/fnv"
 	"html"
 	"io"
+	"math"
 	"sort"
 	"strings"
-
-	"math"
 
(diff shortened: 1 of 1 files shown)


jmooring:
See <https://github.com/terrastruct/d2/issues/2273#issuecomment-2585205118>.

Issue #2276, opened 2025-01-11
Wasm library


nazar-pc:
Would be great to have an official release of WebAssembly, it would make tools in other languages like https://github.com/danieleades/mdbook-d2 not require CLI installation as they could embed WASM version of the library + WebAssembly interpreter/compiler (there is sufficient selection of those for Rust) for more seamless experience.

alixander:
https://github.com/terrastruct/d2/pull/2278#issuecomment-2794963426

bugsbugsbux:
please give us a wasm release

Issue #2272, opened 2025-01-09
Linkage errors in D2 wasm


orange-buffalo:
Hi there,
I am trying to integrate D2 wasm build provided by https://github.com/terrastruct/d2/tree/master/d2js. I built the wasm as per the readme in this directory and trying to load it into my application. However, it fails with linkage errors:

```
Uncaught LinkError: WebAssembly.instantiate(): Import #0 "gojs" "runtime.scheduleTimeoutEvent": function import requires a callable
```
Does it make sense to look into it or better just providing the package as requested in https://github.com/terrastruct/d2/issues/2268?

Issue #2268, opened 2025-01-07
@terrastruct/d2 not published to npm


boehs:
hi, I see https://github.com/terrastruct/d2/tree/master/d2js, looks like I came just in time given it was committed a week ago! I'd love to use it, but the @terrastruct/d2 package 404s. Any chance it could be published?

Issue #136, opened 2022-11-22
feature: WASM web app


caner-cetin:
A web editor, for at least preview can be done with https://go-app.dev/. I may work on this, couldn't find anything similar on issues. pls hit me up if someone already works on a web app.

nhooyr:
This is a goal. @alixander thoughts on doing it for the playground? That way we don't even need a stateful API, all static!!!

alixander:
our wasm build currently is 50M lmao. the vast majority is Goja. the other big dead weight is playwright

<img width="368" alt="Screen Shot 2022-12-06 at 6 56 13 PM" src="https://user-images.githubusercontent.com/3120367/206076873-d3b41dfc-f789-4802-969e-73392ea2ecde.png">

alixander:
even if we remove deadweight and optimize for wasm, realistically we're still looking at 20+ MB. that's a nonstarter for the fast load times desired of the playground

alixander:
i did end up ripping out a lot of shit for tinygo, still no luck. 

<img width="907" alt="Screen Shot 2022-12-06 at 7 54 09 PM" src="https://user-images.githubusercontent.com/3120367/206084363-778b20ba-863c-45e7-8878-53a84f025b6c.png">

alixander:
i suppose 20 MB with a 70% gzip compression isn't that terrible

alixander:
without goja, playwright, elk, dagre, i got the build down to 27M, compressed with gzip is 6.5M. 

compressed with brotli is 4.4M. i think we can work with that. @nhooyr wdyt

nhooyr:
Yea we don't need goja on the web, we can run the JS directly in the browser. See #366 

Sure, 4.4M is excellent.

zachcp:
+1 for this.  Would love to have something like a fully embeddable playground/static site.

alixander:
Thank you for the suggestion. I tried tinygo with some minimal effort ~a year ago but ran into errors. The size shouldn't be an issue anyway once we shed unused dependencies for wasm.

caner-cetin:
i forgot why i asked for this feature and what is D2 but, gratz! and thank you :3

alixander:
haha yeah this one's been on the backburner for a *while*

nhooyr:
Slick! 🚀

zachcp:
Awesome!

F0rgenet:
Tried to build D2js WASM module using tinygo
```fish
[...]

PR #2278, opened 2025-01-12
init d2.js

alixander:
Initializing d2.js but still things to do. Namely, minimize the wasm file size: https://github.com/terrastruct/d2/issues/366

closes https://github.com/terrastruct/d2/issues/2276
closes https://github.com/terrastruct/d2/issues/2272
closes https://github.com/terrastruct/d2/issues/2268
closes https://github.com/terrastruct/d2/issues/136


nazar-pc:
I still don't see `.wasm` file in the 0.6.9 release, how is one supposed to get the file?
https://github.com/terrastruct/d2/issues/2276 was created with the goal of calling it from Rust, but I'm unsure what would be the canonical way to acquire the file if it is not published with the release. Also is there documentation for its API?

I'm asking because https://github.com/terrastruct/d2/issues/2276 was closed, but I'm still not sure how to actually achieve what was requested there.

PR #2305, opened 2025-01-26
d2render: make HTML friendly

alixander:
for https://github.com/terrastruct/d2/issues/2273

allows multiple SVGs to be embedded into HTML doc and pass HTML validation

we need one last flag to close the issue. some unique id that users pass to diagrams, for when you have the same diagram in an html doc, there will be an ID collision (`<mask>` use IDs)


jmooring:
> we need one last flag to close the issue

Will this be a member of the renderOpts?

Issue #2312, opened 2025-01-29
Unique ID that users pass to diagrams


jmooring:
This is neither a bug nor proposal; it's a continuation of the discussion from https://github.com/terrastruct/d2/pull/2305:

> we need one last flag to close the issue. some unique id that users pass to diagrams, for when you have the same diagram in an html doc
>
>> Will this be a member of the renderOpts?
>
> Yes

For our purposes a unique ID is not required. We're already wrapping the `svg` with a `div`, and we can add a unique `id` attribute to the wrapping `div` as needed. 

So unless there's some other compelling use case, I think you can indefinitely defer adding a unique ID and close this issue.

alixander:
@jmooring doesn't HTML validation still not pass regardless of how an SVG is wrapped, so long as an ID it uses isn't globally unique?

jmooring:
To clarify, if we need to add an ID to the wrapper, it will be unique. My point is, we don't need d2 to do that for us.

alixander:
I'm saying when I input something like this in the validator, it doesn't matter how i wrap the svg's, using a duplicate id anywhere within the svg will cause validation to fail.

```xml
<!DOCTYPE html>
<html lang="">
<head>
<title>Test</title>
</head>
<body>
<svg id="unique-1">
     <clipPath id="xyz"></clipPath>
  </svg>
  
  <svg id="unique-2">
     <clipPath id="xyz"></clipPath>
  </svg>
</body>
</html>
```

jmooring:
OK, now I understand. I was looking at this from a wrapper standpoint. We can certainly pass in a unique id with the render opts for every diagram (e.g., ordinal on the page).

PR #2316, opened 2025-01-30
add diagram hash salt

alixander:
closes https://github.com/terrastruct/d2/issues/2312


jmooring:
I tested at 6ae3ede27ff7e6eba8759c842f1fcf1b62c1e45d. Works great. We create a unique ID (ordinal) before making the diagram request, so I just pass that along as the salt value. Nice implementation. Thank you.

PR #2321, opened 2025-02-02
Fix bugs in various files

itsfuad:
Update `getOutputFormat` function in `d2cli/export.go` to properly validate `stdoutFormatFlag`.

* Check if `stdoutFormatFlag` is not nil and not empty before processing.


riotdevy (commented):
good catch

itsfuad:
@riotdevy Yeah but it's very simple one. Nothing too special. Hope they adds it. 😊

alixander (changes requested):
Oh, yes good catch, thank you. Could I trouble you to sign your commit? We require it as an org-wide setting.

itsfuad:
@alixander Hope this works

PR #2343, opened 2025-02-11
d2js: Additional render options

### LINKED ISSUE #2297: d2js: full options args
match cli where it makes sense

### LINKED ISSUE #2307: D2JS - theme/center/pad has no effect on an svg
https://d2lang.com/tour/vars/#configuration-variables

Works on cli but doesn't work with wasm build/d2js.

Image

Image

x-delfino:
PR to add support for the following render options in the wasm module and `d2js`:

  - Pad
  - Center: I'm not exactly sure how this is supposed to be used so not sure how to test it but it does add `preserveAspectRatio="xMidYMid meet"`
  - Dark Theme ID
  - Scale

Fixes #2307 
Fixes #2297 (partially). I haven't added:

- Fonts
- Force Appendix
- Bundle

Force appendix is probably easy, not really sure how bundle works atm and I'm not sure the best way to approach fonts - happy to take more of a look at this though.
  
There is a bit of duplication in the `d2wasm` functions atm (`compile` and `render`) - does that need consolidating? 
  
Neither scale nor dark theme ID currently support `-1` to use default values like the CLI does (rather no value/null should be passed for that option to use the default). does this need to be updated to match CLI behaviour?

screenshot of the customizable example

Screenshot

Thanks for the recent work on d2js, I've got it working in obsidian on iPad which is pretty sweet


x-delfino:
ah I need to sort out the signing on those commits

x-delfino:
I realised this doesn't actually fix #2307 - that issue was specifically referring to configuration variables (e.g. `vars:`) rather than just d2js supporting those options. will look into vars support but probably a bit more complex

x-delfino:
I think `force-appendix` and `bundle` are handled separately by the cli: https://github.com/terrastruct/d2/blob/73f2588f8cbc29c045eb636ddeefd3d4913a6bf3/d2cli/main.go#L901-L908

So defo a bit more work to get them supported

I'm not really sure what's going on for `d2-config` vars. only `sketch: true` seems to make a difference but it only partially sets sketch mode (changes the font):

`sketch: false`:

Screenshot

`sketch: true`:

Screenshot

expected:
Screenshot

stereobooster:
Thanks for the PR.

Just a small note (in case somebody else get this issue). If you use generated SVGs as 

```html
<img src="/path/to.svg" width="..." height="..." />
```

Images inside SVG won't work, unless you use `bundle` option. Because

> External resources (e.g. images, stylesheets) cannot be loaded, though they can be used if inlined through `data:` URLs.
>
> [SVG as an Image Restrictions](https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_as_an_Image#restrictions)

d2js/js/README.md
```diff
@@ -56,6 +56,11 @@ Compiles D2 markup into an intermediate representation.
 Options:
 - `layout`: Layout engine to use ('dagre' | 'elk') [default: 'dagre']
 - `sketch`: Enable sketch mode [default: false]
+- `themeId`: Theme ID to use [default: 0]
+- `darkThemeId`: Theme ID to use when client is in dark mode
```
alixander, line 60: let's use the [capitalization in the codebase](https://github.com/alixander/d2/blob/73f2588f8cbc29c045eb636ddeefd3d4913a6bf3/d2renderers/d2svg/d2svg.go#L79), which is `ID` instead of `Id`

ci/release/changelogs/next.md
```diff
@@ -2,4 +2,6 @@
 
 #### Improvements 🧹
 
+d2js: Support additional render options [#2343](https://github.com/terrastruct/d2/pull/2343)
```
alixander, line 5: let's put which ones

alixander (changes requested):
cool lgtm just some minor comments! Thank you!

P.s. would love to see that Obsidian plugin PR swapping D2 usage to d2.js

x-delfino:
Great - have made those changes now

I realised I haven't updated the defaults in https://github.com/x-delfino/d2/blob/4ce43c324831ff11411362382a41c74694ca18f4/d2js/js/src/index.js#L3-L6

is there much point in setting the additional default options in this wrapper library or just leave it to the d2 defaults?

and yep will get a pr raised for obsidian plugin later - just need to get it down a bit from 80MB...

alixander:
> is there much point in setting the additional default options in this wrapper library or just leave it to the d2 defaults?

All good, just leave as is to d2.

> and yep will get a pr raised for obsidian plugin later - just need to get it down a bit from 80MB...

is that a blocker? I.e. does obsidian have a max plugin size?

d2js/js/examples/customizable.html
```diff
@@ -96,6 +106,74 @@
...
+              <option value="101">Orange creamsicle</option>
+              <option value="102">Shirley temple</option>
+              <option value="103">Earth tones</option>
+              <option value="104">Everglade green</option>
+              <option value="105">Buttered toast</option>
+              <option value="200">Dark mauve</option>
+              <option value="300">Terminal</option>
+              <option value="301">Terminal grayscale</option>
+            </select>
+            Theme ID
```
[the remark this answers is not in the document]
x-delfino: have updated that - I've left the option of no value there still as well, do you want that removed/kept?

x-delfino:
don't think size is a blocker just would be nice to get it down a bit and the wasm module is only ~24MB atm. Also I'm thinking it makes sense to support both running the cli and wasm until they're at feature parity

x-delfino:
I've added in `forceAppendix` support as well, hope it's okay that I've squeezed that into this PR

screenshot

bundle is a bit more complex, I'm not sure of the limitations on WASM but it's failing to download remote images

x-delfino:
> for some reason scale doesn't have effect after a certain point. Not a big deal, probably just some max set by css, just noting it here, not a blocker.
> scale.mp4
> 
> Playing around with it had 2 other small change requests. Sorry wasn't bunched with the first batch, but should be good to go after this.

i think that's set by `max-height` here:

https://github.com/terrastruct/d2/blob/41711d4a4eb6eb827f8fa4e350a324a783824821/d2js/js/examples/customizable.html#L70-L73

I'm assuming that was set for the default scaling to fit

d2js/js/examples/customizable.html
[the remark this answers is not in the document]
alixander: Let's remove as well

alixander:
> bundle is a bit more complex, I'm not sure of the limitations on WASM but it's failing to download remote images

ohh right, yes, bundle is not valid then for d2js client-side, though it can still be used d2js in node.

x-delfino:
missed a null check on the `forceAppendix` which broke things. fixed that now and added a test for force appendix

also I've been digging into some of the d2-config bits and worked out what's going on. The compiler merges in any render options from the source and then sets the defaults:

https://github.com/terrastruct/d2/blob/41711d4a4eb6eb827f8fa4e350a324a783824821/d2lib/d2.go#L72-L73

And the CLI just reuses that same object for the render so has all the correct properties set (including those set in `d2-config`)

So we just need to return an updated render options object from compile that can be passed back to render. I've got this working on [another branch](https://github.com/x-delfino/d2/tree/wasm-options-extended). would you want this as a separate PR or in here? it's a bit more of a change but don't think there's anything breaking. probably needs a bit of a cleanup first

One key thing to getting d2-config working is stripping the defaults from this library and just leaving them to be set by the compiler as d2-config wont be set if the corresponding option is already set

Screenshot

(this removes the need for the blank options in the example UI hence me not updating that here yet)

alixander:
>  remembered that TALA isn't supported so we'll need to keep cli support - are there plans to support it in the wasm module?

I haven't quite decided how that'll work. Since it's client-side, a license key wouldn't really work. I think it would still rely on system calls, but unsure.

alixander:
> would you want this as a separate PR or in here? 

Just throw in here

alixander:
> (this removes the need for the blank options in the example UI hence me not updating that here yet)

That looks 💯 !

d2js/d2wasm/functions.go
```diff
@@ -153,6 +154,26 @@ func GetELKGraph(args []js.Value) (interface{}, error) {
 	return elk, nil
 }
 
+func layoutResolver() func(engine string) (d2graph.LayoutGraph, error) {
+	cached := make(map[string]d2graph.LayoutGraph)
```
x-delfino, line 158: this caching and resolver function might not be required - I just took it from the CLI. Maybe it's just there for the plugin loading?

https://github.com/terrastruct/d2/blob/41711d4a4eb6eb827f8fa4e350a324a783824821/d2cli/main.go#L395-L418

d2js/js/src/worker.browser.js
x-delfino, line 1: so the problem here is that if we conditionally render elk before compiling, it wont be available if the layout is unset but `d2-config` sets the layout to elk. not sure if there's a better way of approaching this

d2js/d2wasm/functions.go
alixander: I think removing it is for the best.

It should still be helpful, but on client-side we don't want the memory usage growing.

d2js/js/src/worker.browser.js
alixander: Oh good point. Yeah this is fine, ELK runs quick.

Can you append to the comment in the code about this?

d2js/d2wasm/types.go
```diff
@@ -33,19 +33,29 @@ type BoardPositionResponse struct {
...
+type CompileOptions struct {
+	RenderOptions
+	Layout *string `json:"layout"`
 }
 
 type CompileResponse struct {
 	FS      map[string]string `json:"fs"`
 	Diagram d2target.Diagram  `json:"diagram"`
 	Graph   d2graph.Graph     `json:"graph"`
+	Options RenderOptions     `json:"options"`
```
alixander, line 58: We should still name this RenderOptions -- it's confusing to get `options` back from a response after sending `options` in the request and have it be a different set

d2js/js/README.md
```diff
@@ -42,24 +42,59 @@ import { D2 } from '@terrastruct/d2';
...
+
+Additional Configuration:
+
+```javascript
+import { D2 } from '@terrastruct/d2';
+
+const d2 = new D2();
+
+const result = await d2.compile('x -> y', {
+    sketch = true,
```
alixander, line 56: `sketch: true,`

d2js/js/README.md
```diff
@@ -42,24 +42,59 @@ import { D2 } from '@terrastruct/d2';
 const d2 = new D2();
 
 const result = await d2.compile('x -> y');
-const svg = await d2.render(result.diagram);
+const svg = await d2.render(result.diagram, result.options);
+```
+
+Additional Configuration:
```
alixander, line 48: `Configuring render options`

(leave a comment linking to section for full render options)

alixander (changes requested):

d2js/d2wasm/functions.go
x-delfino: I've removed the wrapper function and caching now

d2js/js/src/worker.browser.js
x-delfino: I realised it only needs to run if the layout is null or elk so have updated that check now. probably minor but at elk wont need to run if dagre is set as the layout

d2js/js/test/unit/basic.test.js
```diff
@@ -25,6 +25,64 @@ describe("D2 Unit Tests", () => {
...
+    layout-engine: elk
+  }
+}
+x -> y
+`
+    );
+    expect(result.renderOptions.sketch).toBe(true);
+    expect(result.renderOptions.themeID).toBe(4);
+    expect(result.renderOptions.darkThemeID).toBe(200);
+    // expect(result.renderOptions.center).toBe(true);
```
x-delfino, line 48: I think there's a bug for center, but this can be uncommented if https://github.com/terrastruct/d2/pull/2360 is merged

x-delfino: fixed

x-delfino:
alright, I've added also support for `target`, `animateInterval`, `salt` and `noXMLTag`

https://github.com/user-attachments/assets/a8d1a76e-23ab-4ba5-bc9c-5b4168d58642

I mostly copied the logic from the CLI with the exclusion of some relinking stuff. not sure exactly what that does or how best to approach that here

as far as I can tell the main missing functionality now is:
- bundling
- fonts
- outputting different formats

I'm going to have a bit of a look at them to see how much work it would be. happy to either raise those as separate PRs or keep them here if I am successful

x-delfino:
sure thing, I'll keep the other bits for separate PRs so this one doesn't get held up

I'm a bit confused on the failing CI task. It was successful on the prior commit https://github.com/terrastruct/d2/actions/runs/13444494745/job/37566593148. all I changed since then was some css in the d2js example

the tests are running successfully locally, including the specific task that failed here. I don't think the d2js changes should be relevant to this test anyway

```
> go test oss.terrastruct.com/d2/e2etests-cli
ok      oss.terrastruct.com/d2/e2etests-cli     11.420s
```

x-delfino:
hey @alixander - I'm not sure why the CI is failing here, all the tests are running fine locally. the prior run was successful and all I changed since then was CSS in the example webpage. any ideas? other than that, could you take a look at the recent changes please?

have been working on bundling and fonts but will raise another PR for them when they're ready

x-delfino:
sorry noticed there was an error on the conditional elk generation for the browser error

d2js/d2wasm/functions.go
```diff
@@ -241,21 +271,154 @@ func Render(args []js.Value) (interface{}, error) {
 		return nil, &WASMError{Message: "missing 'diagram' field in input JSON", Code: 400}
 	}
 
+	var boardPath []string
+	var noChildren bool
+
+	if input.Opts.Target != nil {
+		switch *input.Opts.Target {
+		case "*":
```
alixander, line 279: `*` should return an error, since d2js doesn't right now support outputting multiple SVGs.

alixander (changes requested):
Sorry about the CI failure, that was just a flaky test (reran the job and it passed).

Looks good to me aside from these things:

d2js/d2wasm/functions.go
x-delfino: so it does actually support targets with children as it supports animated outputs now. I've updated it a bit so it now throws an error if a target with children is picked but `animateinterval` isn't set. In doing so, I've also updated it so it defaults to only outputting the root board. This _doesn't_ match the CLI (which defaults to '*'), but I think makes more sense here, happy to change if you disagree though

alixander (approved):
LGTM, thank you for all these improvements!

MichaelCropper:
Now this PR has been merged, is there a HelloWorld.html example for how to use this? 

i.e. How the release process works or how to embed the relevant .js files from a CDN or similar? 

Happy to pick this up in a Discussion rather than this PR if that's preferred. 

This bit doesn't seem to be particularly well documented.

alixander:
@MichaelCropper Right I have yet to add the CI to publish releases to NPM. We'll get a few more things in before formalizing.

PR #2353, opened 2025-02-14
acii art renderer

### LINKED ISSUE #924: ascii art renderer
our code is set up well to just plug in an ascii art renderer. i suspect still will be quite hard, though we can treat everything as rectangles to start to make it easier.

DrSensor:
/claim #924
close #924

see `docs/**.txt` for the examples output


DrSensor:
still quite big though 😅
maybe caused by default padding 🤔

alixander (changes requested):
Yeah these are too large. I'm looking for diagrams that can be inlined into code. There should be no padding too. 

I'm not quite sure how that can be achieved, because the layout engine gives a certain layout with those long lengths of edges. Shortening those edges would lead to overlaps of objects/text. The bounty placed on the issue is reward for solving that problem.

AkashJana18:
Hey @DrSensor, would you like to collaborate on this issue, or should I go ahead and start working on it? Or are you planning to handle it solo?

DrSensor:
apparently treating it as image processing problem works!
so just resize it like you resize an image.

However,
- It will miss some details when encountering diagonal lines but it should be possible to reconstruct those diagonal lines.
  (TODO: should try other layout engine that doesn't produce diagonal lines).
- Unlike image, text/label in ASCII Art can't be resized. Some option to solve this:
  1. Truncate the text/label.
    example: `d2exporter` -> `d2ex..rter`.
  3. Move text/label to the top or bottom as long as it doesn't collide with the line/edge/connection.
  4. Use ID and Legend.
    example: a shape with text/label `(a)`, then at the bottom of diagram explain that `(a)` is `d2exporter`.

DrSensor:
@alixander does the size looks reasonable to you?

(ignore the missing text/label, it's bug on my side 😂)

alixander:
The size looks reasonable!

1. ELK only uses orthogonal lines (no diagonals). If you can get it working with ELK only and ignore dagre (the layout engine you've been using), that's totally okay.
2. Can you resize only to the point where labels would overflow?

DrSensor:
> 2. Can you resize only to the point where labels would overflow?

It has the caveat that some diagram might get too wide, but sure.
I've been thinking to detect the smallest shape/rect and use it as the resize factor but using the longest label string.length would be much simpler.

alixander:
random inspiration 

https://diagon.arthursonzogni.com/#GraphDAG

Screenshot 2025-02-16 at 10 13 49 PM

DrSensor:
currently it work fine and look decent on dagre layout despite it has some silly quirk (only `/` for diagonal line, no `\` 😅). also the renderer speed could be improved (so many for loop) but for now it's good enough. it can also adjust the height so the shape will fit to the text line count (only the height since the layout engine already fit the shape width with the text length)

NOTE: it will struggle on complex big diagram that has many vertex line but there is no way around this. the least user can do is try to use other layout.

i kinda want to keep this approach instead of limit it to ELK layout. should i continue or is this good enough?

alixander:
It's still far too big =(

[ascii.exp.txt](https://github.com/user-attachments/files/19451294/ascii.exp.txt)

alixander:
It was generated with TALA, but ELK will demonstrate it too (trying to just limit it to orthogonal lines)

https://play.d2lang.com/?script=xJTfbtMwFMbv_RRHHdJgrKjN1pTkAolmsTr1ZmJM3PTGS5zWorEr22GNGA_BE3DDA_IIyEkbN382hNRtN9N6zvH3-77jun9-_fwNQaa0SKmE_ge4uYZApGvCcx-OXx2jqkcizQT34T6NEcDJrq7gNsuBXS0FpxARRdUcASRSpKCX1MhFpdwJuofvCIBTIn2ocdEPhCzXuJgKvoCZ-VOZ-SRystKMKngLa8kiOucgkhKJWvPNKIEZOwVF5TcWUXU650B4DJdXBmz-m86aCa11E1GVpH0JBFuNMq710Jn4kkcipfD65vpNO2-4WVOuqCra5eLrA3uny_ZewfQ_k0272Q60U1yTvLgnTTZUgeBmV6w49KjpzjVb79PZlt8594UsqHqkv2NVzUZSW68CN0fRESidrxhfgNJZkiBkPlK_SJSw1cqH3hF2sYeDHgJQWoqv1NSCyYUbDmytf8divfTBQQC3QsZU9iWJWaZ8GDr16yvFC867HSJJkl5VrCsOqnpD17UHliQWdz5omVErLbjuR2IlpAG447ETjnqdd_LShmpPu8tM6IYTPPxvP87gAfbIGeExLtiNd9RmYxwO8Ohw7LF35n18X7C33-9nZTbe3rOyu95rlwHsBd7hDHgXZ-fDsGbgwat-MrL9te3c-DiYHBAbOsNz12L_te6nov8NAAD__w%3D%3D&layout=tala&

PR #2360, opened 2025-02-18
compiler: support center in d2-config

x-delfino:
This PR fixes support for `center` in `d2-config` variables. Looks like it was only partially implemented but should be supported per [documentation](https://d2lang.com/tour/vars/#configuration-variables).

Test file:

```d2
vars: {
  d2-config: {
    center: true
  }
}

x -> y
```

before change:

`preserveAspectRatio="xMinYMin meet"`

after change:

`preserveAspectRatio="xMidYMid meet"`

### DIFF
--- a/ci/release/changelogs/next.md
+++ b/ci/release/changelogs/next.md
@@ -6,4 +6,6 @@
 
 #### Bugfixes ⛑️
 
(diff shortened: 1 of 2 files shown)


alixander (approved):
good catch, ty!

PR #2382, opened 2025-02-25
d2js: Support Relative Imports

### LINKED ISSUE #2301: d2js: doesn't support import
First of all - I'm grateful for JS version. And it already covers a lot of my wishes.

Second - I do understand why WASM version doesn't support filesystem access out of the box.

But it would be nice to allow for nodejs package (maybe via configuration) to directly access file system

x-delfino:
Adds support for relative imports for d2js

The main change is through adding `InputPath` to `CompileRequest`. This allows for a custom path to be set for the entry point within the provided filesystem e.g.:

```javascript
import { D2 } from '@terrastruct/d2';

const d2 = new D2();

const fs = {
  "project.d2": "a: @import",
  "import.d2": "x: {shape: circle}",
}

const result = await d2.compile({
    fs,
    inputPath: "project.d2",
    options: {
        sketch: true
    }
});
const svg = await d2.render(result.diagram, result.renderOptions);
```

Minor change of throwing any errors in the elk graph generation for easier debugging

I haven't added it to the `customizable.html` example - don't know how complex you want that example to get but I can add another text area and some input boxes for filenames if needed

Fixes #2301


alixander:
> I'm not sure if they were asking for a more dynamic import e.g. you don't have to read the files from disk until d2 compile time

This just comes down to the API. In the CLI interface, we pass a real file system. In the JS interface, we pass a virtual filesystem of a string -> string map. If users want to load files at runtime, the interface would have to allow other types, like string -> callback map. 

I can see both being useful. This PR solves imports for one of the use cases, but indeed not the one the issue is referring to.

d2js/js/README.md
```diff
@@ -92,6 +118,12 @@ All [RenderOptions](#renderoptions) properties in addition to:
 - `salt`: Add a salt value to ensure the output uses unique IDs. This is useful when generating multiple identical diagrams to be included in the same HTML doc, so that duplicate IDs do not cause invalid HTML. The salt value is a string that will be appended to IDs in the output.
 - `noXMLTag`: Omit XML tag `(<?xml ...?>)` from output SVG files. Useful when generating SVGs for direct HTML embedding.
 
+### `CompileRequest`
+
+- `fs`: A mapping of D2 filepaths to their content
+- `inputPath`: The path of the D2 file containing the root D2 board [default: index]
```
alixander, line 124: `The path of the entry D2 file`

d2js/js/README.md
```diff
@@ -92,6 +118,12 @@ All [RenderOptions](#renderoptions) properties in addition to:
 - `salt`: Add a salt value to ensure the output uses unique IDs. This is useful when generating multiple identical diagrams to be included in the same HTML doc, so that duplicate IDs do not cause invalid HTML. The salt value is a string that will be appended to IDs in the output.
 - `noXMLTag`: Omit XML tag `(<?xml ...?>)` from output SVG files. Useful when generating SVGs for direct HTML embedding.
 
+### `CompileRequest`
+
+- `fs`: A mapping of D2 filepaths to their content
```
alixander, line 123: file paths. consistency with above

d2js/js/README.md
```diff
@@ -58,15 +60,39 @@ const result = await d2.compile('x -> y', {
 const svg = await d2.render(result.diagram, result.renderOptions);
 ```
 
+### Supporting Imports
+
+In order to support [imports](https://d2lang.com/tour/imports), a 'filesystem' needs to be passed to the D2 compiler. This is a mapping of D2 file paths to their content.
```
alixander, line 65: Just `### Imports` followed by the example please. A filesystem needs to be passed regardless of whether using imports or not.

alixander (changes requested):
Thanks!

alixander (approved):
lgtm, just a merge conflict from your other PR

x-delfino:
that should be resolved now :)

Issue #1684, opened 2023-10-26
sketch font


imding:
I noticed that the default font for sketch diagrams changed from Architect's Daughter to Fuzzy Bubbles, and when I use the `d2 --font-regular` flag to customise, it only works when the `-s` flag is omitted.

Is there a quick way to customise the fonts? I couldn't find any instructions in the documentation except `style.font = mono`, what's the symbol for the other built-in fonts? i.e. `font.font = ???`

Thanks!

alixander:
related: https://github.com/terrastruct/d2/issues/1460

PR #2384, opened 2025-02-25
d2js: add support for custom fonts

x-delfino:
Adds support for custom fonts in `d2js` by passing a byte array of the ttf file

```javascript
const result = await d2.compile(input, {
  fontBold: [0, 1, 0, 0], // ... contents of .ttf file
});
const svg = await d2.render(result.diagram, result.renderOptions);
```

example screenshot of custom font:

Screenshot

### DIFF
--- a/d2js/d2wasm/functions.go
+++ b/d2js/d2wasm/functions.go
@@ -212,7 +212,7 @@ func Compile(args []js.Value) (interface{}, error) {
 	if fontRegular != nil || fontItalic != nil || fontBold != nil || fontSemibold != nil {
 		fontFamily, err := d2fonts.AddFontFamily("custom", fontRegular, fontItalic, fontBold, fontSemibold)
 		if err != nil {
-			return nil, &WASMError{Message: fmt.Sprintf("custom fonts could not be initialized: %s", err.Error()), Code: 500}
+			return nil, &WASMError{Message: fmt.Sprintf("custom fonts could not be initialized: %s", err.Error()), Code: 400}
 		}
 		compileOpts.FontFamily = fontFamily
 	}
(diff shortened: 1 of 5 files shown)


d2js/d2wasm/functions.go
```diff
@@ -206,9 +229,6 @@ func Compile(args []js.Value) (interface{}, error) {
 	renderOpts := &d2svg.RenderOpts{}
 	if input.Opts != nil && input.Opts.Sketch != nil {
 		renderOpts.Sketch = input.Opts.Sketch
-		if *input.Opts.Sketch {
```
x-delfino, line 209: I removed this as it's already set by the compiiler:

https://github.com/terrastruct/d2/blob/070699309902bf471b5af4e8039fb1e3f3d0df9b/d2lib/d2.go#L218-L220

Although it looks like custom fonts can't be used in sketch mode (both here and the CLI as far as I can tell). Is that expected behaviour or a bug?

alixander: It's a TODO: https://github.com/terrastruct/d2/issues/1684

d2js/d2wasm/functions.go
```diff
@@ -194,6 +193,30 @@ func Compile(args []js.Value) (interface{}, error) {
...
+	if input.Opts != nil && (input.Opts.FontBold != nil) {
+		fontBold = *input.Opts.FontBold
+	}
+	if input.Opts != nil && (input.Opts.FontSemibold != nil) {
+		fontSemibold = *input.Opts.FontSemibold
+	}
+	if fontRegular != nil || fontItalic != nil || fontBold != nil || fontSemibold != nil {
+		fontFamily, err := d2fonts.AddFontFamily("custom", fontRegular, fontItalic, fontBold, fontSemibold)
+		if err != nil {
+			return nil, &WASMError{Message: fmt.Sprintf("custom fonts could not be initialized: %s", err.Error()), Code: 500}
```
alixander, line 215: This is probably a 400, user error.

alixander (changes requested):
Awesome, thanks again -- just a minor comment

d2js/d2wasm/functions.go
x-delfino: okay cool - thanks, sorry should've checked for an issue

alixander (approved):
lgtm!

PR #2393, opened 2025-02-28
d2js: support unicode characters

### LINKED ISSUE #2386: d2js: Garbled Unicode strings
Example:
```js
import { D2 } from '@terrastruct/d2';
const d2 = new D2();
const result = await d2.compile('あいうえお -> ♒️');
const svg = await d2.render(result.diagram, result.options);
```

Text tags in d2js output:
```svg
<text ...>ããããã</text>
<text ...>âï¸</text>
```

Image

Correct output from d2 executable:
```svg
<text ...>あいうえお</text>
<text ...>♒️</text>
```

Image

x-delfino:
fixes unicode support for d2js

closes #2386 (contains screenshot of current behaviour)

Basically `btoa()` doesn't support unicode as detailed [here on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa#unicode_strings) so the unicode characters aren't decoded properly when the processing the response

I followed the approach suggested by MDN and have added in a test as well

working in browser version:

Screenshot

### DIFF
--- a/ci/release/changelogs/next.md
+++ b/ci/release/changelogs/next.md
@@ -27,4 +27,4 @@
   - fixes panic when comment lines appear in arrays [#2378](https://github.com/terrastruct/d2/pull/2378)
   - fixes inconsistencies when objects were double quoted [#2390](https://github.com/terrastruct/d2/pull/2390)
 - CLI: fetch and render remote images of mimetype octet-stream correctly [#2370](https://github.com/terrastruct/d2/pull/2370)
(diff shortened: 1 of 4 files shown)


alixander (approved):
lgtm! thanks

gjf7:
@alixander  Cloud we have npm package release for this? Because we are currently suck by this bug. Thanks!

gjf7:
I also noticed that the last release for @terrastruct/d2 is two month ago

PR #2415, opened 2025-03-09
cli: add validate command

### LINKED ISSUE #2410: Validate command
Hi there!

I'm working on using D2 in my codebase, and in my tests I'd like to just parse the file to validate that the d2 code generated is valid, without actually generating a file. 

I ended up implementing this as `d2 graph.d2 /dev/null`, which works but I think a CLI option would be better since many people don't know about /dev/null

melsonic:
image

/fixes #2410

### DIFF
--- a/ci/release/template/man/d2.1
+++ b/ci/release/template/man/d2.1
@@ -17,6 +17,8 @@
 .Ar fmt Ar file.d2 ...
 .Nm d2
 .Ar play Ar file.d2
+.Nm d2
+.Ar validate Ar file.d2 ...
 .Sh DESCRIPTION
 .Nm
 compiles and renders
@@ -162,6 +164,8 @@ Lists available themes
 Format all passed files
 .It Ar play Ar file.d2
 Opens the file in playground, an online web viewer (https://play.d2lang.com)
+.It Ar validate Ar file.d2 ...
+Validate all input files
 .El
 .Sh ENVIRONMENT VARIABLES
 Many flags can also be set with environment variables.
--- a/ci/release/template/man/d2.1
+++ b/ci/release/template/man/d2.1
@@ -18,7 +18,7 @@
 .Nm d2
 .Ar play Ar file.d2
 .Nm d2
-.Ar validate Ar file.d2 ...
+.Ar validate Ar file.d2
 .Sh DESCRIPTION
 .Nm
 compiles and renders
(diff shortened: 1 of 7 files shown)


ci/release/template/man/d2.1
```diff
@@ -162,6 +164,8 @@ Lists available themes
 Format all passed files
 .It Ar play Ar file.d2
 Opens the file in playground, an online web viewer (https://play.d2lang.com)
+.It Ar validate Ar file.d2
+Validate all input file
```
alixander, line 168: "Validates file.d2"

d2cli/help.go
```diff
@@ -40,6 +41,7 @@ Subcommands:
   %[1]s themes - Lists available themes
   %[1]s fmt file.d2 ... - Format passed files
 	%[1]s play file.d2 - Opens the file in playground, an online web viewer (https://play.d2lang.com)
+  %[1]s validate file.d2  - Validate input file
```
alixander, line 44: Validates file.d2

melsonic:
> Awesome, thanks for taking this!
> 
> For the error case, can we just print out the errors without "failed to validate"? Technically, it successfully validated it to be incorrect.
Yeah, we should be able to do that. 
> 
> For success, I always like printing something out to let users know their command worked. "Success! [x.d2] is valid D2."
Sure, I also thought about that, for better ux.
> 
> Can you add a test to e2etests-cli please?
I'll look into it.

melsonic:
Hey @alixander, should we print "Success! input is valid D2." in case of `Stdin`.

d2cli/watch.go
```diff
@@ -438,6 +438,7 @@ func (w *watcher) compileLoop(ctx context.Context) error {
 			if len(svg) > 0 {
 				err = fmt.Errorf("failed to fully %scompile (rendering partial svg): %w", recompiledPrefix, err)
 			} else {
+				fmt.Println(recompiledPrefix)
```
alixander, line 441: Is this a debug statement accidentally left in?

d2cli/validate.go
```diff
@@ -0,0 +1,41 @@
...
+	"oss.terrastruct.com/util-go/xdefer"
+	"oss.terrastruct.com/util-go/xmain"
+)
+
+func validateCmd(ctx context.Context, ms *xmain.State) (err error) {
+	defer xdefer.Errorf(&err, "")
+
+	ms.Opts = xmain.NewOpts(ms.Env, ms.Opts.Flags.Args()[1:])
+	if len(ms.Opts.Args) == 0 {
+		return xmain.UsageErrorf("validate must be passed an input file")
```
alixander, line 17: `input argument required`

alixander (changes requested):
Can you sign your commits please? We have it as an org-wide setting

d2cli/watch.go
melsonic: Removed

melsonic:
Hey @alixander, updated the PR with signed commits and updated with your changes.

melsonic:
Hey @alixander, updated the PR with signed commits and updated with your suggested changes.

alixander (approved):
LGTM!

Issue #2286, opened 2025-01-14
d2js: Typescript definitions


alixander:

stereobooster:
This what I used in one of my projects

```ts

export type D2Options = {
  /**
   * @default 0
   * Set the diagram theme ID.
   */
  theme?: number;
  /**
   * @default -1
   * The theme to use when the viewer's browser is in dark mode.
   * When left unset --theme is used for both light and dark mode.
   * Be aware that explicit styles set in D2 code will still be
   * applied and this may produce unexpected results. We plan on
   * resolving this by making style maps in D2 light/dark mode
   * specific. See https://github.com/terrastruct/d2/issues/831.
   */
  darkTheme?: number;
  /**
   * Set the diagram layout engine to the passed string. For a
   * list of available options, run layout.
   */
  layout?: "elk" | "dagre";
  /**
   * @default 100
   * Pixels padded around the rendered diagram.
   */
  pad?: number;
  /**
   * @default -1
   * Scale the output. E.g., 0.5 to halve the default size.
   * Default -1 means that SVG's will fit to screen and all others
   * will use their default render size. Setting to 1 turns off
   * SVG fitting to screen.
   */
  scale?: number;
  /**
   * @default false
   * Renders the diagram to look like it was sketched by hand.
   */
  sketch?: boolean;
  /**
   * @default true
   * Bundle all assets and layers into the output svg.
   */
  bundle?: boolean;
  /**
   * Center the SVG in the containing viewbox, such as your
   * browser screen.
   */
  center?: boolean;
};
```

awhitford:
Is there one data structure, or two?  I thought there are `CompileOptions` and `RenderOptions`.

Overall, I'm very interested in this too.

stereobooster:
I wonder if https://github.com/gzuidhof/tygo can be used for this?

hpr1999:
I was experimenting with integrating `d2js` into https://github.com/terrastruct/d2-obsidian and wrote these definitions.
I only have a working prototype for now, but the type definitions seem mostly accurate for the current code.

<details>
<summary>d2.d.ts</summary>

```ts
declare module "@terrastruct/d2" {
  interface RenderOptions {
    /** Enable sketch mode [default: false] */
    sketch?: boolean;
    /** Theme ID to use [default: 0] */
    themeID?: number;
    /** Theme ID to use when client is in dark mode */
    darkThemeID?: number;
    /** Center the SVG in the containing viewbox [default: false] */
[...]

PR #2372, opened 2025-02-23
feat: Typescript definition

jolo-dev:
It closes #2286

PR #2422, opened 2025-03-13
TypeScript signatures

stereobooster:
attempt for https://github.com/terrastruct/d2/issues/2286

all credits go to @hpr1999


alixander:
is this more complete than https://github.com/terrastruct/d2/pull/2372 ?

hpr1999:
It seems there are some properties here that the other PR did not include.

I included all properties from the documentation at the time I wrote them. 

However, the #2372 seems to include some typing of service worker messages that this one does not have.

stereobooster:
> It seems there are some properties here that the other PR did not include.

Well, some of the simply doesn't work. So there is no sense to expose them in documentation. For example, `encode` and `decode` never resolves. Tested with this script:

```js
import { D2 } from "@terrastruct/d2";
const d2 = new D2();
const res = await d2.encode("x -> y");
console.log(res)
```

Other properties meant to be for internal usage. For example, `setupMessageHandler`,  `init`, `worker`,  `sendMessage`. So no sense to expose them.

alixander (approved):

PR #2428, opened 2025-03-16
feat: ability to add border-radius to d2 icons

### LINKED ISSUE #2409: `border-radius` should work on icon
e.g. `icon.style.border-radius: 4`

> I realized yesterday I like their rounded corners of their AWS icons more than the D2 square corner versions (which are otherwise the same icons) so I swapped them out in my diagrams.

https://github.com/terrastruct/d2/issues/2368#issuecomment-2676178195 by @HariSekhon

melsonic:
Screenshot From 2025-03-16 21-10-41

### Final Result

Result
<br/>
/fixes #2409


alixander (commented):
can you add a test in txtar.txt? for 

1. icons on shapes
2. icons on connections
3. icon as an image (even if not done yet)

melsonic:
Hi @alixander, I have updated the expected files with `TESTDATA_ACCEPT=1 go test *the test folder*`, now all tests are passing, and git diff has become pretty big now, so I was wondering if it is usual, or am i doing something wrong here?

d2graph/d2graph.go
```diff
@@ -196,10 +196,11 @@ type Attributes struct {
 	Label           Scalar                  `json:"label"`
 	LabelDimensions d2target.TextDimensions `json:"labelDimensions"`
 
-	Style   Style    `json:"style"`
-	Icon    *url.URL `json:"icon,omitempty"`
-	Tooltip *Scalar  `json:"tooltip,omitempty"`
-	Link    *Scalar  `json:"link,omitempty"`
+	Style     Style    `json:"style"`
+	Icon      *url.URL `json:"icon,omitempty"`
+	IconStyle Style    `json:"iconStyle"`
```
alixander, line 201: I think you can make this an omitempty pointer. The vast majority of objects won't have icon styles (whereas they all have Style). That'll get rid of the large diff

melsonic:
Hey @alixander, I have added `omitempty` option to some struct fields, now the diffs count is significantly reduced. Also added the recommended tests in `txtar.txt`. Please review.

d2compiler/compile.go
```diff
@@ -383,6 +383,17 @@ func (c *compiler) compileField(obj *d2graph.Object, f *d2ir.Field) {
...
 	} else if isReserved {
 		c.compileReserved(&obj.Attributes, f)
+		if keyword == "icon" && f.Map() != nil {
+			for _, ff := range f.Map().Fields {
+				if ff.Name.ScalarString() == "style" && ff.Name.IsUnquoted() {
+					if f.Map() == nil || len(f.Map().Fields) == 0 {
+						c.errorf(f.LastRef().AST(), `"style" expected to be set to a map of key-values, or contain an additional keyword like "style.opacity: 0.4"`)
+						return
+					}
+					c.compileIconStyle(&obj.Attributes, ff.Map())
```
alixander, line 393: Can you just do that within compileReserved? Somewhere around here: https://github.com/alixander/d2/blob/54d751bc873949478eb3adf8bc3e4aa69b70ceb5/d2compiler/compile.go#L548

d2compiler/compile.go
```diff
@@ -747,6 +758,12 @@ func (c *compiler) compileStyle(attrs *d2graph.Attributes, m *d2ir.Map) {
 	}
 }
 
+func (c *compiler) compileIconStyle(attrs *d2graph.Attributes, m *d2ir.Map) {
```
alixander, line 761: let's just reuse `compileStyle`. I know yours has tighter validation for allowed fields, but it's simplest to just re-use, and icons may have more styles applied to them in the future.

d2target/d2target.go
```diff
@@ -493,11 +493,12 @@ type Shape struct {
...
-	Tooltip      string   `json:"tooltip"`
-	Link         string   `json:"link"`
-	PrettyLink   string   `json:"prettyLink,omitempty"`
-	Icon         *url.URL `json:"icon"`
-	IconPosition string   `json:"iconPosition"`
+	Tooltip          string   `json:"tooltip"`
+	Link             string   `json:"link"`
+	PrettyLink       string   `json:"prettyLink,omitempty"`
+	Icon             *url.URL `json:"icon"`
+	IconBorderRadius int      `json:"iconBorderRadius,omitempty"`
```
alixander, line 500: Can you call this `IconStyle`? Sets up for future expansion of styles on icons, e.g. it's own border, its own opacity, etc

e2etests-cli/testdata/TestCLI_E2E/hello_world_png.exp.png
alixander, line 1: you can discard these e2etests-cli tests from the changes. Just slight differences from png generation due to machine architectures.

d2renderers/d2svg/table.go
```diff
@@ -41,6 +41,33 @@ func clipPathForBorderRadius(diagramHash string, shape d2target.Shape) string {
...
+	topX, topY := box.TopLeft.X+box.Width, box.TopLeft.Y
+
+	out := fmt.Sprintf(`<clipPath id="%v-%v-icon">`, diagramHash, shape.ID)
+	out += fmt.Sprintf(`<path d="M %f %f L %f %f S %f %f %f %f `, box.TopLeft.X, box.TopLeft.Y+float64(shape.IconBorderRadius), box.TopLeft.X, box.TopLeft.Y+float64(shape.IconBorderRadius), box.TopLeft.X, box.TopLeft.Y, box.TopLeft.X+float64(shape.IconBorderRadius), box.TopLeft.Y)
+	out += fmt.Sprintf(`L %f %f L %f %f `, box.TopLeft.X+box.Width-float64(shape.IconBorderRadius), box.TopLeft.Y, topX-float64(shape.IconBorderRadius), topY)
+
+	out += fmt.Sprintf(`S %f %f %f %f `, topX, topY, topX, topY+float64(shape.IconBorderRadius))
+	out += fmt.Sprintf(`L %f %f `, topX, topY+box.Height-float64(shape.IconBorderRadius))
+
+	if len(shape.Columns) != 0 {
```
alixander, line 59: Is this special handling for icons on sql-tables? If so, can you include that in the tests to see what it looks like?

d2graph/d2graph.go
```diff
@@ -196,10 +196,11 @@ type Attributes struct {
 	Label           Scalar                  `json:"label"`
 	LabelDimensions d2target.TextDimensions `json:"labelDimensions"`
 
-	Style   Style    `json:"style"`
-	Icon    *url.URL `json:"icon,omitempty"`
-	Tooltip *Scalar  `json:"tooltip,omitempty"`
-	Link    *Scalar  `json:"link,omitempty"`
+	Style     Style    `json:"style"`
+	Icon      *url.URL `json:"icon,omitempty"`
+	IconStyle Style    `json:"iconStyle,omitempty"`
```
alixander, line 201: omitempty doesn't work on a struct, you have to do it on pointers. I guess for consistency, having it be the same type as `style` is cleaner. You should remove the `omitempty` then as its not doing anything.

alixander (changes requested):
looks great, and nice job navigating the compiler code!

d2compiler/compile.go
melsonic: Yeah, Initially I re-used the same function `compileStyle`, but the solution seemed more like a work around, since I need to check if it's a icon style, or general component style. And some tests were failing too. 
But I'll give it a go, if I can find a more feasible solution.

d2target/d2target.go
melsonic: Yep sure, I thought about that too, but since the only requirement was border-radius, so I wrote it that way. I'll change it.

e2etests-cli/testdata/TestCLI_E2E/hello_world_png.exp.png
melsonic: Ok, I'll remote all my changes in this `e2etests-cli` directory. Thanks for your input!

d2renderers/d2svg/table.go
melsonic: No, this will only execute, if `shape: image` is specified. Yeah, we can remove the condition. Thanks again!

d2graph/d2graph.go
melsonic: Yeah, I noticed that too, this didn't reduced the exp file changes. And yeah, since it's not a pointer it will at least have an empty struct.

d2compiler/compile.go
melsonic: Yeah, we can do that too, since we have all the components required in this section
```go
if keyword == "icon" && f.Map() != nil {
	for _, ff := range f.Map().Fields {
		if ff.Name.ScalarString() == "style" && ff.Name.IsUnquoted() {
			if f.Map() == nil || len(f.Map().Fields) == 0 {
				c.errorf(f.LastRef().AST(), `"style" expected to be set to a map of key-values, or contain an additional keyword like "style.opacity: 0.4"`)
				return
			}
			c.compileIconStyle(&obj.Attributes, ff.Map())
		}
	}
}
```
inside compileReserved fn too

melsonic:
Please review @alixander

d2compiler/compile.go
```diff
@@ -741,23 +752,32 @@ func (c *compiler) compileReserved(attrs *d2graph.Attributes, f *d2ir.Field) {
 	}
 }
 
-func (c *compiler) compileStyle(attrs *d2graph.Attributes, m *d2ir.Map) {
+func (c *compiler) compileStyle(attrs *d2graph.Attributes, m *d2ir.Map, compileIconStyle bool) {
```
alixander, line 755: Additional params should always be avoided if possible. 

You'll want to do something like split out the stuff that modifies attrs directly, change existing functions to take a Style pointer, and use that for both regular style and icon style.

alixander (changes requested):

alixander:
Btw you can just hit this button in github to re-request and I'll get pinged =)

Screenshot 2025-03-22 at 8 54 43 AM

d2compiler/compile.go
melsonic: Okay, and that would also probably nullify the need of custom init function for icon Styles `compileIconStyleFieldInit`.

melsonic:
> Btw you can just hit this button in github to re-request and I'll get pinged =)
> 
> Screenshot 2025-03-22 at 8 54 43 AM

Okay, Thanks

d2compiler/compile.go
melsonic: *d2graph.Attributes would be needed in `compileStyleFieldInit` to initialize values for `WidthAttr`, `HeightAttr`, `Top` & `Left`. So, if we need to change existing functions to take Style pointers, We would have to initialize these 4 fields somewhere else.

d2compiler/compile.go
```diff
@@ -741,79 +752,73 @@ func (c *compiler) compileReserved(attrs *d2graph.Attributes, f *d2ir.Field) {
...
 	case "filled":
-		attrs.Style.Filled = &d2graph.Scalar{MapKey: f.LastPrimaryKey()}
-	case "width":
-		attrs.WidthAttr = &d2graph.Scalar{MapKey: f.LastPrimaryKey()}
-	case "height":
-		attrs.HeightAttr = &d2graph.Scalar{MapKey: f.LastPrimaryKey()}
-	case "top":
-		attrs.Top = &d2graph.Scalar{MapKey: f.LastPrimaryKey()}
-	case "left":
-		attrs.Left = &d2graph.Scalar{MapKey: f.LastPrimaryKey()}
```
alixander, line 812: where did these go?

d2compiler/compile.go
```diff
@@ -741,79 +752,73 @@ func (c *compiler) compileReserved(attrs *d2graph.Attributes, f *d2ir.Field) {
...
 		return
 	}
 	if f.Primary() == nil {
 		return
 	}
-	compileStyleFieldInit(attrs, f)
+
 	scalar := f.Primary().Value
-	err := attrs.Style.Apply(f.Name.ScalarString(), scalar.ScalarString())
+	compileStyleFieldInit(styles, f)
```
alixander, line 771: was switching the order intentional?

alixander (changes requested):
nice! few more comments

d2compiler/compile.go
melsonic: So, basically these fields are anyways initialized in `compileReserved` func, whenever width, height, top or left is specified. That's why I removed them from the `styleInit` func. I also checked the scenerio with debugger & it was working fine.

https://github.com/terrastruct/d2/blob/b7b58c592183fb604d051abc8719e76520a57c84/d2compiler/compile.go#L596

d2compiler/compile.go
melsonic: No, changing the order won't have any effect here. Let me know If you would like to preserve the original sequence.

melsonic: Hey @alixander, I have changed the order now, and also resolved the merge conflicts.

d2target/d2target.go
```diff
@@ -503,11 +503,12 @@ type Shape struct {
...
-	Tooltip      string   `json:"tooltip"`
-	Link         string   `json:"link"`
-	PrettyLink   string   `json:"prettyLink,omitempty"`
-	Icon         *url.URL `json:"icon"`
-	IconPosition string   `json:"iconPosition"`
+	Tooltip      string          `json:"tooltip"`
+	Link         string          `json:"link"`
+	PrettyLink   string          `json:"prettyLink,omitempty"`
+	Icon         *url.URL        `json:"icon"`
+	IconStyle    *ShapeIconStyle `json:"iconStyle,omitempty"`
```
alixander, line 510: Let's just make this flat. Just IconBorderRadius. I like having the output be just a one-dimensional data structure that's easy to work with.

alixander: Oh yeah especially since the style on connection is flat

d2renderers/d2svg/table.go
```diff
@@ -41,6 +41,28 @@ func clipPathForBorderRadius(diagramHash string, shape d2target.Shape) string {
...
 
+func clipPathForIconBorderRadius(diagramHash string, shape d2target.Shape) string {
+	box := geo.NewBox(
+		geo.NewPoint(float64(shape.Pos.X), float64(shape.Pos.Y)),
+		float64(shape.Width),
+		float64(shape.Height),
+	)
+	topX, topY := box.TopLeft.X+box.Width, box.TopLeft.Y
+
+	out := fmt.Sprintf(`<clipPath id="%v-%v-icon">`, diagramHash, shape.ID)
```
alixander, line 52: shape.ID might include non-HTML-compliant characters that have to be escaped: https://github.com/alixander/d2/blob/a54f2f66a05c622517a10a89d712c501f0668fcf/lib/svg/text.go#L16

d2renderers/d2svg/table.go
```diff
@@ -41,6 +41,28 @@ func clipPathForBorderRadius(diagramHash string, shape d2target.Shape) string {
 	return out + `fill="none" /> </clipPath>`
 }
 
+func clipPathForIconBorderRadius(diagramHash string, shape d2target.Shape) string {
```
alixander, line 44: `applyIconBorderRadius`

alixander, line 44: why is this defined in table.go?

d2renderers/d2svg/d2svg.go
```diff
@@ -1627,13 +1643,24 @@ func drawShape(writer, appendixWriter io.Writer, diagramHash string, targetShape
...
+			)
+		} else {
+			fmt.Fprintf(writer, `<image href="%s" x="%f" y="%f" width="%d" height="%d" />`,
+				html.EscapeString(targetShape.Icon.String()),
+				tl.X,
+				tl.Y,
+				iconSize,
+				iconSize,
+			)
+		}
```
alixander, line 1663: add a string that gets populated conditionally instead of repeating. `<image href="%s" x="%f" y="%f" width="%d" height="%d"%s/>` (the extra %s at the end)

The reason for this is that when there are more options eventually, we don't want it branching further out

alixander: same in drawConnection

d2renderers/d2svg/table.go
melsonic: `clipPathForBorderRadius` defined for shape is defined here, so thought of keeping it here only. Do you want me to move this to `d2svg.go`?

d2renderers/d2svg/table.go
melsonic: https://github.com/terrastruct/d2/blob/0b2203c107df5319380c1d72753ae8c7814324d9/d2renderers/d2svg/table.go#L25
Should I change this here too?

d2renderers/d2svg/d2svg.go
melsonic: Done

d2renderers/d2svg/table.go
melsonic: Moved to `d2svg.go`, also now passing clipPathId instead of diagramHash, that will reduce the need of creating the same Id twice.

d2target/d2target.go
melsonic: Done

d2compiler/compile.go
```diff
@@ -741,79 +752,73 @@ func (c *compiler) compileReserved(attrs *d2graph.Attributes, f *d2ir.Field) {
...
+func (c *compiler) compileStyleField(styles *d2graph.Style, f *d2ir.Field) {
 	if _, ok := d2ast.StyleKeywords[strings.ToLower(f.Name.ScalarString())]; !(ok && f.Name.IsUnquoted()) {
 		c.errorf(f.LastRef().AST(), `invalid style keyword: "%s"`, f.Name.ScalarString())
 		return
 	}
 	if f.Primary() == nil {
 		return
 	}
-	compileStyleFieldInit(attrs, f)
+
```
alixander, line 769: remove empty line

d2compiler/compile.go
```diff
@@ -741,79 +752,73 @@ func (c *compiler) compileReserved(attrs *d2graph.Attributes, f *d2ir.Field) {
...
+	compileStyleFieldInit(styles, f)
 	scalar := f.Primary().Value
-	err := attrs.Style.Apply(f.Name.ScalarString(), scalar.ScalarString())
+	err := styles.Apply(f.Name.ScalarString(), scalar.ScalarString())
+
```
alixander, line 773: remove empty line

d2renderers/d2svg/d2svg.go
```diff
@@ -1435,6 +1440,11 @@ func drawShape(writer, appendixWriter io.Writer, diagramHash string, targetShape
 		el.Fill = fill
 		el.Stroke = stroke
 		el.Style = style
+		if targetShape.IconBorderRadius != 0 {
+			clipPathId := fmt.Sprintf("%v-%v-icon", diagramHash, svg.SVGID(targetShape.ID))
```
alixander, line 1444: convention in this codebase is to double uppercase ID, `clipPathID` please (everywhere in PR)

d2renderers/d2svg/d2svg.go
```diff
@@ -853,12 +853,17 @@ func drawConnection(writer io.Writer, diagramHash string, connection d2target.Co
 	if connection.Icon != nil {
 		iconPos := connection.GetIconPosition()
 		if iconPos != nil {
-			fmt.Fprintf(writer, `<image href="%s" x="%f" y="%f" width="%d" height="%d" />`,
+			connectionIconClipPath := ""
+			if connection.IconBorderRadius != 0 {
+				connectionIconClipPath = fmt.Sprintf(` clip-path="inset(0 round %fpx)"`, connection.IconBorderRadius)
```
alixander, line 858: it should never be floating point right? Use `%d` instead

d2renderers/d2svg/d2svg.go
```diff
@@ -1435,6 +1440,11 @@ func drawShape(writer, appendixWriter io.Writer, diagramHash string, targetShape
 		el.Fill = fill
 		el.Stroke = stroke
 		el.Style = style
+		if targetShape.IconBorderRadius != 0 {
+			clipPathId := fmt.Sprintf("%v-%v-icon", diagramHash, svg.SVGID(targetShape.ID))
+			fmt.Fprint(writer, applyIconBorderRadius(clipPathId, targetShape))
```
alixander, line 1445: out of curious why does image shape get this treatment while the icons just use a clip-path one liner?

alixander (changes requested):
last set of comments, almost ready to merge!

d2renderers/d2svg/table.go
alixander: yes

d2renderers/d2svg/d2svg.go
melsonic: Basically, it depends on the way `clip-path` is defined in `Render()`. 
https://github.com/terrastruct/d2/blob/0b2203c107df5319380c1d72753ae8c7814324d9/d2themes/element.go#L231
Here it takes `clipPathID` instead of clip-path string.

d2renderers/d2svg/d2svg.go
melsonic: In `connection`, `BorderRadius` is defined as `float64`, so following the same, I have set `IconBorderRadius`'s datatype as `float64`. Also, as per MDN docs, floating values are valid for border-radius.
<br/>
image
<br/>
https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius

In `d2` also, during compilation it stores border-radius value as `Scalar` which in turn stores the values as string, so the user may set it as float right?

alixander: Fair enough

e2etests/txtar.txt
```diff
@@ -171,6 +171,19 @@ a -> b: hi {
   style.underline: true
 }
 
+-- icon-style --
+
+lambda : {
+  icon: https://icons.terrastruct.com/aws%2FCompute%2FAWS-Lambda.svg
+  icon.style.border-radius: 20
```
alixander, line 178: can you add more to this test? e.g. try with 0, 5, 20, 999

https://d2lang.com/tour/style/#border-radius

d2renderers/d2svg/d2svg.go
```diff
@@ -1435,6 +1440,11 @@ func drawShape(writer, appendixWriter io.Writer, diagramHash string, targetShape
 		el.Fill = fill
 		el.Stroke = stroke
 		el.Style = style
+		if targetShape.IconBorderRadius != 0 {
+			clipPathID := fmt.Sprintf("%v-%v-icon", diagramHash, svg.SVGID(targetShape.ID))
+			fmt.Fprint(writer, applyIconBorderRadius(clipPathID, targetShape))
```
alixander, line 1445: Can you just write that here? Would rather it conform to existing code, which is to write outside the `<g>` (https://github.com/melsonic/d2/blob/5ce77a1325ef6b2cf487c48c63854a4a0516c28b/d2renderers/d2svg/d2svg.go#L1328)

alixander (changes requested):

d2renderers/d2svg/d2svg.go
melsonic: Yup, that makes sense

d2renderers/d2svg/d2svg.go
```diff
@@ -1324,6 +1329,13 @@ func drawShape(writer, appendixWriter io.Writer, diagramHash string, targetShape
 	if targetShape.BorderRadius != 0 && (targetShape.Type == d2target.ShapeClass || targetShape.Type == d2target.ShapeSQLTable) {
 		fmt.Fprint(writer, clipPathForBorderRadius(diagramHash, targetShape))
 	}
+	var iconClipPathID string
+	if targetShape.IconBorderRadius != 0 && (targetShape.Type == d2target.ShapeImage) {
+		// https://www.w3.org/Style/CSS/Tracker/issues/29?changelog
```
alixander, line 1334: Add a little more context please. Just so people know this is just documentation of the `min`, and not a todo or bug or anything

e2etests/testdata/txtar/icon-style/elk/sketch.exp.svg
alixander, line 1: can you take a screenshot and post it on this PR to see what this actually looks like?

alixander (changes requested):
looks good to me! Sorry been merging stuff in between, so if you could rebase and regenerate the tests again, it'll be good to go

e2etests/testdata/txtar/icon-style/elk/sketch.exp.svg
melsonic: Sure

d2renderers/d2svg/d2svg.go
melsonic: Done

alixander (approved):
💯 lgtm!

PR #2441, opened 2025-03-22
Fix TypeScript signatures

stereobooster:
I missed small things in the first PR. Some follow up fixes

### DIFF
--- a/d2js/js/CHANGELOG.md
+++ b/d2js/js/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to only the d2.js package will be documented in this file. **Does not
 include changes to the main d2 project.**
 
+## Next
+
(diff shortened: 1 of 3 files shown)


d2js/js/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 All notable changes to only the d2.js package will be documented in this file. **Does not
 include changes to the main d2 project.**
 
+## Next
```
alixander, line 6: ✅

alixander (approved):
I'm not familiar with typescript ecosystem, deferring to you, ty!

stereobooster:
I tested locally, by editing files directly in `node_modules`. It worked at least in my case.

PR #2479, opened 2025-04-02
feat: support for --omit-version flag

### LINKED ISSUE #2377: option to remove version of d2 from rendered contents
Hello and... great tool, thank you!

When used in CI, d2 is installed to render the images.
Images are also committed and become a part of the repo.
Images, as of today, contain the tool version somewhere, so each time the d2 tool version is changed (be it latest or bump on demand, it's CI after all) - all of the images are being changed due to the version change.

Is it possible to exclude d2 **version** from the generated content?
If no such an option now - possible to add?

melsonic:
/fixes #2377


alixander (changes requested):
can you add a test to e2etests-cli please? looks good otherwise!

alixander (approved):
💯 lgtm!

PR #2487, opened 2025-04-08
fix: made gradients work in sketch mode

### LINKED ISSUE #2481: Linear gradients don't work with sketch mode
I was trying out the new linear gradient feature and noticed I get an error when trying to use sketch mode. 

The example from the docs:

```
direction: right
x -> y: hi
y -> z
x.style.fill: "#f4a261"
y.style.fill: honeydew
z.style.fill: "linear-gradient(#f69d3c, #3f87a6)"
```

Returns an error in sketch mode ([playground](https://play.d2lang.com/?script=VMrBCsIwDIDhe54itBcFN9BJ1R58l7Kma6B0EAMue3rxuNvPz5dZaFZee0ThpSpsOLzRIlYG--cO2_hRazQWbi2i8-WebuHqwA6_rp0s0xf2I2_cKcmwSMpMXU--hFee5gv6qTwfKZwd_AIAAP__&sketch=1&)).

```
❯ d2 --sketch 002.d2 002.png
err: failed to compile 002.d2: Invalid color format, url('#grad-984dfec132f72578000afe61cf8daaed70c7c3de')
```

melsonic:
Before d2 was throwing error while using gradient color values in sketch mode. The error was originating from the following line
https://github.com/terrastruct/d2/blob/75633193a9c25e09f9bd9ad869fccb91432c69cb/d2themes/sketch_overlay.go#L27
Basically, It can't parse a element fill value set as `url('#grad-<hash>')`

Now, as a part of the fix, we are checking fill value against a regex `^url\('#grad-[a-f0-9]{40}'\)$`, it it matches, we set the luminance value to `normal`, since we don't have access to actual gradient-colors anymore. Else we calculate the luminance value from the color value.

/fixes #2481


d2themes/sketch_overlay.go
```diff
@@ -24,9 +24,15 @@ func (o *ThemableSketchOverlay) Render() (string, error) {
 	if color.IsThemeColor(o.fill) {
 		o.el.ClassName += fmt.Sprintf(" sketch-overlay-%s", o.fill) // e.g. sketch-overlay-B3
 	} else {
-		lc, err := color.LuminanceCategory(o.fill)
-		if err != nil {
-			return "", err
+		var lc string
+		if color.IsUrlGradientId(o.fill) {
```
alixander, line 28: Url => URL
Id => ID

lib/color/gradient.go
```diff
@@ -246,3 +246,9 @@ var GradientRegex = regexp.MustCompile(`^(linear|radial)-gradient\((.+)\)$`)
 func IsGradient(color string) bool {
 	return GradientRegex.MatchString(color)
 }
+
+var UrlGradientId = regexp.MustCompile(`^url\('#grad-[a-f0-9]{40}'\)$`)
```
alixander, line 250: same here

d2themes/sketch_overlay.go
alixander, line 28: better to fix at the source. Calling Luminance() with a url should give an error, and then `LuminanceCategory` should check for that error, and return normal if it's the type that url gives.

alixander (changes requested):
can you add a test to d2compiler?

melsonic:
> can you add a test to d2compiler?

Sure, I was waiting for your review on the implementation part. Will add it in the next commit.

d2themes/sketch_overlay.go
melsonic: But in case someone sets the fill value as `$45329#`, in this scenerio the error is genuine.
Basically I am thinking of calling `IsURLGradientID(o.fill)` beforehand, and if if matches we return `normal, nil`.

alixander: The error should give a different error type. You would only handle the error types you expect.

melsonic: Hey @alixander, as per your suggestion I went through the `color` validation code, here are my findings -
1. In case of `hex colors`, d2 validates the format using `regex` beforehand, so we are good with hex color codes.
2. In case of `gradient colors` however, the regex validation is not complete, basically it allows anything other than `\n` inside `linear-gradient()` or `radial-gradient()` as long as it has at least one byte (char). 
For example, in case of `linear-gradient(#ggg, #hhh)`, the regex parses it as correct, but in reality hex color codes can't exceed `f`. And `linear-gradient(#ggg, #hhh)` & `linear-gradient(#fff, #000)` gives the same error. I am also attaching one ss below.

image

So, one solution is this one, the second one being we need to change the `gradient regex` that's used to prevalidate gradient colors.

I hope I was able to explain my point, let me know if I missed anything.

melsonic: Here is the `func` for your reference
https://github.com/terrastruct/d2/blob/75633193a9c25e09f9bd9ad869fccb91432c69cb/lib/color/color.go#L509-L515

d2compiler/compile_test.go
```diff
@@ -3894,6 +3894,20 @@ svc_1.t2 -> b: do with B
...
+				vars: {
+					d2-config: {
+						sketch: true
+					}
+				}
+				x->y
+				x.style.fill: "linear-gradient(#000000, #ffffff)"
+				y.style.fill: "linear-gradient(#ffffff, #000000)"
+			`,
+			expErr: "",
```
alixander, line 3909: you don't need to explicitly include this line

d2compiler/compile_test.go
```diff
@@ -3894,6 +3894,20 @@ svc_1.t2 -> b: do with B
 				tassert.Equal(t, "d2/testdata/d2compiler/TestCompile/meow.d2", g.Layers[0].Layers[0].AST.Range.Path)
 			},
 		},
+		{
+			name: "test-gradient-fill-values-in-sketch-mode",
+			text: `
+				vars: {
+					d2-config: {
+						sketch: true
```
alixander, line 3902: i'm not sure if d2-config is actually read in this test suite runner. you can either check by running this test in the master branch, or perhaps adding it to d2sketch tests (i recall we had specific tests for that)

lib/color/color.go
```diff
@@ -86,6 +86,9 @@ func darkenCSS(colorString string) (string, error) {
 }
 
 func LuminanceCategory(colorString string) (string, error) {
+	if IsURLGradientID(colorString) {
```
alixander, line 89: can you leave a comment for this explaining

alixander (changes requested):

d2compiler/compile_test.go
melsonic: Ok

d2compiler/compile_test.go
melsonic: Sure, I’ll check and update with necessary changes.

lib/color/color.go
melsonic: Sure, will do

alixander (approved):
lgtm!

Lastly, could you add a line to `next.md`?

melsonic:
> lgtm!
> 
> Lastly, could you add a line to `next.md`?

Sorry, I forgot to update it this time. Will update in next commit

PR #2492, opened 2025-04-15
validate gradient color stops

### LINKED ISSUE #2490: Validate gradient color stops
Currently, d2 does not validate gradient color stops, it just renders black color in case the gradient colors are invalid. 

Example config: 
```
direction: right
x->y
y.style.fill: "linear-gradient(#ggg, #hhh)"
```

Image

Expected Result:
It should validate each color stop, and throw an error in case it's invalid.

melsonic:
Currently, D2 does not validate gradient color stops, it just renders black color in case the gradient colors are invalid.
This PR adds the functionality to validate each color stop & in case it's invalid, it throws an error.

As per D2 website `fill` & `stroke` supports css color names & hex colors apart from gradients, however inside gradient D2 supports all color formats (e.g rgb, hsl, etc), hence I used `csscolorparser` library instead of checking against `NamedColors` & `ColorHexRegex`.

`fill` docs: https://d2lang.com/tour/style#fill

/fixes #2490

### DIFF
--- a/lib/color/color.go
+++ b/lib/color/color.go
@@ -512,19 +512,18 @@ var NamedColors = []string{
 var ColorHexRegex = regexp.MustCompile(`^#(([0-9a-fA-F]{2}){3}|([0-9a-fA-F]){3})$`)
 
 func ValidColor(color string) bool {
-
 	if IsGradient(color) {
 		gradient, err := ParseGradient(color)
+		if err != nil {
+			return false
+		}
 		for _, colorStop := range gradient.ColorStops {
 			_, err = csscolorparser.Parse(colorStop.Color)
 			if err != nil {
-				break
+				return false
 			}
 		}
-		return err == nil
-	}
-
-	if !go2.Contains(NamedColors, strings.ToLower(color)) && !ColorHexRegex.MatchString(color) {
+	} else if !go2.Contains(NamedColors, strings.ToLower(color)) && !ColorHexRegex.MatchString(color) {
 		return false
 	}
 
--- a/lib/color/color.go
+++ b/lib/color/color.go
@@ -512,7 +512,19 @@ var NamedColors = []string{
 var ColorHexRegex = regexp.MustCompile(`^#(([0-9a-fA-F]{2}){3}|([0-9a-fA-F]){3})$`)
 
 func ValidColor(color string) bool {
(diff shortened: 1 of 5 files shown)


alixander (changes requested):
I think the check is more precise/expected in `ValidColor`. Like it's got a gradient pattern, but it's just not valid.

melsonic:
Hey @alixander , there's another way to solve this problem i.e. by validating colors during rendering instead of at compile time.
Approach 1 (Current PR):
- It validates at compile time (faster error feedback).
- Code changes are limited to gradient.go.
I know this approach also has cons, like we are parsing gradients twice here(once during compilation & secondly during rendering) & as you might said `ValidColor` check might not be as expected.

Approach 2 Closed PR(https://github.com/terrastruct/d2/pull/2491):
- It validates colors while rendering.
- Code changes in multiple files + adding error returns in multiple `func` s.

I went with the first approach since it fails faster and keeps changes contained, but I’d love your thoughts on which one you prefer.

alixander:
I meant if you just add your current check to here: https://github.com/terrastruct/d2/blob/master/lib/color/color.go#L514 , it's cleaner

melsonic:
> I meant if you just add your current check to here: https://github.com/terrastruct/d2/blob/master/lib/color/color.go#L514 , it's cleaner

Ah okay, that makes sense.

lib/color/color.go
```diff
@@ -512,7 +512,19 @@ var NamedColors = []string{
 var ColorHexRegex = regexp.MustCompile(`^#(([0-9a-fA-F]{2}){3}|([0-9a-fA-F]){3})$`)
 
 func ValidColor(color string) bool {
-	if !go2.Contains(NamedColors, strings.ToLower(color)) && !ColorHexRegex.MatchString(color) && !IsGradient(color) {
+
+	if IsGradient(color) {
+		gradient, err := ParseGradient(color)
```
alixander, line 517: check this error first

lib/color/color.go
```diff
@@ -512,7 +512,19 @@ var NamedColors = []string{
...
+		for _, colorStop := range gradient.ColorStops {
+			_, err = csscolorparser.Parse(colorStop.Color)
+			if err != nil {
+				break
```
alixander, line 521: just return false

alixander (changes requested):

alixander (approved):
lgtm!

PR #2504, opened 2025-05-01
fix(lib/jsrunner): use the key parameter in MustGet instead of hardcoded "elkResult"

chai2010:
Previously, MustGet always accessed the hardcoded "elkResult" global key,
ignoring the provided key parameter. Now it correctly uses the key argument,
making the method behavior consistent with its signature.


alixander:
thanks @chai2010 . Would you mind signing your commit? We have it as an org-wide setting

chai2010:
I've signed the commit, PTAL

PR #2509, opened 2025-05-04
feat: add ability to fade multiple

### LINKED ISSUE #1516: faded multiple
looks pretty cool. maybe we can do like `multiple.opacity: 0.5` for this

Screen Shot 2023-07-31 at 1 15 09 PM

melsonic:
Added ability to face the background shape when multiple style is specified.
/fixes #1516 

image


d2ast/keywords.go
```diff
@@ -63,9 +64,10 @@ var StyleKeywords = map[string]struct{}{
 	"text-transform": {},
 
 	// Only for shapes
-	"shadow":        {},
-	"multiple":      {},
-	"double-border": {},
+	"shadow":           {},
+	"multiple":         {},
+	"multiple.opacity": {},
```
alixander, line 69: This is not the way. Otherwise you'd have to manually add every combination of style keywords, e.g. `label.near`

alixander (changes requested):

d2ast/keywords.go
melsonic: Yep, that would be a trade-off of this approach. I'll explore alternative ways to solve this problem and check back with you once I find one.

d2ast/keywords.go
```diff
@@ -75,6 +76,10 @@ var StyleKeywords = map[string]struct{}{
 	"filled":   {},
 }
 
+var CompositeStyleKeywords = map[string]string{
```
melsonic, line 79: Currently, CompositeStyleKeywords is of type `map[string]string`, but in future if we support more styles under `multiple` we can convert it to `map[string][]string` to hold more values under one style.

d2compiler/compile.go
```diff
@@ -813,6 +845,15 @@ func compileStyleFieldInit(styles *d2graph.Style, f *d2ir.Field) {
 	}
 }
 
+func compileCompositeStyleFieldInit(styles *d2graph.Style, f *d2ir.Field, parent string) {
+	switch f.Name.ScalarString() {
```
melsonic, line 849: I coundn't find an way to fetch parent value for current keyword, so defined a new function instead of using the existing func `compileStyleFieldInit`, since we needed an extra parameter holding the value of parent. We can use the existing function also, but have to include a 3rd parameter parent.

d2graph/d2graph.go
```diff
@@ -493,6 +494,25 @@ func (s *Style) Apply(key, value string) error {
 	return nil
 }
 
+func (s *Style) ApplyComposite(key, value, parent string) error {
```
melsonic, line 497: Similar to `compileStyleFieldInit`, we need an extra paramater holding the parent value here also, so created a new func `ApplyComposite`.

d2ast/keywords.go
alixander: we shouldn't need this. for example, the ability to do `label.near` is implemented, but there's no special map for that.

d2compiler/compile.go
alixander: Similar to the other comment, since nested keywords are already implemented elsewhere, you should question why you'd need a new function for this case.

melsonic:
> Big apologies for the delay @melsonic . Had a personal matter for a couple weeks but I'm 100% back now.

Hey @alixander, no worries at all. I noticed you were inactive for a couple of weeks, I just put the comment to notify you once you cone back.

 And welcome back btw : )

melsonic:
Regarding the comments, I'll look into it and come up with an efficient solution.

e2etests/testdata/stable/multiple_person_label/elk/sketch.exp.svg
alixander, line 1: the test results aren't right. this shouldn't have changed

d2graph/d2graph.go
```diff
@@ -486,6 +487,15 @@ func (s *Style) Apply(key, value string) error {
 			return fmt.Errorf(`expected "text-transform" to be one of (%s)`, strings.Join(d2ast.TextTransforms, ", "))
 		}
 		s.TextTransform.Value = value
+	case "multipleopacity":
```
alixander, line 490: where does this come from?

alixander (changes requested):
Sorry about the delay @melsonic .

d2compiler/compile.go
```diff
@@ -756,6 +756,35 @@ func (c *compiler) compileStyleField(styles *d2graph.Style, f *d2ir.Field) {
...
+						return
+					}
+				}
+			} else {
+				if ff.LastPrimaryKey() != nil {
+					c.errorf(ff.LastPrimaryKey(), `unexpected field %s`, ff.Name.ScalarString())
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
melsonic: Sure, we can simplify the process by modifying  
<https://github.com/terrastruct/d2/blob/b2841d586f6c2b238b677b4043f7c8cadfc5f266/d2compiler/compile.go#L788-L792>  
as:

```go
func (c *compiler) compileStyle(styles *d2graph.Style, m *d2ir.Map) {
	for _, f := range m.Fields {
		c.compileStyleField(styles, f)
		if f.Map() != nil {
			for _, ff := range f.Map().Fields {
				c.compileStyleField(styles, ff)
			}
		}
	}
}
```

However, as it’s not possible to get the parent field of a particular field, we need to pass a new parameter to  
<https://github.com/terrastruct/d2/blob/b2841d586f6c2b238b677b4043f7c8cadfc5f266/d2compiler/compile.go#L794>  
to identify the parent field for opacity.

PR #2529, opened 2025-06-03
feat: add d2 pre-commit hook

maxbrunet:
#1528 added the `d2-fmt` hook, and with the addition of #2479, a pre-commit hook can now yield the same output as the prebuilt binaries. Without `--omit-version`, the version is slightly different: `data-d2-version="v0.7.0-HEAD"`. The end-user can always override the `entry` attribute to specify the flags as they prefer

Can be tested with:

```
pre-commit try-repo --all-files --ref=feat/d2-pre-commit-hook https://github.com/maxbrunet/d2 d2
```

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,3 +1,9 @@
+- id: d2
+  name: d2
+  description: Compiles and renders d2 files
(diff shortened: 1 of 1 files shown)


alixander (approved):
lgtm, thanks!

PR #2594, opened 2025-08-05
Fix/js types

mateothegreat:
This is to fix the problems users are facing for importing EJS and CJS modules from d2js.

In addition to the update of `package.json`, the test suite now adequately validates that both ESM and CJS builds can be imported without errors, exports are properly structured, and the main D2 class is
constructible.

Also, the paths for the outputs under `dist/` were changed at some point from `dist/(esm|cjs)` to `dist/(node-esm|node-cjs)` - this is now fixed for future npmjs.com publishing.

> NOTE:
> No code or functionality was affected by this change.

### Changes

- Updated `package.json` and cleaned up + fixed `exports` structure.
- Updated with additional coverage for ESM `import` and CJS `require` functionality.


d2js/js/package.json
```diff
@@ -15,35 +15,33 @@
...
     "index.d.ts"
   ],
-  "types": "./index.d.ts",
   "scripts": {
     "build": "./make.sh build",
-    "test": "bun test test/unit",
+    "test:unit": "bun test test/unit",
     "test:integration": "bun test test/integration",
-    "test:all": "bun run test && bun run test:integration",
+    "test:all": "bun run test",
```
alixander, line 44: 👍 , though I think it's just `bun test` now. 

```
▶ bun run test
error: "/bin/test" exited with code 1
note: a package.json script "test" was not found
```

alixander:
Hi @mateothegreat I plan to do a release of d2 this week and would love to get this in!

PR #2655, opened 2025-09-18
Fix playwright install

alixander:


sanmai-NL:
@alixander Is there a global CLI option to disable fetching and execution of binaries? And an option to invoke `d2` non-interactively (e.g., for CI)?

PR #2666, opened 2025-10-02
animations: icons and non-dashed connections

### LINKED ISSUE #1461: more `animated`
this is a pretty cool diagram.

we should be able to achieve the same effect if we extend `animated`.
- for shapes, it'll do blinking
- for non-dashed lines, it'll make a circle to move along src to dst.

once connections can get icons, `animated` on connections can move the icon instead

### LINKED ISSUE #1949: Animated connections w/o dashed lines have icons
Got this from looking at issue #1461. 

Just some brainstorming, what would the syntax look like? Putting `animated: true` makes the line dashed automatically, so how would we differentiate whether the user wants a dashed line or a line with a moving icon? I'd like to take this issue if possible as well.

alixander:
closes https://github.com/terrastruct/d2/issues/1461
closes https://github.com/terrastruct/d2/issues/1949

animated-connections-gif exp


scottw-kr:
How's the rewrite going? It looks like development kinda dried up around September 2025

PR #2674, opened 2025-10-11
svg: Text may overflow legend bounds when monospace font is used

ocadaruma:
First, thanks a lot for providing great product as an open source!

## Summary
- When we create legend with certainly long text with monospace theme (e.g. Terminal), the text overflows the bound
  * image
  * [playground link](https://play.d2lang.com/?script=hJBNasQwDIX3PsUjew_zszOli256gV5AxGpsxrWD7HQIxXcv-S0tga4s6-npE--TJBt8KcBedZviu--WL1Acf7D21uB2Ps-dQGMaiubY-cgGHO4KqIs3cMfRbt6mwZN-RtMYvKZkIRyo-BSz8_06AuQyBj7lIunOBsL2oK8tZWdwPZIe3hZncJm1unFX7Av9pv6R3zgOacj7LYWk46JJJD0ckz1lRz0btF7awDuhqqoU9V5f1mfa16dcOuGs1FYtMWwrxuCjZdmtczRTdV3nDnI4zKD-MKfittj_Ob2q7wAAAP__&layout=elk&theme=300&)
- The cause is d2svg doesn't pass font family when calculating text dimensions. As the result, the dimension is calculated based on `SourceSansPro` font even when mono is used.
- The screenshot when this fix is applied:
  * image

## Note
- I plan to add e2e test, but I'm not sure about convention e.g. where we should place test and which layout engine should be target... etc.
  * Could you give me some advices about this? Thanks,


alixander (commented):
Thank you for the clear report and PR!

You can add a test here: 

https://github.com/terrastruct/d2/blob/master/e2etests/txtar.txt#L1072

If you use it with Terminal theme (https://github.com/terrastruct/d2/blob/master/e2etests/txtar.txt#L729), it should use the monospace font.

Everything else looks good to me

ocadaruma:
@alixander Thank you for your advice! I added a test.

ocadaruma:
I signed the commit to resolve CI failure. Could you take a look again?

alixander (approved):
perfect, ty again

PR #2707, opened 2026-01-15
Fix SVG class attribute escaping for user-defined classes

muness:
## Summary
- Fix invalid SVG output when user-defined class names contain special characters like quotes
- Add `escapeClassNames` helper function using `html.EscapeString` to properly escape class names
- Add unit tests for the new escaping function

## Problem
When using `{class: foo label: "bar"}` syntax, D2 generates invalid SVG with unescaped quotes in class attributes:

```xml
<g class="YQ== test label: "Hello World"">
```

This breaks SVG validation:
```bash
echo 'a: {class: test label: "Hello World"}' > /tmp/bug.d2
d2 /tmp/bug.d2 /tmp/bug.svg
xmllint --noout /tmp/bug.svg  # fails
```

## Solution
The fix escapes user-defined class names with `html.EscapeString` before they are written to the SVG class attribute. After the fix:

```xml
<g class="YQ== test label: &#34;Hello World&#34;">
```

## Test plan
- [x] Added unit tests for `escapeClassNames` function covering various special characters
- [x] Verified `xmllint --noout` passes on output with quoted class values
- [x] Ran `go test ./d2renderers/d2svg/...` - all tests pass

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/d2renderers/d2svg/d2svg.go
+++ b/d2renderers/d2svg/d2svg.go
@@ -96,6 +96,15 @@ type RenderOpts struct {
 	OmitVersion *bool
 }
 
+// escapeClassNames escapes HTML special characters in class names to produce valid SVG attributes.
+func escapeClassNames(classes []string) []string {
+	escaped := make([]string, len(classes))
+	for i, class := range classes {
+		escaped[i] = html.EscapeString(class)
+	}
+	return escaped
+}
+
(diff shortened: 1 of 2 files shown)


juniorcastro0152-hub:
Protocol /0.?0)

PR #2742, opened 2026-05-02
shape: cycle - trace arcs to shape borders (close #1578)

### LINKED ISSUE #1578: `shape: cycle`
see: https://github.com/terrastruct/d2/discussions/1330

Still undecided about name. I may want to reserve "radial" for 1 thing in center and N connected.

yasumorishima:
## Summary

Closes #1578.

This is a continuation of #2362 by @alixander. That PR introduced the `shape: cycle` layout but the curves did not start or stop on shape borders, which the bounty on #1578 asks to fix. This branch is based on `alixander:cycle` (the head of #2362) and adds the missing border-tracing on top, plus tests and a couple of small geometry fixes.

If preferred, I am happy to either land this as the supersede PR or rebase the four new commits onto a fresh branch off master. Glad to follow whichever workflow suits the maintainers best.

## Problem

In #2362 the cycle edges were sampled as a 30-point arc between source and destination shape centers, then handed to `Edge.TraceToShape` to clip the route to the shape borders. With small shapes on a layout circle, every arc sample near the source center fell inside the source rectangle, so the segment-edge clipper found no crossing and left the curve starting at the shape center instead of on the border.

## Fix

Solve the layout circle against each shape analytically and limit the arc to the angles where it crosses the source and destination borders, then refine the endpoints onto the actual shape outline:

* New `Segment.IntersectCircle(center, radius)` helper in `lib/geo` solves a quadratic for segment-circle crossings (single point on a tangent contact).
* `d2cycle.createCircularArc` now intersects the layout circle with the four edges of each shape's bounding box, picks the angles where the arc enters / exits, and samples between them.
* The endpoints are then passed through `shape.TraceToShapeBorder` so non-rectangular shapes (circle, oval, hexagon, cloud, ...) end on their actual outline rather than on the bounding box border. For rectangles this is a no-op.
* When the analytic step degenerates (zero sweep, no crossing in arc range), `fallbackStraightRoute` produces a straight connection whose endpoints are also clipped to the shape border via the same helper, instead of returning the shape centers.

The renderer side (cubic-bezier emission for `IsCurve`) is left exactly as in #2362, so the visual style of the curves matches that PR.

## Tests

* Regenerated `e2etests/testdata/txtar/cycle-diagram/{dagre,elk}/{board.exp.json,sketch.exp.svg}` with `TA=1`. The new SVGs render every cycle connection starting and ending on a shape border.
* New `TestSegmentIntersectCircle` in `lib/geo/segment_test.go` covers eight cases (chord, tangent, off-origin circle, endpoint on circle, both endpoints inside / outside, vertical chord, degenerate zero-length segment).
* `go test ./d2layouts/... ./lib/geo/...` all green.
* Full `go test ./e2etests/...` was run; only failures are the existing Go 1.26 `url.URL` JSON field-order shifts that already affect `master` (e.g. `icon-label`, `cycle-order`, `sql-table-reserved`) and are unrelated to this change.

## Self-review

Both findings from CodeRabbit on the equivalent fork PR were addressed:

* Endpoints are now clipped to the actual shape outline (not just the bounding box) via `shape.TraceToShapeBorder`.
* `IntersectCircle` no longer reports a tangent contact twice (`disc > 0` guard added with regression test).
* The fallback paths now produce shape-border endpoints instead of returning the shape centers.

## Bounty

/claim #1578


yasumorishima:
Merged master in and resolved the one conflict, which was in `pathData()`. This branch was still carrying the original draft's rewrite of that function, while master has since reformatted the same code with `svg.FormatFloat`, so I took master's version verbatim. None of the draft's extra branches are needed for border tracing: `createCircularArc` is the only place that sets `IsCurve`, and it always emits `ARC_STEPS+1 = 31` points, so master's final-curve branch stays in range, while the degenerate path goes through `fallbackStraightRoute`, which sets `IsCurve = false`. The regenerated `sketch.exp.svg` keeps every arc coordinate (`M 28.476319 -197.929132 …`); the only diff is trailing zeros and master's unused-CSS pruning, and `board.exp.json` is unchanged.

Two follow-ups came out of the merge. They change what a cycle container looks like, so I want to flag them rather than bury them in the diff.

**1. The cycle shape was not drawn by either renderer.** `TestTargetShapePixels` (new on master) requires every entry of `d2target.Shapes` to render, and `cycle` failed it with `unsupported typed geometry for cycle`: square-family shapes have no path commands, so the scene builder's default branch cannot draw them. I handle `cycle` next to `sequence_diagram` and `hierarchy` in the four square-family switches, and added it to the two shape lists those tests enumerate. The SVG renderer had the same gap silently — `cycle` fell to `default:`, where `baseShape.GetSVGPathData()` returns nil, so containers were never painted — so `cycle` is in that rectangle case now too, and SVG and raster agree.

**2. Once the container was visible, it was not sitting on the board it wraps.** `d2layouts` fits the container to the nested board and then places it at `(0, 0)`, and `PositionNested` offsets the contents by the container position alone, so a nested board is expected to start at the origin. `positionObjects` centers the ring on `(0, 0)` instead, which leaves the upper-left half of the board in negative coordinates, and the container was drawn offset by half the ring. `normalizeToOrigin` at the end of the layout translates the shapes and the arcs together — the arcs are included in the bounds so the container encloses them — and records the size on the root for `FitToGraph`. Being a pure translation it does not disturb what this PR is about: the first arc keeps `x = 28.476319 + 227 = 255.476319` exactly, and every shape is inside its container in the regenerated board.

Before (containers absent) and after:

before

after

Also added the changelog entry, which I had missed.

`go build ./...`, `go vet ./...` and `go test -count=1 ./...` are clean on go1.27 (57 packages, e2e included).

If you would rather cycle containers stayed invisible, I am happy to take the rendering back out and find another way to satisfy the pixel-coverage test — I just did not want to leave SVG and PNG disagreeing with each other.

yasumorishima:
Rebased onto master after v0.9.0, and two things in the push are worth a note because they move goldens.

**TALA now ships in the tree**, so the txtar suite runs every case under dagre, elk and tala. This case had no `tala` goldens and failed with a missing board; 51 of the 54 txtar cases carry them, so it now does too.

**The cycle board is rounded up to whole pixels** (`d2layouts/d2cycle/layout.go`). The exporter truncates positions and sizes to integers, and a circle leaves shapes on half pixels, so the container could be written one pixel narrower than the child sitting on its right edge, and its border was drawn through that child:

```
container 1  int(246.815) + int(453.5) = 699
child b      int(647.315) + 53         = 700
```

It only showed under tala, where the board lands on a fractional offset — dagre and elk happened to land flush. Sweeping all 162 txtar goldens for a child outside its container, this was the only case in the suite that is not a sequence diagram lifeline. Rounding the board size up makes the box contain its children under every engine, which is why all three goldens moved: the three now agree on the same relative geometry, where before they differed by a pixel each way.

`go build`, `go vet` and `go test ./...` are clean (78 packages, no failures).

PR #2748, opened 2026-05-12
Trim cycle diagram arcs to object bounds

mgalavai:
## Summary
- route cycle diagram edges as true cubic Bezier circle segments instead of sampled pseudo-curves
- trim arc start/end angles at source and destination object boxes so edges begin/end at bounds
- regenerate cycle-diagram txtar expected JSON/SVG fixtures

## Status
Withdrawn from bounty consideration.

## Test plan
- `go test ./d2layouts/d2cycle ./d2renderers/d2svg`
- `TA=1 ./ci/test.sh ./e2etests -run TestE2E/txtar/cycle-diagram -v`


nagiexplorer88:
I think this still misses the border-clipping part of #1578 for non-rectangular cycle children.

`createCircularArc` trims only against `edge.Src.Box` and `edge.Dst.Box` via `trimStartAngle` / `trimEndAngle`, then assigns `edge.Route = cubicArcRoute(...)`. There is no follow-up through `TraceToShape` / `TraceToShapeBorder`, and the regenerated `cycle-diagram` fixture appears to cover only default rectangle nodes.

That means a cycle child with `shape: circle`, `oval`, or `hexagon` can stop on the bounding box instead of the visible shape outline, while the bounty asks for the paths to start/stop at shape bounds. Adding one non-rectangle cycle fixture and clipping the first/last route points to the actual shape border would cover that case.

mgalavai:
Withdrawing this PR from bounty consideration.

PR #2764, opened 2026-06-06
Fix remote SVG bundling and non-interactive image rendering tests

itsfuad:
imgbundler/png/xgif: fix nondeterministic remote SVG bundling and non-interactive render paths

## Summary

This fixes three issues that were showing up across CI and local runs:

- normalize remote SVG responses before embedding so output does not depend on CDN transfer encoding
- avoid Chromium install prompt failures in non-interactive test runs
- fix GIF frame generation for short animation intervals

## Changes

- `lib/imgbundler`
  - decode `Content-Encoding` for remote image responses before embedding
  - add regression coverage for compressed remote SVG responses

- `lib/png`
  - auto-install Playwright/Chromium when stdin is not a TTY instead of prompting

- `lib/xgif`
  - fix frame count calculation for short durations
  - fix frame delay handling so static and animated paths preserve intended timing
  - add regression coverage for frame count and delay logic

- `e2etests-cli`
  - update the `chain_icon_import` golden output to match normalized embedded SVG content

## Root Cause

`chain_icon_import` was flaky because we were embedding raw remote response bytes for SVG icons. When the server returned compressed content, snapshots could differ by environment even when the SVG itself had not changed.

Local `go test` failures with `failed to read user input: EOF` came from prompting for Chromium installation in a non-interactive process.

GIF failures came from integer division in frame count calculation, which could produce `0` frames for short intervals such as `10ms`.

## Verification

- `GOTOOLCHAIN=go1.25.0 go test ./lib/imgbundler ./lib/xgif -count=1`
- `GOTOOLCHAIN=go1.25.0 go test ./e2etests-cli -count=1`
- `COLOR=1 ./make.sh`

## Screenshots

- N/A


itsfuad:
@alixander Is this PR still relevant or I should close it?

alixander:
closed thank you. sorry haven't looked at PRs and issues in a while, will do a reset shortly

PR #2778, opened 2026-07-14
d2svg: respect reduced motion for animated connections

### LINKED ISSUE #1979: Animated connections don't honor prefers-reduced-motion
Hello there

Coming here because some people at my organisation complained about other people making animated connections in their diagrams which makes that first group kinda dizzy.

I checked the styles that are generated by d2 and it seems you're adding a css animation properties for `path` element.

I'd strongly suggest enhancing it with some sort of CSS property that reacts on `prefers-reduced-motion` query so it's NOT animating at all when it's set to `reduced`.

I believe this small change would increase accessibility of this wonderful tool.

coyaSONG:
## Summary

- add a `prefers-reduced-motion: reduce` override for animated SVG connections
- keep the override conditional so static diagrams and default animation behavior are unchanged
- cover the generated CSS directly and update every affected renderer/E2E snapshot
- add #2778 to the next release changelog

## Why

Animated connection paths receive their animation through an inline style. The reduced-motion rule therefore uses `animation: none !important` on `.animated-connection` so the operating-system accessibility preference reliably wins without changing the normal rendered output.

## Validation

- `./make.sh lint` with the repository-pinned Go 1.25.0 toolchain
- `./make.sh build` with the repository-pinned Go 1.25.0 toolchain
- targeted unit, Dagre/ELK, sketch, dark-theme, bidirectional, and Markdown renderer tests
- `./make.sh test`: all core packages and the complete `e2etests` package pass; the remaining `e2etests-cli` failures are unrelated external-fixture issues (changed remote icon bytes and a Playwright 1.52 driver URL returning 404)

Implementation and validation were assisted by OpenAI Codex; I reviewed the resulting diff and test output.

Fixes #1979


coyaSONG:
I am reducing my active draft inventory to the project-wide review-load limit, so I am closing this unreviewed draft for now while preserving the branch and linked issue #1979. The branch is current with master and includes the generated regression fixtures. It can be reopened after an active slot clears and the head is revalidated.

Issue #1221, opened 2023-04-18
More spacing options for ELK output


martinber:
I'm working with very large diagrams in the ELK layout and I would like to be able to reduce the spacing between the blocks and between the lines.

Running `d2 layout elk` I see that it is possible to set some parameters:

```
      --elk-algorithm string              layout algorithm (default "layered")
      --elk-nodeNodeBetweenLayers int     the spacing to be preserved between any pair of nodes of two adjacent layers (default 70)
      --elk-padding string                the padding to be left to a parent element’s border when placing child elements (default "[top=50,left=50,bottom=50,right=50]")
      --elk-edgeNodeBetweenLayers int     the spacing to be preserved between nodes and edges that are routed next to the node’s layer (default 40)
      --elk-nodeSelfLoop int              spacing to be preserved between a node and its self loops (default 50)
```

But I see that ELK has [A LOT of extra options](https://www.eclipse.org/elk/reference.html) unavailable from d2 and I would like to try them. At least I would like to change the port spacing back to what we had in earlier versions

Do I have a way to change these things without forking d2 and editing the hardcoded values? It would be nice to be able to tweak all the [parameters present in the documentation](https://www.eclipse.org/elk/reference.html)

Thank you

martinber:
Thanks for the answer.

I would like to reduce the spacing in diagrams like the on the picture below. In my case I tend to have blocks like `x` that has arrows going to almost every block downstream. The biggest problem seems to be the spacing between lines and the spacing between the ports of `x`. The spacings should be as small as possible because I have hundreds of nodes

![imagen](https://user-images.githubusercontent.com/6921727/233062400-e59b253f-de5a-45cf-b99c-90b254eac89d.png)

And if I can dream, I would like it to do port sharing and some smart branching of lines so I don't have lots of lines going from `x` to the bottom

![imagen](https://user-images.githubusercontent.com/6921727/233065059-ccb65713-b231-4b0c-a3ca-f80f7f16cde8.png)

[...]

PR #2887, opened 2026-09-08
elk: expose edgeEdgeBetweenLayers spacing

jonapoul:
ELK's `edgeEdgeBetweenLayers` spacing is hardcoded to 50, so there's no way to change it.

Every edge between two layers gets its own routing slot, and each slot adds this much space. So the more links you have between two layers, the further apart ELK pushes them. Lowering `--elk-nodeNodeBetweenLayers` doesn't help, because it only sets a minimum. Especially noticeable in tall charts like https://github.com/d2lang/d2/issues/1221#issuecomment-1514593239.

This adds a `--elk-edgeEdgeBetweenLayers` flag, next to the other four spacing flags. The default stays 50, so nothing changes unless you pass it. ELK's own default is 10, for reference.

Related to #1221.

| edgeEdgeBetweenLayers=10 | edgeEdgeBetweenLayers=50 (default) | edgeEdgeBetweenLayers=100 |
|--|--|--|
| spacing-10 | spacing-50 | spacing-100 |


d2layouts/d2elklayout/layout.go
```diff
@@ -198,31 +199,30 @@ type elkOpts struct {
...
+		FixedAlignment:        "BALANCED",
+		ConsiderModelOrder:    "NODES_AND_EDGES",
+		CycleBreakingStrategy: "GREEDY_MODEL_ORDER",
+		NodeSizeConstraints:   "MINIMUM_SIZE",
+		ContentAlignment:      "H_CENTER V_CENTER",
 		ConfigurableOpts: ConfigurableOpts{
 			Algorithm:       opts.Algorithm,
 			NodeSpacing:     opts.NodeSpacing,
 			EdgeNodeSpacing: opts.EdgeNodeSpacing,
+			EdgeEdgeSpacing: opts.EdgeEdgeSpacing,
```
alixander: **[P2] Preserve the default for existing options literals**

Existing Go callers that construct a `ConfigurableOpts` literal with the five previously available fields now implicitly pass `EdgeEdgeSpacing == 0`. Copying that value here (and in `newContainerLayoutOptions`) replaces the former hardcoded spacing of 50, even though the caller has not opted into changing it.

I verified this with all five previous fields set to their defaults and the graph `a -> x; a -> y; b -> x; b -> y; c -> x; c -> y`: the base uses horizontal routing lanes at y=92,142,192,242, while the PR head places every horizontal segment at y=92, causing unrelated links to overlap.

Please distinguish an unset option from an explicitly requested zero and resolve unset values to 50 in both root and container options. A regression test using an options literal that omits the new field would cover this compatibility case.

- sent from alixander's Codex

alixander (changes requested):
Found one compatibility regression for existing Go API callers: omitting the newly added option changes edge spacing from 50 to 0 and causes unrelated edge segments to overlap. Please preserve the previous default for callers that do not explicitly configure the new option.

Validation at 55977f5: native ELK and CLI test suites passed, selected rendering tests passed (sanity, stable, root, unicode, and themes), and targeted WASM option/profile tests passed. The regression was confirmed by running the same Go program against the PR head and the merge-base layout implementation.

- sent from alixander's Codex

d2layouts/d2elklayout/layout.go
jonapoul: https://github.com/d2lang/d2/pull/2887/commits/745a239bed01769af7fc2c10c4e962c681479473

Made the new value an int pointer to handle fallback behaviour while still allowing explicit zero-values. nil means unset and gets 50.

alixander: Thanks, this resolves the original compatibility issue: omitted values retain 50, explicit zero is preserved, and the original example now matches the pre-PR output.

There is one small follow-up before clearing the review: **[P2] Copy the spacing value before layout.** In `edgeEdgeSpacingOrDefault`, returning `opts.EdgeEdgeSpacing` directly shares the caller's storage (including the global `DefaultOpts` pointer). `Layout` later unmarshals ELK output into that same pointer, so concurrent `DefaultLayout` calls race even when their graphs are independent.

Please change the non-nil branch to:

```go
return go2.Pointer(*opts.EdgeEdgeSpacing)
```

The nil branch already creates independent storage. I verified this on `745a239` with the existing parallel sanity tests:

```sh
go test -race ./e2etests -run '^TestE2E$/^sanity$' -count=1
```

They fail with a data race on the current revision and pass with just the pointer-copy change applied through a local Go overlay. The same tests also pass against the pre-PR layout implementation. The native ELK, CLI, and targeted WASM tests otherwise pass.

- sent from alixander's Codex

jonapoul: Done: https://github.com/d2lang/d2/pull/2887/commits/6f19fd1d8bb79dabab377f472ed9aa06a00938dc

